PR #10, opened 2022-04-14
feat: add core repo dev files and folder

r0x4xn:
## Description
- Adds new folder `apps/core` (subject to change on later stage after re-organizing the apps)
- Migrates tsconfig to be integrated with monorepo
- Migrates eslint to be integrated with monorepo
- Moves `.gitattributes` to the root folder
- Updates `.gitignore` 
- Moves some previous devDependencies to dependencies since the app will be considered as a node-application and eslint requires them to be part of dependencies
- Adds new script command `test` to package.json to follow consistency on naming the `test` script as we do on the other apps


.gitignore
```diff
@@ -84,3 +74,14 @@ artifacts
 packages/prepo-sdk/lib
 tmp
 docs
+
+# --------- prepo-core ---------
+typechain/
+apps/core/artifacts*
+apps/core/cache*
+apps/core/cache
```
ramenforbreakfast, line 82: is there any reason for keeping both of these rather than just `apps/core/cache*`?

ramenforbreakfast (approved):
LGTM, besides one comment!

.gitignore
```diff
@@ -84,3 +74,14 @@ artifacts
...
+cache
+apps/core/deployments/localhost
+apps/core/deployments/goerli
+apps/core/.openzeppelin/unknown-*.json
```
ramenforbreakfast, line 86:
```suggestion
apps/core/.openzeppelin/unknown-31337.json
```
This is something I realized later on that Polygon deployments also fall under "unknown" and we don't want to ignore those. So instead we should specify exactly the `hardhat` test chainId 31337.

apps/core/.solcover.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
+    skipFiles: ["trident", "core/interfaces", "core/mocks"],
```
ramenforbreakfast, line 2:
```suggestion
    skipFiles: ["core/interfaces", "core/mocks"],
```

We are not longer using trident so this can be removed.

.prettierignore
```diff
@@ -6,4 +6,5 @@ public
 dist
 .next
 generated
-LICENSE
\ No newline at end of file
+LICENSE
+apps/core/contracts/trident/
```
ramenforbreakfast, line 10:
```suggestion
```
Same here, no longer need to ignore trident contracts.

.gitignore
```diff
@@ -7,6 +7,7 @@ node_modules
 
 # testing
 coverage
+coverage*
```
XavierEkkel, line 10: do we need both `coverage` and `coverage*`?

.gitignore
```diff
@@ -60,17 +61,6 @@ generated
...
-# Hardhat
-apps/**/data
-apps/hardhat-core/*.txt
-apps/hardhat-core/artifacts*
-apps/hardhat-core/deployments
-apps/hardhat-core/cache*
-apps/hardhat-core/cache
-cache
-artifacts
-
```
XavierEkkel, line 72: why were these here / why are we removing them?

r0x4xn: That was part of a boilerplate I had initially for the monorepo. I'm removing them since they where not relevant for how we have our folder structure and we are handling the gitignore of the hardhat files already 👍

.gitignore
r0x4xn: `coverage` checks for folders
`coverage*` checks for files 

Might be a bit redundant if we know that all the prefixed `coverage-something` files are inside a `coverage/` folder. But I don't think it hurts to check for both possibilities

PR #12, opened 2022-04-15
feat: add core repo sc and deploy files

r0x4xn:
## Description
- Migrated Solidity files as they where. **Not modified**
- Migrated `deploy/*.ts` files. Modified due to eslint rules:
  - Changed the order of import
  - Add return value to each function


XavierEkkel (approved):
LGTM if LGT @ramenforbreakfast

ramenforbreakfast:
LGTM, except I think now that we are moving to a monorepo, we can ditch the `core` subfolder under `contracts`. Any helpers we might build in the future can just sit in their own contract folder and reference other parts of the monorepo.

PR #13, opened 2022-04-15
feat: add core repo sc test files

r0x4xn:
## Note
Test pipeline is not migrated yet. You can still make sure that tests are passing by running `yarn test` on this folder

## Description
- Added test and fixtures files.
- Modified the files to include return values on functions
- Eslint exception for special cases for allowing use of `any`
- Order of imports
- Removed unused imports on files

-----
- Fixed issues where naming convention of variables was colliding with the name of the import

### Wrong
```
import Foo from './foo.ts'

describe('=> Some test on Foo ', () => {
  const Foo = new Foo() 
})
```

### Right
```
import Foo from './foo.ts'

describe('=> Some test on Foo ', () => {
  const foo = new Foo() 
})
```


apps/core/tasks/faucet.ts
```diff
@@ -0,0 +1,46 @@
+// const fs = require("fs");
```
XavierEkkel, line 1: can delete this file

XavierEkkel (approved):
LGTM if tests still pass and if LGT @ramenforbreakfast

ramenforbreakfast (approved):
I have now gotten the core tests to work!
No need to change anything, just documenting problems I encountered trying to install the monorepo packages for future reference.
- Needed to generate a PAT token from my Github account with `read:packages` and `repo` privileges. I used this article to figure out how to authenticate https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-npm-registry
- Needed to be given access to the `prepo-packages` repo to pull down our teams packages.
- apps/frontend/docs requires Node 14.X, I had Node 16.X, thankfully this did not cause any problems for our core contract packages.

PR #20, opened 2022-04-25
feat: acquisition royale lib

chrisling-dev:
moving lib folder of acquisition royale into src
https://www.notion.so/Migrate-prepo-acquisition-royale-fe-private-repository-to-monorepo-18829a9ef30a40e989dc4034eb61fed7


r0x4xn (approved):
We should later take some time to update these files and use the prepo packages for the network constants

PR #42, opened 2022-04-30
feat: acquisition royale enterprises carousel

chrisling-dev:
https://www.notion.so/FE-Project-Management-d3878c29e9534234a50f4f8838525bee?p=ca0ef4b10dcf4e10b17a0ff06e6a0246

https://www.notion.so/FE-Project-Management-d3878c29e9534234a50f4f8838525bee?p=963684290dc04fe9b669e1ee71947db7

Changes
- simplified EnterpriseCard (no more glowing, no more clicking)
- refactored EnterprisesStore, moved signer related stuffs to SignerStore (e.g. signerEnterprises, signerActiveEnterprise), and competition related stuffs to CompetitionStore (e.g. competitionEnterprises, competitionActiveEnterprise)
- simplified lazyLoading algorithm: previously it was not using proper "observable approach", values were loaded in autorun, and updated via a property like `signerEnterprises` and `competitionEnterprises`. Now `signerEnterprises` and `competitionEnterprises` are directly the computed value (with values cached nicely on navigate)
- simplified `EnterpriseCarousel` component to work with this main feature of this PR
- simplified `Competition.tsx` and `MyEnterprises.tsx` by moving redundant computation into `EnterpriseCarousel`

### DIFF
--- a/apps/frontend/acquisition-royale-home/src/components/EnterpriseCarousel.tsx
+++ b/apps/frontend/acquisition-royale-home/src/components/EnterpriseCarousel.tsx
@@ -206,7 +206,7 @@ const EnterpriseCarousel: React.FC<Props> = ({
           <ArrowComponent direction="right" onClick={(): void => swiperRef?.slideNext()} />
         )}
       </InnerWrapper>
-      {enterprises && enterprises.length > 0 && activeIndex !== undefined && (
+      {enterprises && enterprises.length > 1 && activeIndex !== undefined && (
         <Labels>
           {activeIndex + 1}/{enterprises.length}
         </Labels>
--- a/apps/frontend/acquisition-royale-home/src/components/EnterpriseCarousel.tsx
+++ b/apps/frontend/acquisition-royale-home/src/components/EnterpriseCarousel.tsx
@@ -169,7 +169,7 @@ const EnterpriseCarousel: React.FC<Props> = ({
         <SwiperComponent
           onActiveIndexChange={(swiper): void => {
             onActiveSlidesChange?.({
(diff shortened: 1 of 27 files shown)


apps/frontend/acquisition-royale-home/src/components/EnterpriseCarousel.tsx
```diff
@@ -99,108 +95,122 @@ const OverlayActionWrapper = styled.div`
...
+      ))
+
+    return placeholderEnterprises.map((id) => (
+      <SwiperSlideComponent key={id} style={SwiperStyle}>
+        <LoadingCarouselCard key={id} loading={enterprises === undefined} />
+      </SwiperSlideComponent>
+    ))
+  }, [activeIndex, enterprises, placeholderEnterprises])
+
+  const renderContent = useMemo(
```
r0x4xn, line 166: We might not need useMemo on `renderContent` since we already have on on `renderCards` which is the one with the mapping and the conditions

r0x4xn (commented):
Nice! Good job with the refactor. Much cleaner code and easier to follow. 

Not needed for this PR in particular but I think we might need to eventually start writing some tests on those stores we have. Maybe could be done as a maintenance task

apps/frontend/acquisition-royale-home/src/components/EnterpriseCarousel.tsx
chrisling-dev: yes good observation! Thanks!

r0x4xn (approved):

PR #48, opened 2022-05-04
feat: remove AccountAccessController from base DepositHook

ramenforbreakfast:
To make the public testnet more manageable, it was decided to instead have the base `DepositHook.sol` contract exclude the `AccountAccessController.sol`. In addition, for greater flexibility so that we do not have to redeploy the `Hook` in certain situations, I have made the `depositRecord` reconfigurable.

In a following PR, we can add back the tests removed here and have a `WhitelistDepositHook.sol` that extends off `DepositHook.sol` and includes the `AccountAccessController`. The `DepositHook` will need to be modified to allow `hook` to be overridable by an inheriting contract.


Nilay27 (approved):
LGTM!

apps/smart-contracts/core/contracts/WithdrawHook.sol
```diff
@@ -31,6 +31,10 @@ contract WithdrawHook is IHook, Ownable {
     emit VaultChanged(_newVault);
   }
 
+  function setDepositRecord(address _newDepositRecord) external onlyOwner {
```
XavierEkkel, line 34: even better would be:

- removing `_newDepositRecord` from the constructor
- updating `hook()` logic to skip the call to deposit record if deposit record is the zero address

can do this in a followup PR though

apps/smart-contracts/core/contracts/DepositHook.sol
```diff
@@ -26,11 +24,6 @@ contract DepositHook is IHook, Ownable {
     uint256 _initialAmount,
     uint256 _finalAmount
   ) external override onlyVault {
-    require(
-      _accountAccessController.isAccountAllowed(_sender) &&
-        !_accountAccessController.isAccountBlocked(_sender),
-      'Account not allowed to deposit'
-    );
     _depositRecord.recordDeposit(_sender, _finalAmount);
```
XavierEkkel, line 27: even better would be:

- removing `_newDepositRecord` from the constructor
- updating `hook()` logic to skip the call to deposit record if deposit record is the zero address

apps/smart-contracts/core/contracts/WithdrawHook.sol
ramenforbreakfast: Ok, I will merge this and do this in a follow up PR.

PR #57, opened 2022-05-09
feat: market creation utility

ramenforbreakfast:
Added `hardhat` task `create-market` to deploy a `PrePOMarket` based on given parameters.


ramenforbreakfast:
@r0x4xn There is an issue with the introduction of hardhat tasks, the checks fail due to this issue https://github.com/NomicFoundation/hardhat/issues/1511

Essentially the tasks need the generated types, but with Github Actions it is always running as if it is the first run, so the tasks cannot compile unless the types have already been generated.

The solution seems to be to append in front of `npx hardhat compile`, `TS_NODE_TRANSCRIPT_ONLY=1 npx hardhat compile`, would this be doable?

PR #58, opened 2022-05-09
feat: utility for funding collateral and market positions

ramenforbreakfast:
This utility is for minting of `Collateral` and market positions using `BaseToken` either minted by or provided by a `minter`


ramenforbreakfast:
hold off on reviewing until issue mentioned in #57 is resolved.

PR #59, opened 2022-05-09
fix: ignore openzeppelin json file

r0x4xn:
https://www.notion.so/Fix-ignore-openzeppelin-deploy-json-files-and-folder-604cf5ef72f14a1e959e948e30a6aca1


ramenforbreakfast:
Wait @r0x4xn so using the `**/localhost` will specifically omit `deployments/localhost`, or any subfolder named `localhost`? Same question with `unknown-31337.json`, although I am less worried about that one since it is a very specific filename that another folder wouldn't use.

r0x4xn:
@ramenforbreakfast Yes, that's how it would be. Though it was enough with a generic form of ignore.

I've made it more specific to `.openzepplin/` & `deployments/` folders to avoid concerns in that case.

ramenforbreakfast (approved):
LGTM, verified these files are ignored.

PR #61, opened 2022-05-12
fix: hardhat tasks typechain conflict

r0x4xn:
## Description
After investigating other possible solutions, the only solution at the moment is to run `TS_NODE_TRANSPILE_ONLY=1 npx hardhat typechain` before compiling hardhat, running tests or even running hardhat clean.

Reference from the Typechain repo: https://github.com/dethcrypto/TypeChain/issues/371


ramenforbreakfast (approved):
LGTM, I ran deployments and tasks locally and they are working fine. I also see now that the GitHub Actions are not failing anymore.

Thanks for looking into this!

PR #122, opened 2022-06-09
feat: use numberFormatter for percentages

r0x4xn:
https://www.notion.so/Normalize-only-1-2-util-functions-to-handle-number-formatting-across-the-app-e52b2469c7dd470eaa7fa890e6fda841

## Description
Removes `formatPercentage` from the app. We can now use `numberFormatter.percentage()` or `numberFormatter.rawPercentage()` for those cases


georgesmithishere:
perfect example of PR train

PR #152, opened 2022-06-16
feat: subgraph track long short token transfer

chrisling-dev:
https://www.notion.so/FE-Project-Management-d3878c29e9534234a50f4f8838525bee?p=6f2fac3cc92d42219a233528dcd35766

### DIFF
--- a/apps/subgraphs/core-dapp/src/mappings/transaction.ts
+++ b/apps/subgraphs/core-dapp/src/mappings/transaction.ts
@@ -1,6 +1,7 @@
 import { Address, BigInt, ethereum } from '@graphprotocol/graph-ts'
-import { CollateralToken, Transaction } from '../generated/types/schema'
+import { CollateralToken, Token, Transaction } from '../generated/types/schema'
 import { Transfer as CollateralTokenTransfer } from '../generated/types/templates/CollateralToken/CollateralToken'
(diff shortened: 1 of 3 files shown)


apps/subgraphs/core-dapp/src/mappings/transaction.ts
```diff
@@ -54,3 +55,26 @@ export function addCollateralTransactions(event: CollateralTokenTransfer): void
...
+  const toTransaction = makeTransaction(event, event.params.to, ACTIONS_RECEIVE)
+
+  const valueBD = event.params.value.toBigDecimal()
+  const valueUSD = longShortToken.priceUSD.times(valueBD)
+
+  fromTransaction.amount = valueBD
+  fromTransaction.amountUSD = valueUSD
+  fromTransaction.event = EVENTS_TRANSFER
+  fromTransaction.market = longShortToken.market
+  fromTransaction.save()
```
georgesmithishere, line 73: is it possible that 1 transaction will fail and another one succeed?

r0x4xn: If I understand this, the transaction done on chain is only 1. 
After that transaction is done, we split it in two `fromTransaction` and `toTransaction` by grabbing the metadata of it.

So with that said, there's only 1 transaction that will always be truthy here

r0x4xn (approved):

PR #156, opened 2022-06-17
feat: generic erc20 data for every token type

chrisling-dev:
https://www.notion.so/FE-Project-Management-d3878c29e9534234a50f4f8838525bee?p=441899ef824e4b189e78689126df674c


apps/subgraphs/core-dapp/.eslintrc.js
```diff
@@ -11,6 +11,7 @@ module.exports = {
     'prefer-destructuring': 'off',
     'prefer-exponentiation-operator': 'off',
     'no-param-reassign': 'off',
+    eqeqeq: 'off',
```
r0x4xn, line 14: Why do we have to turn it off?

chrisling-dev: wasm/assemblyscript can't compare strings with === here, actually, a lot of the common JS/TS habits we have don't work with wasm/assemblyscript, I'm documenting some of the notes here under FE handbook
https://www.notion.so/WIP-Subgraph-notes-ddd9ab552a5643c482874198d3f2d66d

chrisling-dev: seems like it's best practice for assemblyscript: https://github.com/AssemblyScript/assemblyscript/issues/621

I'm also expecting to use == quite often here

r0x4xn: Got it, as long as we document this then it's cool :) thanks!

r0x4xn (approved):

PR #166, opened 2022-06-21
fix: fixed slippage issue

georgesmithishere:
https://www.notion.so/eb96722b397d42c9908002ca06af88e3?v=1f519ba1849347aa9740faedda4acf3b&p=3e63bca1592341238362137142c52422


apps/frontend/core-dapp/src/features/trade/TradeStore.ts
```diff
@@ -48,6 +51,37 @@ export class TradeStore {
...
+  async quoteExactInput(selectedMarket: MarketEntity): Promise<number | null> {
+    const selectedToken = selectedMarket?.[`${this.direction}Token`]
+    const pool = selectedMarket?.[`${this.direction}Pool`]
+    const state = pool?.poolState
+    const fee = pool?.poolImmutables?.fee
+    if (!fee || !selectedToken || !state || !this.openTradeAmount) {
+      return null
+    }
+    const tokenAddressFrom = this.root.preCTTokenStore.uniswapToken.address
+    const tokenAddressTo = selectedToken.address ?? ''
```
r0x4xn, line 67: We should check that this value is truthy and not fallbacking to an empty string. That will cause an error

apps/frontend/core-dapp/src/features/trade/TradeStore.ts
```diff
@@ -48,6 +51,37 @@ export class TradeStore {
...
+    const quoterContract = new ethers.Contract(
+      UNISWAP_QUOTER_ADDRESS.mainnet ?? '', // all uniswap contracts has same address on all chains
+      QuoterABI,
+      this.root.web3Store.coreProvider
+    )
+    const func: ethers.ContractFunction<BigNumber> = quoterContract.callStatic.quoteExactInputSingle
+    try {
+      const result = await func(tokenAddressFrom, tokenAddressTo, fee, this.openTradeAmount, 0)
```
r0x4xn, line 75: What is the `0` representing here?

r0x4xn (changes requested):
image

What does expected output amount is representing here? 

There's no way the user will not get $182 after opening a trade position of $100 just instantly.
Feels that the value you are getting back from the `quoteExactInputSingle` should be manipulated to calculate this a bit different. I would expect a value very close to the `$100` here

chrisling-dev (changes requested):

apps/frontend/core-dapp/src/features/trade/TradeStore.ts
```diff
@@ -48,6 +53,59 @@ export class TradeStore {
...
+        const longTokenPrice = this.direction === 'long' ? price : 1 - price
+        this.valuation = calculateValuation({ longTokenPrice, payoutRange, valuationRange })
+      }
+      return this.amountOut
+    } catch (e) {
+      this.amountOut = undefined
+      this.valuation = undefined
+      this.root.toastStore.errorToast('Error calculating output amount', e)
+      return null
+    }
```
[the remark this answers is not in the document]
georgesmithishere: thanks!

chrisling-dev (changes requested):
some minor comments

apps/frontend/core-dapp/src/features/trade/TradeStore.ts
```diff
@@ -48,24 +51,68 @@ export class TradeStore {
...
+
   // eslint-disable-next-line require-await
   async openTrade(selectedMarket: MarketEntity): Promise<{ success: boolean; error?: string }> {
     const selectedToken = selectedMarket[`${this.direction}Token`]
     const price = selectedMarket[`${this.direction}TokenPrice`]
     const fee = selectedMarket[`${this.direction}Pool`]?.poolImmutables?.fee
     const { swap } = this.root.swapStore
     const { uniswapToken } = this.root.preCTTokenStore
-    if (!selectedToken?.address || price === undefined || fee === undefined)
+    const amountOut = await this.quoteExactInput(selectedMarket)
```
chrisling-dev, line 106: another UX issue here is that, because there's no single source of truth for amountOut, users could be viewing valuation based on amountOut at time X, and place a trade at time Y when the rate has changed, I have a supporting PR to help solve this

chrisling-dev: Feel free to review and merge if all looks good: https://github.com/prepo-io/prepo-monorepo/pull/204

chrisling-dev (changes requested):

liamaharon:
Notes from FE sync

Let's follow Uniswap and also show
1. ~Average Valuation Price~ -> Expected Valuation [before slippage]
2. Valuation Impact [average valuation - current valuation (percent change)]
3. Maximum Valuation After Slippage ([slippage]%)

Screen Shot 2022-07-12 at 10 47 26

- can format 3. or not bother formatting 3
- make Valuation Impact Percentage red if > 5%

apps/frontend/core-dapp/src/features/trade/TradeTransactionSummary.tsx
```diff
@@ -59,7 +66,7 @@ const TradeTransactionSummary: React.FC = () => {
     {
       label: 'Average Valuation Price',
       tooltip: <EstimatedValuation marketName={selectedMarket.name} />,
-      amount: estimatedValuation,
+      amount: valuation === undefined ? undefined : `$${significantDigits(valuation)}`,
```
chrisling-dev, line 69: one thing I missed is that we should disable the button inside TransactionSummary and make it "loading" when `valuation === undefined`

chrisling-dev (changes requested):

chrisling-dev:
`Maximum Valuation After Slippage (1.5%)` should always be higher than Expected Valuation - after slippage, `amountOut` should be lower which means higher price hence higher valuation)

Screenshot 2022-07-14 at 10 04 14 PM

apps/frontend/core-dapp/src/features/trade/TradeTransactionSummary.tsx
```diff
@@ -57,9 +75,20 @@ const TradeTransactionSummary: React.FC = () => {
...
+    },
+    {
+      label: 'Valuation Impact',
+      amount: `${valuationImpact}%`,
+      ignoreFormatAmount: true,
+      warning: +(valuationImpact || 0) > HIGHLIGHT_IMPACT_BREAKPOINT,
+    },
+    {
+      label: `Maximum Valuation After Slippage (${numberFormatter.percent(slippage)})`,
+      amount: `$${significantDigits(estimatedValution * (1 + slippage))}`,
```
chrisling-dev, line 91: This isn't quite right, we can't just apply slippage directly to estimatedValuation. Because of the relation between valuation and the actual long/short token price, 1.5% applied directly to estimated valuation could become a completely different number.

The math here is extremely complicated to explain IMO, I'll just send in an assisting PR shortly

chrisling-dev: Actually, I'll just try my best to explain here:

If we use the formula here https://docs.prepo.io/concepts/markets#calculating-market-valuation-and-payouts to find the Long Token Price based on Valuation, we can get long token price of valuation with slippage applied directly to valuation (valuation taken at a trade size of 1000 usd):
0.01 + ((32077621228.958897 - 15000000000) / (45000000000 - 15000000000)) * (0.99 - 0.01) = **0.5678689601**

But if we apply slippage directly to token amount out (which is what slippage is about, the lowest amount out user is willing to accept):
const amountOutAfterSlippage = 1810.3374221679871 * 0.985 = 1783.1823608354673
const tokenPriceAfterSlippage = 1000 / amountOutAfterSlippage

We'll get **tokenPriceAfterSlippage = 0.5607951390521124** which is quite different from applying slippage directly to valuation.

apps/frontend/core-dapp/src/features/trade/TradeTransactionSummary.tsx
```diff
@@ -47,8 +63,10 @@ const TradeTransactionSummary: React.FC = () => {
 
   if (selectedMarket === undefined) return null
 
-  const estimatedValuation = selectedMarket.estimatedValuation?.value
-    ? `$${significantDigits(selectedMarket.estimatedValuation?.value)}`
+  const estimatedValution = valuation ?? 0
```
chrisling-dev, line 66: typo 
```suggestion
  const estimatedValuation = valuation ?? 0
```

apps/frontend/core-dapp/src/features/trade/TradeTransactionSummary.tsx
chrisling-dev: Also this would be wrong for shorting, when shorting, the valuation after slippage should be lower than "Expected Valuation", cause user will receive less short token, which means more expensive short token price, hence causing lower long token price, hence lower valuation after slippage

chrisling-dev: Understand that this could be overwhelming, will send in a PR to assist shortly

chrisling-dev (changes requested):

apps/frontend/core-dapp/src/features/trade/TradeTransactionSummary.tsx
georgesmithishere: thanks @chrisling-dev !

chrisling-dev (changes requested):
Nice! Very clean solution! Just one last thing and this is good to go, It should be Minimum valuation after slippage when user is shorting, cause the valuation is going downwards

Screenshot 2022-08-02 at 3 57 46 PM

chrisling-dev (commented):
LGTM but test seems to be failing

chrisling-dev (approved):

PR #132, opened 2022-06-12
fix: position icon on same with level with button main label

georgesmithishere:
https://www.notion.so/eb96722b397d42c9908002ca06af88e3?v=1f519ba1849347aa9740faedda4acf3b&p=6291a00edd9941b2862287b361e11d13

PR #175, opened 2022-06-24
chore: syncing the last changes to the prepo-token-private since the initial migration

ramenforbreakfast:
Migrating the last `prepo-token-private` PRs #132-138 including changes to `PurchaseHook.sol`, `TokenShop.sol`, their test suites, and the `PPOStaking.sol` test suite. 

In addition
- moved `mock` contracts that did not pertain to `PPOStaking.sol` and its periphery contracts out of the `ppostaking` folder and into the root `token` folder.
- removed `ethereum-waffle` package which is no longer needed due to usage of `@nomicfoundation/hardhat-chai-matchers`

Note: The previous PR migrates the majority of test suites to use a shared `utils` file, but there seems to be still some test suites that were not migrated. **I did not do so in this PR**, and will need to do this in a later PR.


Nilay27 (approved):
LGTM, changes are working fine and tests are running.

PR #188, opened 2022-06-27
feat: update all contracts to use Solidity version 0.8.7

Nilay27:
* Updated solidity pragma for all smart contracts from 0.8.6 to 0.8.7.
* Had to modify hardhat config, changed runs from 99999 to 999. As 99999 was causing the following error for `PPOStaking.sol:`
`Error: Transaction reverted: trying to deploy a contract whose code is too large`

* The default value of the optimiser is 200 and by increasing its value the deployment becomes expensive but later transactions become cheaper ([source](https://docs.soliditylang.org/en/v0.8.4/using-the-compiler.html#optimizer-options))
* I checked for other projects like UniswapV3, the value of the optimiser was 800 ([source
](https://github.com/Uniswap/v3-core/blob/main/hardhat.config.ts)) so setting a value of 999 should be fine.
* Moreover, the value of `runs` for version 0.8.6 in our repo was only 200.


apps/smart-contracts/token/contracts/ppostaking/rewards/staking/StakingRewardsDistribution.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
-pragma solidity 0.8.6; // TODO: Change version to 0.8.7
+pragma solidity =0.8.7; // TODO: Change version to 0.8.7
```
XavierEkkel, line 2:
```suggestion
pragma solidity =0.8.7;
```

apps/smart-contracts/token/contracts/ppostaking/interfaces/IStakingRewardsDistribution.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
-pragma solidity 0.8.6; // TODO: Change version to 0.8.7
+pragma solidity =0.8.7; // TODO: Change version to 0.8.7
```
XavierEkkel, line 2:
```suggestion
pragma solidity =0.8.7;
```

XavierEkkel (commented):
this PR needs rebasing onto `main`

apps/smart-contracts/token/hardhat.config.ts
```diff
@@ -26,7 +26,7 @@ const config: HardhatUserConfig = {
         settings: {
           optimizer: {
             enabled: true,
-            runs: 99999,
+            runs: 999,
```
XavierEkkel, line 29: we should ideally use the largest `runs` value possible that won't cause deployment gas to be too high.
try decreasing it from 99999 until it compiles

as another reference, Aave V3 uses `runs: 100000`
https://github.com/aave/aave-v3-core/blob/master/hardhat.config.ts#L40

XavierEkkel: also, are we able to just override the runs here value without needing to write the other fields (version, enabled) again?

apps/smart-contracts/token/contracts/ppostaking/governance/staking/WithdrawalRights.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: UNLICENSED
-pragma solidity =0.8.6;
+pragma solidity =0.8.7;
 
 // TODO update compiler to 0.8.7
```
XavierEkkel, line 4:
```suggestion
```

apps/smart-contracts/token/contracts/ppostaking/governance/staking/deps/PPOGamifiedTokenStructs.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
-pragma solidity =0.8.6;
+pragma solidity =0.8.7;
 
 // TODO move to IPPOGamifiedToken
 // TODO update compiler to 0.8.7
```
XavierEkkel, line 5:
```suggestion
```

apps/smart-contracts/token/contracts/ppostaking/governance/staking/interfaces/ITimeMultiplierCalculator.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
-pragma solidity =0.8.6;
+pragma solidity =0.8.7;
 
 // TODO change compiler to 0.8.7
```
XavierEkkel, line 4:
```suggestion
```

apps/smart-contracts/token/hardhat.config.ts
ramenforbreakfast: I am going to make a call here, as Roan has explained, due to the way the config file is structured, I am not sure how we could make it so that we only have to override the `runs` field. Because the `runs` field ultimately belongs to the root `solidity` field, we have to override the entire `solidity` field of the config.

However, we should still continue with identifying a ballpark number of `runs` for the optimizer by assuming that we don't care about deployment cost and want to subsidize usage.

We already gain a lot of benefits from sourcing the config from a default config `...hardhatConfig`, and just need to override the `solidity` field. This is par for the course in most SC projects and is already more than what many projects do.

I have just discussed all of the above with @Nilay27. 
@XavierEkkel or @r0x4xn , let us know here if you have any strong objections.

ramenforbreakfast:
Have looked over this PR, but will make a final review once rebased.

apps/smart-contracts/token/hardhat.config.ts
Nilay27: * Changed runs to 25000 by taking the approach as mentioned by @XavierEkkel i.e started with 99999 and decreased it till the code broke. 
* It was failing for up to 25500.

ramenforbreakfast (approved):
Need to change `pragma solidity 0.8.7` to `pragma solidity =0.8.7` for `PPO.sol`, `PregenPass.sol` and `MockPregenPass.sol`.

Other than that, I have tested your branch and it looks fine with the new optimizer settings.

XavierEkkel (approved):
LGTM

PR #189, opened 2022-06-27
feat: switch token contracts to SafeOwnable

Nilay27:
* Switching to SafeOwnable version of nominate + accept for all the ownable files of TokenContracts, except OwnableUpgradeable ones (i.e `IOUPPO.sol` and `PPO.sol`).  Added TODO for switching to SafeOwnableUpgradeable in future.
* Will Switch Core Dapp contracts in a later PR as they are a bit more complicated and can not be directly switched to SafeOwnable.


apps/smart-contracts/token/contracts/PregenPass.sol
```diff
@@ -1,16 +1,16 @@
...
-import "@openzeppelin/contracts/access/Ownable.sol";
 import "@openzeppelin/contracts/token/ERC721/extensions/ERC721Enumerable.sol";
+import "prepo-smart-contracts/contracts/SafeOwnable.sol";
 
-contract PregenPass is Ownable, ERC721Enumerable {
+contract PregenPass is SafeOwnable, ERC721Enumerable {
   uint256 private _id;
   string private _uri;
 
   /// @dev owner needs to be set as the prePO governance address not the deployer
```
XavierEkkel, line 11: replace the old comment with:
`Owner will be set as the deployer. After deployment, the nominated owner needs to accept the nomination for ownership to be transferred.`

and let's change `_owner` to `_nominatedOwner` to be clearer

could you add this to all constructors where `_owner` (now `_nominatedOwner`) is passed in?

ramenforbreakfast:
I might be mistaken, but please in the future refer to this document https://dev.to/michalbryxi/rebasing-a-pr-train-does-not-have-to-be-a-pain-3amc and stick to _interactive_ rebasing. 

It looks like from the commit history that you resolved conflicts for and re-committed all the stuff from the previous `feat/smart-contract-pragma-update`. 

Looks like there was a ton of commits you had to work through which was probably very annoying and its very easy to make mistakes and accidentally change things you didn't mean to change in previous commits.

As I've explained in my DM to you, interactive rebasing is going to make your life much easier and reduce the chance of problems from human-error.

apps/smart-contracts/token/test/PregenPass.test.ts
```diff
@@ -15,14 +15,19 @@ describe('PregenPass', () => {
...
   const URI_2 = 'https://newBaseURI2/'
 
-  const setupPregenPass = async (): Promise<void> => {
+  const deployPregenPass = async (): Promise<void> => {
     ;[deployer, owner, user, user2] = await ethers.getSigners()
     pregenPass = await mockPregenPassFixture(owner.address, URI)
   }
 
+  const setupPregenPass = async (): Promise<void> => {
+    await deployPregenPass()
```
ramenforbreakfast, line 24:
```suggestion
```
If you're going to start separating out the setup from the deployment, dont tie them together again like this. It can get confusing since this function both deploys and sets the contract up even though you've decided that deploying and setup is a different process.

apps/smart-contracts/token/contracts/ppostaking/rewards/staking/StakingRewardsDistribution.sol
```diff
@@ -1,13 +1,13 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 pragma solidity =0.8.7;
 
-import "@openzeppelin/contracts/access/Ownable.sol";
+import "prepo-shared-contracts/contracts/SafeOwnable.sol";
```
ramenforbreakfast, line 4: move this to the bottom of the import stack for consistency

apps/smart-contracts/token/contracts/ppostaking/governance/staking/WithdrawalRights.sol
```diff
@@ -1,10 +1,10 @@
 // SPDX-License-Identifier: UNLICENSED
 pragma solidity =0.8.7;
 
-import "@openzeppelin/contracts/access/Ownable.sol";
+import "prepo-shared-contracts/contracts/SafeOwnable.sol";
```
ramenforbreakfast, line 4: move this to bottom of imports for consistency

apps/smart-contracts/token/test/PregenesisPoints.test.ts
```diff
@@ -44,14 +44,19 @@ describe('PregenesisPoints', () => {
     merkleTree = generateAccountAmountMerkleTree(eligibleNodes)
   }
 
+  const setupPregenesisPoints = async (): Promise<void> => {
+    await deployPregenesisPoints()
```
ramenforbreakfast, line 48:
```suggestion
```

Same here, in fact why even have a `setup` function if its just doing the accepting? Lets just remove the `setup` function (also do it above). It's more necessary when there are many configurations that need to be done, but here we are just accepting ownership.

apps/smart-contracts/token/test/TokenShop.test.ts
```diff
@@ -41,6 +41,11 @@ describe('TokenShop', () => {
     tokenShop = await tokenShopFixture(owner.address, paymentToken.address)
   }
 
+  const setupTokenShop = async (): Promise<void> => {
+    await deployTokenShop()
```
ramenforbreakfast, line 45:
```suggestion
```
Same here.

apps/smart-contracts/token/test/Vesting.test.ts
```diff
@@ -52,15 +52,24 @@ describe('Vesting', () => {
     )
   }
 
+  const setupVesting = async (): Promise<void> => {
+    await deployVesting()
```
ramenforbreakfast, line 56:
```suggestion
```
Same here.

apps/smart-contracts/token/test/Vesting.test.ts
```diff
@@ -52,15 +52,24 @@ describe('Vesting', () => {
...
     before(async () => {
-      await setupVesting()
+      await deployVesting()
     })
 
-    it('sets owner from initialize', async () => {
-      expect(await vesting.owner()).to.not.eq(deployer.address)
+    it('sets nominee from initialize', async () => {
+      expect(await vesting.getNominee()).to.not.eq(deployer.address)
+
```
ramenforbreakfast, line 67:
```suggestion
```
No new line here.

apps/smart-contracts/token/test/PregenPass.test.ts
```diff
@@ -37,8 +42,12 @@ describe('PregenPass', () => {
       expect(await pregenPass.symbol()).to.eq('PREGENPASS')
     })
 
-    it('sets owner from constructor', async () => {
-      expect(await pregenPass.owner()).to.eq(owner.address)
+    it('sets nominee from constructor', async () => {
```
ramenforbreakfast, line 45: Why is it that in some `sets nominee from constructor/initialize` testcases, you `expect` that the nominee is _not_ the deployer AND that it is the owner, and some (like here) you don't? 

Lets keep it consistent and ensure in all of these sorts of testcases we check both.

ramenforbreakfast (approved):
LGTM, besides Xaviers comments.

My concerns are mostly somewhat minor regarding the test suite.

apps/smart-contracts/token/test/PregenPass.test.ts
XavierEkkel: I think I disagree here - deployment is part of setting up (we do this in other test files), and even though this function will only be 2 lines (for now), it'll still save many extra lines in total, so I think it's fine how it was before.

XavierEkkel (approved):
LGTM after comments

apps/smart-contracts/token/test/PregenPass.test.ts
ramenforbreakfast: Ok @Nilay27, just ignore what we discussed this morning and revert it back to the way it was.

PR #191, opened 2022-06-28
feat: improve deploy prod pipeline

r0x4xn:
https://www.notion.so/Fix-deploy-to-production-pipeline-for-frontend-projects-when-smart-contract-project-is-changed-52b33266a80d4365b205a7f447da5dbb

### DIFF
--- a/.github/workflows/deploy-production-apps.yml
+++ b/.github/workflows/deploy-production-apps.yml
@@ -82,6 +82,7 @@ jobs:
 
     - name: Download artifact
       uses: dawidd6/action-download-artifact@v2
+      continue-on-error: true
       with:
         github_token: ${{ secrets.GH_PAT }}
         workflow: deploy-preview-apps.yml


.github/workflows/deploy-production-apps.yml
```diff
@@ -82,6 +82,7 @@ jobs:
 
     - name: Download artifact
       uses: dawidd6/action-download-artifact@v2
+      continue-on-error: true
```
XavierEkkel, line 85: is this going to continue on error for all apps or just smart contract apps?

r0x4xn: Yes, this rule on github ignores all errors. Smart contract apps doesn't use this workflow so it doesn't matter. Frontend apps use. If if it errors for a frontend app (which is very unlikely) it will still error on the next step for them

XavierEkkel: what about other kinds of apps, e.g. subgraphs?

r0x4xn: Only frontend apps use this workflow. Subgraphs are deployed manually, there's no pipeline for deploying them atm

r0x4xn: In other words, it's safe to trigger error here. `Download artifact` step is only for frontend apps and if something fails downloading the artifact, deploying the app to production will also fail (the next step on the workflow)

XavierEkkel: For future reference: it errors on that step if there is no change done on any frontend app, so we need to continue in those cases.

PR #206, opened 2022-07-15
feat: adding interface for allow list for ppo contract

Nilay27:
* Adding `IAllowlist` interface.
* This contract will have the functionality to add and remove the list of addresses to allowed sources and destinations.
* Also, `resetDestination` and `resetSources` will follow a logic similar to the one implemented in one of the core dapp [contracts](https://github.com/prepo-io/prepo-monorepo/blob/fae573c0bf17b36628ce0fc5e0cf21f4b594659d/apps/smart-contracts/core/contracts/AccountAccessController.sol#L21).

### DIFF
--- a/apps/smart-contracts/token/contracts/interfaces/IAllowList.sol
+++ b/apps/smart-contracts/token/contracts/interfaces/IAllowList.sol
@@ -3,15 +3,11 @@ pragma solidity =0.8.7;
 
 //TODO: add all natspecs at the end
 interface IAllowlist {
-  function allowDestinations(address[] memory accounts) external;
+  function setDestinations(address[] calldata destinations, bool[] calldata allowed) external;
 
-  function removeDestinations(address[] memory accounts) external;
+  function resetDestinations(address[] calldata allowedDestinations) external;
 
-  function resetDestinations() external;
+  function setSources(address[] calldata sources, bool[] calldata allowed) external;
 
-  function allowSources(address[] memory accounts) external;
-
-  function removeSources(address[] memory accounts) external;
-
-  function resetSources() external;
+  function resetSources(address[] calldata allowedSources) external;
 }
--- a/apps/smart-contracts/token/contracts/interfaces/IAllowList.sol
+++ b/apps/smart-contracts/token/contracts/interfaces/IAllowList.sol
@@ -1,0 +1,17 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+pragma solidity =0.8.7;
+
+//TODO: add all natspecs at the end
+interface IAllowlist {
+  function allowDestinations(address[] memory accounts) external;
+
+  function removeDestinations(address[] memory accounts) external;
+
+  function resetDestinations() external;
+
+  function allowSources(address[] memory accounts) external;
+
+  function removeSources(address[] memory accounts) external;
+
+  function resetSources() external;
+}


ramenforbreakfast (approved):
LGTM, just want to note that I use `remove` instead of `block` to avoid confusion with  the `Blocklist`

apps/smart-contracts/token/contracts/interfaces/IAllowList.sol
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+pragma solidity =0.8.7;
+
+//TODO: add all natspecs at the end
+interface IAllowlist {
+  function allowDestinations(address[] memory accounts) external;
```
XavierEkkel, line 6:
```suggestion
  function setDestinations(address[] calldata destinations, bool[] calldata allowed) external;
```

- `setDestinations` allows us to do all setting (assuming we're under the gas limit) within one call, which is convenient when calls are made via governance processes
- `calldata` should be used since we won't be modifying the passed-in array
- `accounts` is implied by the data type, so should call it something more meaningful

apps/smart-contracts/token/contracts/interfaces/IAllowList.sol
```diff
@@ -0,0 +1,17 @@
...
+
+  function removeDestinations(address[] memory accounts) external;
```
XavierEkkel, line 8:
```suggestion
```
no longer needed as it's grouped within `setDestinations`

apps/smart-contracts/token/contracts/interfaces/IAllowList.sol
```diff
@@ -0,0 +1,17 @@
...
+
+  function resetDestinations() external;
+
+  function allowSources(address[] memory accounts) external;
+
+  function removeSources(address[] memory accounts) external;
```
XavierEkkel, line 14:
```suggestion
  function setSources(address[] calldata sources, bool[] calldata allowed) external;
```

apps/smart-contracts/token/contracts/interfaces/IAllowList.sol
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+pragma solidity =0.8.7;
+
+//TODO: add all natspecs at the end
+interface IAllowlist {
+  function allowDestinations(address[] memory accounts) external;
+
+  function removeDestinations(address[] memory accounts) external;
+
+  function resetDestinations() external;
```
XavierEkkel, line 10:
```suggestion
  function resetDestinations(address[] calldata allowedDestinations) external;
```
resets, and allows any destination addresses passed in

apps/smart-contracts/token/contracts/interfaces/IAllowList.sol
```diff
@@ -0,0 +1,17 @@
...
+
+  function allowSources(address[] memory accounts) external;
+
+  function removeSources(address[] memory accounts) external;
+
+  function resetSources() external;
```
XavierEkkel, line 16:
```suggestion
  function resetSources(address[] calldata allowedSources) external;
```
resets, and allows any sources addresses passed in

XavierEkkel (approved):
LGTM after suggestions

PR #208, opened 2022-07-16
fix: add missing functions to ISafeOwnable, fixed license

ramenforbreakfast:
Since we need `SafeOwnableUpgradeable` to form the base of `PPO.sol`, I was looking at `ISafeOwnable` since its going to use the same interface.

I noticed that the interface `ISafeOwnable` is missing functions. I suspect it is related to the fact that `Ownable` does not have an interface because it is an abstract contract, and the functions that were already inherited from `Ownable` were left out.

However, our convention is we typically list all functions inside the interface, and then override all of them in the implementation. So I have fixed `ISafeOwnable` to also include functions from `Ownable`.

Also I noticed, `Ownable` is an abstract contract because it isn't meant to be deployed on its own, yet we made `SafeOwnable` **not**  abstract. I suspect this was speed along testing since that meant no need to define a mock contract to deploy it? 

I think this is mostly a code-quality issue and doesn't functionally affect anything. We can backlog this for later if we want to make it `abstract`.


XavierEkkel:
> Since we need `SafeOwnableUpgradeable` to form the base of `PPO.sol`, I was looking at `ISafeOwnable` since its going to use the same interface.
> 
> I noticed that the interface `ISafeOwnable` is missing functions. I suspect it is related to the fact that `Ownable` does not have an interface because it is an abstract contract, and the functions that were already inherited from `Ownable` were left out.
> 
> However, our convention is we typically list all functions inside the interface, and then override all of them in the implementation. So I have fixed `ISafeOwnable` to also include functions from `Ownable`.

Agree we have a full interface file (regardless of whether it's abstract or not).

> Also I noticed, `Ownable` is an abstract contract because it isn't meant to be deployed on its own, yet we made `SafeOwnable` **not** abstract. I suspect this was speed along testing since that meant no need to define a mock contract to deploy it?
> 
> I think this is mostly a code-quality issue and doesn't functionally affect anything. We can backlog this for later if we want to make it `abstract`.

I think there's no particular advantage to defining it as abstract since it's already fully implemented (and having it as non-abstract has the advantage of not needing a mock contract as you said).

packages/prepo-shared-contracts/contracts/SafeOwnable.sol
```diff
@@ -12,11 +12,16 @@ contract SafeOwnable is ISafeOwnable, Ownable {
     _;
   }
 
-  function transferOwnership(address _account) public virtual override onlyOwner {
+  function transferOwnership(address _account)
```
XavierEkkel, line 15: nit:
```suggestion
  function transferOwnership(address _nominee)
```
to make the behaviour clearer

packages/prepo-shared-contracts/contracts/interfaces/ISafeOwnable.sol
```diff
@@ -1,9 +1,13 @@
-// SPDX-License-Identifier: UNLICENSED
+// SPDX-License-Identifier: AGPL-3.0-or-later
 pragma solidity =0.8.7;
 
 // TODO: add natspec comments
 interface ISafeOwnable {
   event NomineeUpdate(address indexed previousNominee, address indexed newNominee);
 
+  function transferOwnership(address account) external;
```
XavierEkkel, line 8: nit:
```suggestion
  function transferOwnership(address nominee) external;
```
to make the behaviour clearer

XavierEkkel (approved):
LGTM after nits

PR #210, opened 2022-07-16
Removing PPO and IOUPPO

ramenforbreakfast:
There was not much done to the previous iteration of `PPO.sol` and it will just be confusing when we review if we are changing the existing contract.

So i think it's best we remove `PPO.sol` and `IOUPPO.sol`, since we are also not using `IOUPPO.sol` and it relies on `PPO.sol`


XavierEkkel (approved):

Nilay27 (approved):
LGTM

PR #213, opened 2022-07-16
feat: initialized ppo and added method to set transfer hook

Nilay27:
* initialised PPO with name, symbol and nominee.
* Added method to set TransferHook with relevant tests.

### DIFF
--- a/apps/smart-contracts/token/contracts/PPO.sol
+++ b/apps/smart-contracts/token/contracts/PPO.sol
@@ -2,21 +2,19 @@
 pragma solidity =0.8.7;
 
 import "./interfaces/IPPO.sol";
-import "@openzeppelin/contracts-upgradeable/security/ReentrancyGuardUpgradeable.sol";
 import "@openzeppelin/contracts-upgradeable/token/ERC20/extensions/ERC20BurnableUpgradeable.sol";
 import "prepo-shared-contracts/contracts/SafeOwnableUpgradeable.sol";
 
-contract PPO is IPPO, ReentrancyGuardUpgradeable, SafeOwnableUpgradeable, ERC20BurnableUpgradeable {
+contract PPO is IPPO, SafeOwnableUpgradeable, ERC20BurnableUpgradeable {
   ITransferHook private _transferHook;
 
   function initialize(
-    address _nominatedOwner,
     string memory _name,
-    string memory _symbol
+    string memory _symbol,
+    address _nominatedOwner
   ) public initializer {
     __Ownable_init();
     __ERC20_init(_name, _symbol);
-    __ReentrancyGuard_init_unchained();
     transferOwnership(_nominatedOwner);
   }
 
--- a/apps/smart-contracts/token/contracts/PPO.sol
+++ b/apps/smart-contracts/token/contracts/PPO.sol
@@ -8,11 +8,11 @@ import "prepo-shared-contracts/contracts/SafeOwnableUpgradeable.sol";
 contract PPO is IPPO, ReentrancyGuardUpgradeable, SafeOwnableUpgradeable {
   function initialize(address _nominatedOwner) public initializer {}
 
-  function setTransferHook(ITransferHook newTransferHook) external override onlyOwner {}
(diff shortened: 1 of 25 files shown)


apps/smart-contracts/token/contracts/PPO.sol
```diff
@@ -3,16 +3,38 @@ pragma solidity =0.8.7;
...
 import "prepo-shared-contracts/contracts/SafeOwnableUpgradeable.sol";
 
-contract PPO is IPPO, ReentrancyGuardUpgradeable, SafeOwnableUpgradeable {
-  function initialize(address _nominatedOwner) public initializer {}
+contract PPO is IPPO, ReentrancyGuardUpgradeable, SafeOwnableUpgradeable, ERC20BurnableUpgradeable {
+  ITransferHook private _transferHook;
 
-  function setTransferHook(ITransferHook _newTransferHook) external override onlyOwner {}
+  function initialize(
+    address _nominatedOwner,
```
XavierEkkel, line 13: nit: make this the 3rd argument to keep the order consistent with how we call the functions within

apps/smart-contracts/token/contracts/PPO.sol
```diff
@@ -3,16 +3,38 @@ pragma solidity =0.8.7;
...
+    _transferHook = _newTransferHook;
+  }
 
   function mint(uint256 _amount) external override onlyOwner {}
 
-  function burn(uint256 _amount) external override onlyOwner {}
+  function burn(uint256 _amount) public override(IPPO, ERC20BurnableUpgradeable) onlyOwner {}
+
+  function burnFrom(address _account, uint256 _amount)
+    public
```
XavierEkkel, line 32: We want these to be `external` not `public` since we won't be calling these functions internally.

apps/smart-contracts/token/test/fixtures/PPOFixtures.ts
```diff
@@ -1,7 +1,11 @@
 import { ethers, upgrades } from 'hardhat'
 import { PPO } from '../../types/generated'
 
-export async function ppoFixture(governanceAddress: string): Promise<PPO> {
+export async function ppoFixture(
+  governanceAddress: string,
```
XavierEkkel, line 5: nit:
```suggestion
  nominatedOwnerAddress: string,
```
for consistency with actual function

XavierEkkel (approved):
LGTM after comments

apps/smart-contracts/token/contracts/PPO.sol
Nilay27: We can not make these functions `external` as it is already defined as public in `ERC20BurnableUpgradeable` which we inherit. 

So the function visibility of these methods can not be changed from `public` to `external` when we are overriding them.

XavierEkkel: ok, please run solhint and address the warning regarding the ordering of public vs external functions

apps/smart-contracts/token/contracts/PPO.sol
```diff
@@ -3,16 +3,38 @@ pragma solidity =0.8.7;
...
 
-  function setTransferHook(ITransferHook _newTransferHook) external override onlyOwner {}
+  function initialize(
+    address _nominatedOwner,
+    string memory _name,
+    string memory _symbol
+  ) public initializer {
+    __Ownable_init();
+    __ERC20_init(_name, _symbol);
+    __ReentrancyGuard_init_unchained();
```
ramenforbreakfast, line 19:
```suggestion
    __ReentrancyGuard_init();
```
Don't need to run the unchained one since it has no conflicting initializations with `ERC20` and `Ownable`

Also because there is no conflict, can re-order inits to follow inheritance.
```
__ReentrancyGuard_init();
__Ownable__init();
__ERC20_init(_name, _symbol);
```

ramenforbreakfast (approved):
LGTM, besides a few minor comments.

apps/smart-contracts/token/contracts/PPO.sol
Nilay27: Reentrancy guard is not getting used anywhere, so removing this.

PR #215, opened 2022-07-16
feat: added burn and burnFrom methods in PPO

Nilay27:
* adding methods for burn and burnFrom with relevant tests.
* Had to override `burn` and `burnFrom` despite inheriting `ERC20BurnableUpgradeable` as they are also defined in IPPO.sol, and since we directly implement our interfaces, solidity will complain that both contain a reference to `burn` and `burnFrom`.
* Also, as visibility of both `burn` and `burnFrom` are `public` in `ERC20BurnableUpgradeable` hence their visibility in main ppo contract is `public`.


apps/smart-contracts/token/test/PPO.test.ts
```diff
@@ -153,4 +154,109 @@ describe('=> PPO', () => {
...
+      await ppo.connect(owner).mint(user1.address, 10)
+    })
+
+    it('reverts if burn from zero address', async () => {
+      await expect(ppo.connect(user1).burnFrom(ZERO_ADDRESS, 1)).revertedWith(
+        'ERC20: burn amount exceeds allowance'
+      )
+    })
+
+    it('reverts if caller not approved', async () => {
```
ramenforbreakfast, line 209:
```suggestion
    it('reverts if allowance < amount', async () => {
```
The way you setup this testcase doesn't really verify any of the starting parameters needed for this testcase. 

You should be checking that the allowance is less than the amount being burnt and that the user has a sufficient balance to be burnt.

Edit: **you can delete this testcase**, refer to the comment below for `reverts if caller approved and amount > allowance`

apps/smart-contracts/token/test/PPO.test.ts
```diff
@@ -153,4 +154,109 @@ describe('=> PPO', () => {
...
+      const user1PPOBalanceBefore = await ppo.balanceOf(user1.address)
+
+      await expect(
+        ppo.connect(user2).burnFrom(user1.address, user1PPOBalanceBefore.sub(1))
+      ).to.revertedWith('ERC20: burn amount exceeds allowance')
+    })
+
+    it('reverts if caller approved and amount > allowance', async () => {
```
ramenforbreakfast, line 217:
```suggestion
    it('reverts if amount > allowance', async () => {
```
Oh I think i see the misunderstanding. `approve` is just the function someone uses to increase their allowance. So there is no need to make a distinction between `if caller not approved`, and `caller approved and amount > allowance`, because someone calling `approve` is inherently increasing their allowance.

apps/smart-contracts/token/test/PPO.test.ts
```diff
@@ -153,4 +154,109 @@ describe('=> PPO', () => {
...
+      const user1PPOBalanceBefore = await ppo.balanceOf(user1.address)
+      await ppo.connect(user1).approve(user2.address, user1PPOBalanceBefore.sub(1))
+
+      await expect(
+        ppo.connect(user2).burnFrom(user1.address, user1PPOBalanceBefore)
+      ).to.revertedWith('ERC20: burn amount exceeds allowance')
+    })
+
+    it('reverts if caller approved and amount > balance', async () => {
```
ramenforbreakfast, line 226:
```suggestion
    it('reverts if allowance > amount and amount > balance', async () => {
```
Once again, doesn't really make sense to include "caller approved" as a descriptor.
What you are more specifically saying is that the owner's allowance is sufficient, but their balance is insufficient.

ramenforbreakfast (approved):
LGTM besides some test suite nits.

apps/smart-contracts/token/test/PPO.test.ts
XavierEkkel: nit:
```suggestion
    it('reverts if amount > allowance', async () => {
```
to match order in revert message

apps/smart-contracts/token/test/PPO.test.ts
XavierEkkel: nit:
```suggestion
    it('reverts if amount > balance', async () => {
```
don't need the `amount < allowance` part - that's implied already since we've already tested the revert cases above

apps/smart-contracts/token/test/PPO.test.ts
```diff
@@ -197,4 +198,109 @@ describe('=> PPO', () => {
         .withArgs(ZERO_ADDRESS, user1.address, 1)
     })
   })
+
+  describe('# burn', () => {
```
XavierEkkel, line 202: for both functions, add tests around events being emitted, including in the 0-amount case

XavierEkkel, line 202: for both functions, add a test around burning resulting in the total supply being updated
and tests that total supply doesn't update when 0-amount

XavierEkkel (approved):
LGTM after comments and @ramenforbreakfast's comments

PR #221, opened 2022-07-17
feat: initialized Allowlist and added relavant tests

Nilay27:
* initialized Allowlist contract with nominee set in the constructor.
* Also added `_sourceIndex` and `_destinationIndex`, which will be used when resetting the source and destination allowed lists by incrementing the index.
* Following nested mappings are used to keep track of the source and destination lists:
 `mapping(uint256 => mapping(address => bool)) private _sourceIndexToAccountToAllowed;`
  `mapping(uint256 => mapping(address => bool)) private _destinationIndexToAccountToAllowed;`

### DIFF
--- a/apps/smart-contracts/token/contracts/Allowlist.sol
+++ b/apps/smart-contracts/token/contracts/Allowlist.sol
@@ -5,7 +5,14 @@ import "./interfaces/IAllowlist.sol";
 import "prepo-shared-contracts/contracts/SafeOwnable.sol";
 
 contract Allowlist is IAllowlist, SafeOwnable {
-  constructor(address _nominatedOwner) {}
+  uint256 private _sourceAccountsIndex;
+  uint256 private _destinationAccountsIndex;
+  mapping(uint256 => mapping(address => bool)) private _allowedSourceAccounts;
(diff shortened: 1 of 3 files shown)


apps/smart-contracts/token/contracts/Allowlist.sol
```diff
@@ -22,4 +29,20 @@ contract Allowlist is IAllowlist, SafeOwnable {
...
+
+  function getDestinationAccountsIndex() external view override returns (uint256) {
+    return _destinationAccountsIndex;
+  }
+
+  function isAccountSourceAllowed(address _account) external view override returns (bool) {
+    return _allowedSourceAccounts[_sourceAccountsIndex][_account];
+  }
+
+  function isAccountDestinationAllowed(address _account) external view override returns (bool) {
```
ramenforbreakfast, line 45:
```suggestion
  function isDestinationAllowed(address _account) external view override returns (bool) {
```

No need for account since account is already implied by the parameter.

ramenforbreakfast (approved):
LGTM, besides function naming nits.

ramenforbreakfast:
closing due to decision to switch over to Accesslist

PR #225, opened 2022-07-18
feat: added base contract with set and reset unimplemented

ramenforbreakfast:


apps/smart-contracts/token/contracts/AccountList.sol
```diff
@@ -0,0 +1,22 @@
...
+
+  constructor(address _nominatedOwner) {
+    transferOwnership(_nominatedOwner);
+  }
+
+  function set(address[] calldata accounts, bool[] calldata included) external override onlyOwner {}
+
+  function reset(address[] calldata _newIncludedAccounts) external override onlyOwner {}
+
+  function isIncludedAccount(address _account) external view override returns (bool) {
```
XavierEkkel, line 19:
```suggestion
  function isIncluded(address _account) external view override returns (bool) {
```
my bad, the `Account` is redundant because it's already the argument input

apps/smart-contracts/token/contracts/AccountList.sol
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+pragma solidity =0.8.7;
+
+import "./interfaces/IAccountList.sol";
+import "prepo-shared-contracts/contracts/SafeOwnable.sol";
+
+contract AccountList is IAccountList, SafeOwnable {
+  uint256 private _listIndex;
+  mapping(uint256 => mapping(address => bool)) private _indexToIncludedAccounts;
```
XavierEkkel, line 9:
```suggestion
  uint256 private _resetIndex;
  mapping(uint256 => mapping(address => bool)) private _resetIndexToAccountToIncluded;
```

PR #226, opened 2022-07-18
feat: added set and reset implementations

ramenforbreakfast:
Tests were taken from the old `Blocklist.test.ts`, except with a modification to the `sets lattermost bool value...` test.

The old test starts with the account not being included, and the latter most value is also false, so there will effectively be no change. I have made it start with the account to be included so there is a change to be reflected.


XavierEkkel (approved):
LGTM

PR #227, opened 2022-07-19
feat: added initial RestrictedTransferHook contract with setters

ramenforbreakfast:
This submission does not include any logic for the `hook` itself, just the setters.


apps/smart-contracts/token/test/RestrictedTransferHook.test.ts
```diff
@@ -0,0 +1,225 @@
...
+    await deployHook()
+    await restrictedTransferHook.connect(owner).acceptOwnership()
+    await restrictedTransferHook.connect(owner).setPPO(ppoToken.address)
+    allowedSources = await smockAccountListFixture(owner.address)
+    await restrictedTransferHook.connect(owner).setAllowedSources(allowedSources.address)
+    allowedDestinations = await smockAccountListFixture(owner.address)
+    await restrictedTransferHook.connect(owner).setAllowedDestinations(allowedDestinations.address)
+    blockedAccounts = await smockAccountListFixture(owner.address)
+    await restrictedTransferHook.connect(owner).setBlockedAccounts(blockedAccounts.address)
+  }
```
Nilay27, line 37: As a generic comment, we should ideally break `setupHookAndLists` into separate separate `setupHook` and `setupLists`. The reason is that apart from tests for `hook`, none of the other tests requires the lists to be set up.

Because they are being set up together, all of the tests become painfully slow.  All the tests for `setPPO`, `setAllowedSources`, `setAllowedDestinations` and `setBlockedAccounts` are taking a long time to run as they are setting up the lists in the `beforeEach` block. 

I am going to make changes to the tests and break them into the following parts:
```
  const setupHook = async (): Promise<void> => {
    await deployHook()
    await restrictedTransferHook.connect(owner).acceptOwnership()
  }

  const setupLists = async (): Promise<void> => {
    allowedSources = await smockAccountListFixture(owner.address)
    await restrictedTransferHook.connect(owner).setAllowedSources(allowedSources.address)
    allowedDestinations = await smockAccountListFixture(owner.address)
    await restrictedTransferHook.connect(owner).setAllowedDestinations(allowedDestinations.address)
    blockedAccounts = await smockAccountListFixture(owner.address)
    await restrictedTransferHook.connect(owner).setBlockedAccounts(blockedAccounts.address)
  }
```

Nilay27:
As a generic comment, we should ideally break `setupHookAndLists` into separate separate `setupHook` and `setupLists`. The reason is that apart from tests for `hook`, none of the other tests requires the lists to be set up.

Because they are being set up together, all of the tests become painfully slow.  All the tests for `setPPO`, `setAllowedSources`, `setAllowedDestinations` and `setBlockedAccounts` are taking a long time to run as they are setting up the lists in the `beforeEach` block. 

I am going to make changes to the tests and break them into the following parts:
```
  const setupHook = async (): Promise<void> => {
    await deployHook()
    await restrictedTransferHook.connect(owner).acceptOwnership()
  }

  const setupLists = async (): Promise<void> => {
    allowedSources = await smockAccountListFixture(owner.address)
    await restrictedTransferHook.connect(owner).setAllowedSources(allowedSources.address)
    allowedDestinations = await smockAccountListFixture(owner.address)
    await restrictedTransferHook.connect(owner).setAllowedDestinations(allowedDestinations.address)
    blockedAccounts = await smockAccountListFixture(owner.address)
    await restrictedTransferHook.connect(owner).setBlockedAccounts(blockedAccounts.address)
  }
```

Nilay27 (approved):
LGTM, apart from setting up lists, I will proceed to change them.

apps/smart-contracts/token/contracts/RestrictedTransferHook.sol
```diff
@@ -0,0 +1,60 @@
...
+
+  function setBlockedAccounts(IAccountList _newBlockedAccounts) external onlyOwner {
+    _blockedAccounts = _newBlockedAccounts;
+  }
+
+  function hook(
+    address _from,
+    address _to,
+    uint256 _amount
+  ) external override onlyPPO {}
```
XavierEkkel, line 43: the `hook` isn't going to change state right?
if it's not changing state then we can make it `view` and delete the `onlyPPO`

apps/smart-contracts/token/contracts/RestrictedTransferHook.sol
```diff
@@ -0,0 +1,60 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity =0.8.7;
+
+import "./interfaces/ITransferHook.sol";
+import "prepo-shared-contracts/contracts/SafeOwnable.sol";
+import "./interfaces/IAccountList.sol";
+
+contract RestrictedTransferHook is ITransferHook, SafeOwnable {
+  address private _ppo;
```
XavierEkkel, line 9: why do we need this?

apps/smart-contracts/token/contracts/RestrictedTransferHook.sol
```diff
@@ -0,0 +1,60 @@
...
+  IAccountList private _allowedSources;
+  IAccountList private _allowedDestinations;
+  IAccountList private _blockedAccounts;
```
XavierEkkel, line 12:
```suggestion
  IAccountList private _blocklist;
  IAccountList private _sourceAllowlist;
  IAccountList private _destinationAllowlist;
```
since the interface is `AccountList` this naming makes more sense.
and rearranged the order to match the order they are considered in.

apps/smart-contracts/token/contracts/RestrictedTransferHook.sol
ramenforbreakfast: @XavierEkkel I have left this open because there is the possibility that we might want the hook to call to change state in an external contract, similar to how the hooks worked in ACQR with `Moat`. I do not see any downside to keeping it ambiguous and flexible.

If you still disagree with this decision, I will change it.

apps/smart-contracts/token/contracts/RestrictedTransferHook.sol
ramenforbreakfast: I figured based on prior naming comments, since this is already a list, there was no need to attach the name "list" to every one of them. But I will make this change.

apps/smart-contracts/token/contracts/RestrictedTransferHook.sol
ramenforbreakfast: This is for `onlyPPO`, and dependent on the comment above.

ramenforbreakfast:
> As a generic comment, we should ideally break `setupHookAndLists` into separate separate `setupHook` and `setupLists`. The reason is that apart from tests for `hook`, none of the other tests requires the lists to be set up.
> 
> Because they are being set up together, all of the tests become painfully slow. All the tests for `setPPO`, `setAllowedSources`, `setAllowedDestinations` and `setBlockedAccounts` are taking a long time to run as they are setting up the lists in the `beforeEach` block.
> 
> I am going to make changes to the tests and break them into the following parts:
> 
> ```
>   const setupHook = async (): Promise<void> => {
>     await deployHook()
>     await restrictedTransferHook.connect(owner).acceptOwnership()
>   }
> 
>   const setupLists = async (): Promise<void> => {
>     allowedSources = await smockAccountListFixture(owner.address)
>     await restrictedTransferHook.connect(owner).setAllowedSources(allowedSources.address)
>     allowedDestinations = await smockAccountListFixture(owner.address)
>     await restrictedTransferHook.connect(owner).setAllowedDestinations(allowedDestinations.address)
>     blockedAccounts = await smockAccountListFixture(owner.address)
>     await restrictedTransferHook.connect(owner).setBlockedAccounts(blockedAccounts.address)
>   }
> ```

Yup, I completely agree. I suggested this in your prior PR when we were debating whether to separate out the `deploy` and `setup`, but perhaps because there was only one kind of setup, it was harder to argue against separating them out. 

Here is a perfect example where there are multiple testcases with different kinds of setup, which is why we need to separate out distinct kinds of setup operations, even if it results in an additional line per `beforeEach` call.

I originally batched all these operations together in a single `setup` function to save on one line per `beforeEach` call, even though some of the testcases did not need the extra setup.

apps/smart-contracts/token/contracts/RestrictedTransferHook.sol
XavierEkkel: that makes sense. can keep this then 👍

XavierEkkel:
> > As a generic comment, we should ideally break `setupHookAndLists` into separate separate `setupHook` and `setupLists`. The reason is that apart from tests for `hook`, none of the other tests requires the lists to be set up.
> > Because they are being set up together, all of the tests become painfully slow. All the tests for `setPPO`, `setAllowedSources`, `setAllowedDestinations` and `setBlockedAccounts` are taking a long time to run as they are setting up the lists in the `beforeEach` block.
> > I am going to make changes to the tests and break them into the following parts:
> > ```
> >   const setupHook = async (): Promise<void> => {
> >     await deployHook()
> >     await restrictedTransferHook.connect(owner).acceptOwnership()
> >   }
> > 
> >   const setupLists = async (): Promise<void> => {
> >     allowedSources = await smockAccountListFixture(owner.address)
> >     await restrictedTransferHook.connect(owner).setAllowedSources(allowedSources.address)
> >     allowedDestinations = await smockAccountListFixture(owner.address)
> >     await restrictedTransferHook.connect(owner).setAllowedDestinations(allowedDestinations.address)
> >     blockedAccounts = await smockAccountListFixture(owner.address)
> >     await restrictedTransferHook.connect(owner).setBlockedAccounts(blockedAccounts.address)
> >   }
> > ```
> 
> This was brought up previously in a prior PR and I don't think we reached a satisfactory conclusion, let me try to bring up an example to help bring about a conclusion for how we go about this going forward.
> 
> Some testcases require some `setupX`(**not deployment**), and some require an additional `setupY`. The additional `setupY` doesn't break the testcases that just need `setupX`.
> 
> Do we:
> 
> 1. Combine `setupX` and `setupY` into one function `setupXY` and just call it for every testcase? This will save one line per `beforeEach` call since we don't need to call both `setupX` and `setupY` for the testcases that need both X and Y. However, this means that testcases that just need X take additional time to setup since they perform the setup for Y as well. **This is what I originally went with**, however, I don't think this is the best way and went with this due to the simplicity of this test suite, but I think it's better we come up with a philosophy we apply to all test suites rather than ad-hoc deciding different setup styles per test suite.
> 2. Keep `setupX` and `setupY` as separate functions, only calling `setupX` in the `beforeEach` for testcases that just need X, and calling both `setupX` and `setupY` for those that need X and Y. **(This is what @Nilay27 has done above).**
> 3. Keep `setupX`, and then make a `setupXY` that calls `setupX` and then defines additional steps `Y`. This makes it so that testcases that need both X and Y still only need one line to setup in `beforeEach`, but means there is no standalone `setupY` function that can be used by a testcase that doesn't need X, but needs Y.
> 
> I suspect @XavierEkkel would prefer number 3, so @Nilay27 I think it best you have a `setupHook`, and `setupHookAndLists`, that calls `setupHook` within it, so that the tests that need both `setupHook` and `setupLists` only need one function call in their `beforeEach`.
> 
> I personally prefer number 2, since it doesn't assume that test case setups will always work in a way where some need X, X and Y, X Y and Z, etc.. some might only need Y, Y and Z, X and Z, etc..

Definitely not number 1.
Number 2 is okay, but if we know that Y is always called after X, then it is smarter to just have XY (number 3).

apps/smart-contracts/token/contracts/RestrictedTransferHook.sol
```diff
@@ -0,0 +1,60 @@
...
+    require(msg.sender == _ppo, "Only PPO can call hook");
+    _;
+  }
+
+  constructor(address _nominatedOwner) {
+    transferOwnership(_nominatedOwner);
+  }
+
+  function setPPO(address _newPPO) external onlyOwner {
+    _ppo = _newPPO;
```
XavierEkkel, line 24: might as well make this more generic
```suggestion
  function setToken(address _newToken) external onlyOwner {
    _token = _newToken;
```

apps/smart-contracts/token/contracts/RestrictedTransferHook.sol
```diff
@@ -0,0 +1,60 @@
...
+import "./interfaces/IAccountList.sol";
+
+contract RestrictedTransferHook is ITransferHook, SafeOwnable {
+  address private _ppo;
+  IAccountList private _allowedSources;
+  IAccountList private _allowedDestinations;
+  IAccountList private _blockedAccounts;
+
+  modifier onlyPPO() {
+    require(msg.sender == _ppo, "Only PPO can call hook");
```
XavierEkkel, line 15:
```suggestion
  modifier onlyToken() {
    require(msg.sender == _token, "msg.sender != token");
```
make more generic
and remember our guide about revert messages: "Revert messages should describe why the error occurred, not describe a solution"

apps/smart-contracts/token/test/RestrictedTransferHook.test.ts
```diff
@@ -0,0 +1,247 @@
...
+    it('sets allowed sources list to zero address', async () => {
+      expect(await restrictedTransferHook.getAllowedSources()).to.eq(ZERO_ADDRESS)
+    })
+
+    it('sets allowed destinations list to zero address', async () => {
+      expect(await restrictedTransferHook.getAllowedDestinations()).to.eq(ZERO_ADDRESS)
+    })
+  })
+
+  describe('setPPO', () => {
```
XavierEkkel, line 72:
```suggestion
  describe('setToken', () => {
```

XavierEkkel (approved):
LGTM after suggestions

apps/smart-contracts/token/contracts/RestrictedTransferHook.sol
```diff
@@ -0,0 +1,60 @@
...
+
+  constructor(address _nominatedOwner) {
+    transferOwnership(_nominatedOwner);
+  }
+
+  function setPPO(address _newPPO) external onlyOwner {
+    _ppo = _newPPO;
+  }
+
+  function setAllowedSources(IAccountList _newAllowedSources) external onlyOwner {
```
XavierEkkel, line 27: add TODO to define an interface file for RestrictedTransferHook and then add `override`

PR #228, opened 2022-07-19
feat: adding logic for restricted transfer hook

Nilay27:
* Adding tests for `hook` implementation done by @ramenforbreakfast.
* The `hook` checks for the criteria that both `source` and `destination` should not be in the blocklist, and either `source` or `destination` should be allowlists to be able to transfer.

### DIFF
--- a/apps/smart-contracts/token/contracts/RestrictedTransferHook.sol
+++ b/apps/smart-contracts/token/contracts/RestrictedTransferHook.sol
@@ -40,7 +40,14 @@ contract RestrictedTransferHook is ITransferHook, SafeOwnable {
     address _from,
     address _to,
     uint256 _amount
-  ) external override onlyPPO {}
+  ) external override onlyPPO {
+    require(
+      !_blockedAccounts.isIncluded(_from) &&
+        !_blockedAccounts.isIncluded(_to) &&
+        (_allowedDestinations.isIncluded(_to) || _allowedSources.isIncluded(_from)),
+      "Account blocked"
+    );
+  }
 
   function getPPO() external view returns (address) {
     return _ppo;
--- a/apps/smart-contracts/token/contracts/RestrictedTransferHook.sol
+++ b/apps/smart-contracts/token/contracts/RestrictedTransferHook.sol
@@ -1,0 +1,60 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity =0.8.7;
+
+import "./interfaces/ITransferHook.sol";
+import "prepo-shared-contracts/contracts/SafeOwnable.sol";
+import "./interfaces/IAccountList.sol";
+
+contract RestrictedTransferHook is ITransferHook, SafeOwnable {
+  address private _ppo;
+  IAccountList private _allowedSources;
+  IAccountList private _allowedDestinations;
+  IAccountList private _blockedAccounts;
+
+  modifier onlyPPO() {
+    require(msg.sender == _ppo, "Only PPO can call hook");
+    _;
+  }
+
+  constructor(address _nominatedOwner) {
+    transferOwnership(_nominatedOwner);
+  }
+
+  function setPPO(address _newPPO) external onlyOwner {
+    _ppo = _newPPO;
+  }
+
+  function setAllowedSources(IAccountList _newAllowedSources) external onlyOwner {
+    _allowedSources = _newAllowedSources;
+  }
+
+  function setAllowedDestinations(IAccountList _newAllowedDestinations) external onlyOwner {
+    _allowedDestinations = _newAllowedDestinations;
+  }
+
+  function setBlockedAccounts(IAccountList _newBlockedAccounts) external onlyOwner {
+    _blockedAccounts = _newBlockedAccounts;
+  }
+
+  function hook(
+    address _from,
+    address _to,
+    uint256 _amount
+  ) external override onlyPPO {}
+
+  function getPPO() external view returns (address) {
+    return _ppo;
+  }
+
+  function getAllowedSources() external view returns (IAccountList) {
+    return _allowedSources;
+  }
+
+  function getAllowedDestinations() external view returns (IAccountList) {
+    return _allowedDestinations;
+  }
+
+  function getBlockedAccounts() external view returns (IAccountList) {
+    return _blockedAccounts;
+  }
+}
--- a/apps/smart-contracts/token/contracts/RestrictedTransferHook.sol
+++ b/apps/smart-contracts/token/contracts/RestrictedTransferHook.sol
@@ -35,7 +35,7 @@ contract RestrictedTransferHook is IRestrictedTransferHook, SafeOwnable {
     require(!_blocklist.isIncluded(_from), "Sender blocked");
     require(!_blocklist.isIncluded(_to), "Recipient blocked");
     if (_sourceAllowlist.isIncluded(_from)) return;
-    require(_destinationAllowlist.isIncluded(_to), "Destination blocked");
+    require(_destinationAllowlist.isIncluded(_to), "Destination not allowed");
   }
 
   function setToken(address _newToken) external onlyOwner {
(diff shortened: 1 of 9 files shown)


apps/smart-contracts/token/contracts/RestrictedTransferHook.sol
```diff
@@ -40,7 +40,14 @@ contract RestrictedTransferHook is ITransferHook, SafeOwnable {
...
     address _to,
     uint256 _amount
-  ) external override onlyPPO {}
+  ) external override onlyPPO {
+    require(
+      !_blockedAccounts.isIncluded(_from) &&
+        !_blockedAccounts.isIncluded(_to) &&
+        (_allowedDestinations.isIncluded(_to) || _allowedSources.isIncluded(_from)),
+      "Account blocked"
+    );
```
XavierEkkel, line 49: I think we should separate these into multiple require statements so that we can isolate the effects:
```suggestion
    require(!_blockedAccounts.isIncluded(_from), "Sender blocked");
    require(!_blockedAccounts.isIncluded(_to), "Recipient blocked");
    if (_allowedSources.isIncluded(_from)) return;
    require(_allowedDestinations.isIncluded(_to), "Destination blocked");
```

apps/smart-contracts/token/contracts/RestrictedTransferHook.sol
```diff
@@ -40,7 +40,14 @@ contract RestrictedTransferHook is ITransferHook, SafeOwnable {
     address _from,
     address _to,
     uint256 _amount
-  ) external override onlyPPO {}
+  ) external override onlyPPO {
```
XavierEkkel, line 43:
```suggestion
  ) external view override {
```

apps/smart-contracts/token/test/RestrictedTransferHook.test.ts
```diff
@@ -244,4 +245,216 @@ describe('RestrictedTransferHook', () => {
...
+
+    it('reverts if caller is not PPO', async () => {
+      expect(await restrictedTransferHook.getToken()).to.not.eq(user1.address)
+
+      await expect(
+        restrictedTransferHook.connect(user1).hook(source.address, destination.address, 1)
+      ).to.be.revertedWith('msg.sender != token')
+    })
+
+    describe('if source blocked', () => {
```
XavierEkkel, line 266:
```suggestion
    describe('if sender blocked', () => {
```
so we don't overlap in terminology

please also update the terminology in every test below so it specifically uses one of: sender, recipient, source, destination

apps/smart-contracts/token/test/RestrictedTransferHook.test.ts
```diff
@@ -244,4 +245,216 @@ describe('RestrictedTransferHook', () => {
...
+      it('reverts if destination allowed', async () => {
+        allowedDestinations.isIncluded.whenCalledWith(destination.address).returns(true)
+
+        await expect(
+          restrictedTransferHook.connect(ppoToken).hook(source.address, destination.address, 1)
+        ).to.be.revertedWith('Sender blocked')
+      })
+    })
+
+    describe('if destination blocked', () => {
```
XavierEkkel, line 304:
```suggestion
    describe('if recipient blocked', () => {
```

apps/smart-contracts/token/test/RestrictedTransferHook.test.ts
```diff
@@ -244,4 +245,216 @@ describe('RestrictedTransferHook', () => {
       expect(await restrictedTransferHook.getBlocklist()).to.eq(JUNK_ADDRESS)
     })
   })
+  describe('hook', () => {
+    let source: SignerWithAddress
+    let destination: SignerWithAddress
```
XavierEkkel, line 250: also add `sender` and `recipient` so that we don't confuse the terminology

XavierEkkel (commented):
almost LGTM, just need to use the correct terminology to avoid confusion

apps/smart-contracts/token/test/RestrictedTransferHook.test.ts
Nilay27: I do not understand why we need four names? A transfer will always be between two participants. 
Either we can call them `sender`, `recipient` or `source`, `destination`. 

The recipient/destination is going to be in `blocklist` (in this case, revert) or in `allowedDestinations`(in this case succeed) or both (in this case also, revert).

In order to avoid confusion of Recipient vs Destination, in the contract we can have:

```
    require(!_blockedAccounts.isIncluded(_from), "Sender blocked");
    require(!_blockedAccounts.isIncluded(_to), "Recipient blocked");
    if (_allowedSources.isIncluded(_from)) return;
    require(_allowedDestinations.isIncluded(_to), "Recipient not allowed");
```

or 

```
    require(!_blockedAccounts.isIncluded(_from), "Source blocked");
    require(!_blockedAccounts.isIncluded(_to), "Destination blocked");
    if (_allowedSources.isIncluded(_from)) return;
    require(_allowedDestinations.isIncluded(_to), "Destination not allowed");
```

Nilay27:
* Discussed the test naming with @ramenforbreakfast we decided to rename the tests in such a way that it avoids confusion of source/destination vs sender/recipient without needing 4 signers (source, destination, sender and recipient). 

The txn will always be between sender and recipient, but we differentiate b/w sender/recipient vs source/destination by the following:
`if sender is allowed source` and `if recipient is allowed destination`

apps/smart-contracts/token/test/RestrictedTransferHook.test.ts
```diff
@@ -244,4 +245,216 @@ describe('RestrictedTransferHook', () => {
...
+        ).to.be.revertedWith('Recipient blocked')
+      })
+
+      describe('if both sender and recipient not blocked', () => {
+        beforeEach(() => {
+          blockedAccounts.isIncluded.whenCalledWith(sender.address).returns(false)
+          blockedAccounts.isIncluded.whenCalledWith(recipient.address).returns(false)
+        })
+
+        it("doesn't revert if recipient is not allowed destination", async () => {
```
XavierEkkel, line 369: nit: `succeeds` instead of `doesn't revert` - same meaning but more concise

same for rest of file

apps/smart-contracts/token/test/RestrictedTransferHook.test.ts
```diff
@@ -244,4 +245,216 @@ describe('RestrictedTransferHook', () => {
...
+
+      it('reverts if recipient blocked', async () => {
+        blockedAccounts.isIncluded.whenCalledWith(recipient.address).returns(true)
+
+        await expect(
+          restrictedTransferHook.connect(ppoToken).hook(sender.address, recipient.address, 1)
+        ).to.be.revertedWith('Recipient blocked')
+      })
+
+      describe('if both sender and recipient not blocked', () => {
```
XavierEkkel, line 363: i dont think we need this additional nesting.
as with all our other test files, the revert conditions that came prior are implied to be avoided so we don't have to mention the opposite of the revert conditions explicitly.
as for the beforeEach below, returning false for a blocklist is the expected default, so we should just move that to the beforeEach above.

same for the other nested describe below.

XavierEkkel (approved):
LGTM after comments!

apps/smart-contracts/token/test/RestrictedTransferHook.test.ts
```diff
@@ -244,4 +245,216 @@ describe('RestrictedTransferHook', () => {
...
     })
   })
+  describe('hook', () => {
+    let sender: SignerWithAddress
+    let recipient: SignerWithAddress
+    beforeEach(async () => {
+      await setupHookAndLists()
+      await restrictedTransferHook.connect(owner).setToken(ppoToken.address)
+      sender = user1
+      recipient = user2
```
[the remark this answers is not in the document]
ramenforbreakfast: can't do this since the `user1` and `user2` are signers that are reinitialized every single time along with the contract stack, and thus are technically different accounts every time.

apps/smart-contracts/token/test/RestrictedTransferHook.test.ts
ramenforbreakfast: We have reorganized into a cleaner 4 `describe` block format, with no additional nesting.

PR #231, opened 2022-07-20
fix: added hardhat-deploy to token folder's hardhat config

ramenforbreakfast:
While writing out a deployment process for our `PPO` token and trying to integrate `proposeUpgrade` since we want upgrades to be approved by governance, I encountered a problem where we were not importing the `hardhat-deploy` library for our `token` project.

In addition, I took this opportunity to sort the imports, the linter seems to sort the imports itself, but the two files were so divergent in their order that its sorting did not result in the same order for both `hardhat.config.ts` files. I have sorted them so that imports are in the same order for both.


XavierEkkel (approved):
LGTM

PR #202, opened 2022-07-07
feat: changes for deployment of PrePOMarketFactory on polygon

chrisling-dev:
https://www.notion.so/FE-Project-Management-d3878c29e9534234a50f4f8838525bee?v=177bee72a49349b7b5af7c370371389e&p=589fd4716aac4ad2bd417ca05bdb925b

PR #232, opened 2022-07-20
fix: assertIsTestnetChain needs a number instead of ChainId

ramenforbreakfast:
Another thing found while building our new deployment process for the `PPO` token. 
`assertIsTestnetChain` was refactored when the team switched over to the new `prepo-constants` library to share definitions for networks and chainIds across the whole project.

However, the way the `Networks` `enum` works, `.include` will only succeed if a `number` is passed in for some reason. Even though the previous way was passing in a `ChainId` type, this does not seem to correctly detect that the passed in `ChainId` is included in the list of testnet `ChainId`'s. The function will always fail even if a testnet chainId is passed in.

I unfortunately did not detect this issue when we refactored the entire repo with `prepo-constants`.

Instead of using the `ChainId` type defined in `prepo-constants`, I have decided to stick with the `string` hardhat returns when polling the chainId, and converting that to a `number` type, since this function will only be used by the SC team with the `hardhat` library.


apps/smart-contracts/core/tsconfig.json
```diff
@@ -1,4 +1,4 @@
 {
   "extends": "tsconfig/smart-contracts.json",
-  "include": ["./hardhat.config.ts"]
+  "include": ["./hardhat.config.ts", "deploy/*"]
```
XavierEkkel, line 3: should we also update this for the `/token/tsconfig.json`?

XavierEkkel (approved):
LGTM after question

ramenforbreakfast:
Closing due to being also addressed by PR #202

PR #233, opened 2022-07-20
feat: init Staking ABI, stake implementation

georgesmithishere:
https://www.notion.so/eb96722b397d42c9908002ca06af88e3?v=1f519ba1849347aa9740faedda4acf3b&p=1dd09682d2914387a5e2384b30b35dba
once you approve it, i will mock Stake button again


apps/frontend/core-dapp/src/features/ppo/stake/StakeStore.ts
```diff
@@ -25,15 +25,15 @@ export class StakeStore {
...
   }
 
-  stake(): {
+  stake(): Promise<{
     success: boolean
     error?: string | undefined
-  } {
+  }> {
     const balance = this.getBalance()
     if (this.currentStakingValue > balance) {
```
chrisling-dev, line 33: Maybe we could update this code to just do this below, and delete the `getBalance` method, this looks really redundant, it's reassigning the same thing to different variables twice while there's essentially no complexity at all, (token is "assigned" to getBalance, then getBalance is called to assign to `balance`, when we could've just used `tokenBalance` directly) understand that it's not changes from this PR, maybe good to have it fixed since it's relevant

Also the existing code is somewhat wrong, ppoTokenStore cannot be undefined, and `tokenBalance` is deprecated

```suggestion
    const { tokenBalanceFormat } = this.root.ppoTokenStore
    if (this.currentStakingValue > +(tokenBalanceFormat ?? 0)) {
```

apps/frontend/core-dapp/src/features/ppo/stake/PPOStakingStore.ts
```diff
@@ -0,0 +1,133 @@
...
+    if (result === undefined) return undefined
+    return result[0]
+  }
+
+  async stake(amount: number): Promise<{ success: boolean; error?: string }> {
+    try {
+      runInAction(() => {
+        this.staking = true
+        this.stakingHash = undefined
+      })
```
chrisling-dev, line 66: Correct me if I'm wrong but we shouldn't need to wrap any state changes with runInAction until we some await occurred since the function is already configured as `action.bound`
```suggestion
        this.staking = true
        this.stakingHash = undefined
```

apps/frontend/core-dapp/src/features/ppo/stake/StakeUnstakeLayout.tsx
```diff
@@ -46,8 +46,9 @@ const StakeUnstakeLayout: React.FC<{
   const theme = useTheme()
   const {
     delegateStore: { selectedDelegate: delegate },
-    stakeStore: { isCurrentStakingValueValid },
+    stakeStore: { isCurrentStakingValueValid, stake },
     unstakeStore: { isCurrentUnstakingValueValid },
+    ppoStakingStore: { staking },
     uiStore: { disableMocks },
   } = useRootStore()
   const loading = delegate && !delegate.delegateAddress
```
chrisling-dev, line 54: loading should be true when it's staking

apps/frontend/core-dapp/src/features/ppo/stake/StakeUnstakeLayout.tsx
```diff
@@ -56,11 +57,10 @@ const StakeUnstakeLayout: React.FC<{
   const content = pageMap[tab]
 
   const buttonDisabled =
-    disableMocks ||
-    loading ||
-    !(isStake ? isCurrentStakingValueValid : isCurrentUnstakingValueValid)
+    staking || loading || !(isStake ? isCurrentStakingValueValid : isCurrentUnstakingValueValid)
```
chrisling-dev, line 60: Then staking here can be removed since loading will be true when staking is true
```suggestion
    loading || !(isStake ? isCurrentStakingValueValid : isCurrentUnstakingValueValid)
```

chrisling-dev (changes requested):
only few comments, other than that, my main concern is how do we hide these features (block the button) so users don't try this page and think it's broken, and how do we do that so that it doesn't block other PRs built on top of this?

apps/frontend/core-dapp/src/features/ppo/stake/PPOStakingStore.ts
georgesmithishere: I think you are correct, but have to investigate more 
https://github.com/mobxjs/mobx/blob/mobx4and5/docs/refguide/api.md#enforceactions
there's a cool rule - WDYT

apps/frontend/core-dapp/src/hooks/useFeatureFlag.ts
```diff
@@ -10,11 +11,19 @@ type UseFeatureFlag = {
...
   const body = {
     featureName,
     userAddress,
   }
+  if (featureName === FeatureFlag.enableStakingLocally) {
+    if (!window) {
+      return false
+    }
+    return window.origin.startsWith(LOCALHOST)
+  }
```
chrisling-dev, line 26: 👍

chrisling-dev (changes requested):
LGTM after comments are addressed

apps/frontend/core-dapp/src/features/ppo/stake/StakeStore.ts
```diff
@@ -13,8 +13,13 @@ export class StakeStore {
...
 
   get isCurrentStakingValueValid(): boolean {
-    const balance = this.root.ppoTokenStore.tokenBalance ?? 0
-    return Boolean(this.currentStakingValue) && this.currentStakingValue <= balance
+    const { tokenBalanceRaw } = this.root.ppoTokenStore
+    // gte = greater than or equal
+    return (
+      tokenBalanceRaw !== undefined &&
+      this.currentStakingValue > 0 &&
+      tokenBalanceRaw.gte(this.currentStakingValue)
```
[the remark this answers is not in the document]
georgesmithishere: thank you! this is great!

chrisling-dev (approved):
LGTM for mocking purposes

PR #238, opened 2022-07-21
feat: added generic token setter contract with tests

Nilay27:
* Adding files for the generic shared token contracts to be used by other contracts for setting tokens and using the inherited modifier `onlyToken`.
* To add a test for the modifier, I had to create a wrapper around `Token.sol`. The wrapper `TokenWrapper.sol` only has one function `testTokenModifier`, which can be called by `onlyToken`.
* The contract `Token.sol` itself inherits from `SafeOwnable` without any constructor arguments, so by default, the `nominee` for `Token.sol` will be `ZERO_ADDRESS`, and the `owner` will be `deployer`.  But when the contract is inherited by any other contract, the `nominee` and `owner` will be of the inheriting contracts.

### DIFF
--- a/packages/prepo-shared-contracts/contracts/Token.sol
+++ b/packages/prepo-shared-contracts/contracts/Token.sol
@@ -1,0 +1,24 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+pragma solidity =0.8.7;
+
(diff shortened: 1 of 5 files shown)


Nilay27:
Closing due to the decision of not using a generic setter contract for Transfer Hooks.

PR #239, opened 2022-07-21
feat: removing duplicate token setters from transfer hook

Nilay27:
* Switching transfer hook to inherit from the shared `Token` contract.
* Removed duplicated functions and related tests.


Nilay27:
* Closing due to the decision of not using a generic setter contract for Transfer Hooks

PR #241, opened 2022-07-21
feat: startCooldown & endCooldown Staking logic

georgesmithishere:
https://www.notion.so/eb96722b397d42c9908002ca06af88e3?v=1f519ba1849347aa9740faedda4acf3b&p=206eaa3446ba4d13979fda77a1d863c8

### DIFF
--- a/apps/frontend/core-dapp/src/features/ppo/stake/PPOStakingStore.ts
+++ b/apps/frontend/core-dapp/src/features/ppo/stake/PPOStakingStore.ts
@@ -1,3 +1,4 @@
+import { addDays } from 'date-fns'
 import { BigNumber } from 'ethers'
 import { action, makeObservable, observable, reaction, runInAction } from 'mobx'
 import { ContractReturn, Factory } from 'prepo-stores'
@@ -7,12 +8,19 @@ import { Erc20Store } from '../../../stores/entities/Erc20.entity'
 import { RootStore } from '../../../stores/RootStore'
 
 type Stake = PPOStakingAbi['functions']['stake']
+type StartCooldown = PPOStakingAbi['functions']['startCooldown']
 type BalanceData = PPOStakingAbi['functions']['balanceData']
 
 // use this key to store balance in localStorage, while there's no SC
 const MOCK_KEY = 'MOCK_BALANCE_DATA'
 
 const TOKEN_SYMBOL = 'PPO_STAKING'
+
+const MOCKED_CONFIG = {
+  cooldownPeriod: 21, // 21 days cooldown period
+  withdrawWindow: 7, // 7 days to withdraw - post-cooldown
+  fee: 10.12, // Fee will calculated based on how long staking time length
+}
 
 // Storing data in localStorage - will allow us to test withdraw/startCooldown functionality
 function getMockedBalance(address: string): BalanceStructOutput | undefined {
@@ -26,6 +34,11 @@ export class PPOStakingStore extends Erc20Store {
   staking = false
   stakingHash?: string
   mockRawBalance?: BigNumber
+  startingCooldown = false
+  startingCooldownHash?: string
+  endingCooldown = false
+  endingCooldownHash?: string
+
   constructor(root: RootStore) {
     super({ root, tokenName: TOKEN_SYMBOL, factory: PPOStakingAbi__factory as unknown as Factory })
     this.symbolOverride = TOKEN_SYMBOL
@@ -35,6 +48,8 @@ export class PPOStakingStore extends Erc20Store {
       stake: action.bound,
       mockRawBalance: observable,
       getBalanceData: observable,
+      startCooldown: action.bound,
+      endCooldown: action.bound,
     })
     this.subscribe()
   }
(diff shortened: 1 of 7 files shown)


r0x4xn:
App <strong>core-dapp</strong> has been deployed to Vercel.

<table>
  <tr>
    <td><strong>✅ Preview:</strong></td>
    <td><a href='https://core-dapp-l1ws771go-prepo.vercel.app'>https://core-dapp-l1ws771go-prepo.vercel.app</a></td>
  </tr>
</table>


<hidden purpose="for-rewritable-pr-comment-action-use" value="prj_JWVJ3WoV71UcOmB52qrcezjyBL3V"></hidden>

apps/frontend/core-dapp/src/features/ppo/stake/UnstakeStore.ts
```diff
@@ -1,18 +1,21 @@
...
   confirm = false
   currentUnstakingValue = TRANSACTION_SETTING.DEFAULT_AMOUNT
-  fee = FEE_MOCK
 
   get isCurrentUnstakingValueValid(): boolean {
-    return Boolean(this.currentUnstakingValue) // TODO: add staked value check
+    const stakedPPO = this.root.ppoStakingStore.balanceData?.raw
+    if (!stakedPPO) return false
+    return (
+      BigNumber.from(stakedPPO).toNumber() >= this.currentUnstakingValue &&
```
chrisling-dev, line 16: A better approach would be to have values like balanceData.raw converted to BigNumber inside getMockedBalance, so we don't have to do BigNumber.from everywhere, which isn't something we should do when actual SC is integrated.

Also it's kinda redundant to write BigNumber.from(anotherBigNumber)

```suggestion
      stakedPPO.toNumber() >= this.currentUnstakingValue &&
```

apps/frontend/core-dapp/src/features/ppo/stake/UnstakeStore.ts
```diff
@@ -26,4 +29,16 @@ export class UnstakeStore {
...
     this.currentUnstakingValue = validateNumber(+value) // TODO: compare with staked value from SC
   }
+
+  startCooldown(): Promise<{
+    success: boolean
+    error?: string | undefined
+  }> {
+    const { balanceData } = this.root.ppoStakingStore
+    // lt === less than
+    if (balanceData === undefined || balanceData.raw.lt(this.currentUnstakingValue)) {
```
chrisling-dev, line 39: By type this is correct, but if this code is executed it will fail. I think the main issue is in getMockBalance which is causing a bit of problem in many places. Highly recommend returning actual BigNumber values from getMockBalance instead of return from json object and cast to BigNumber type because methods like lt, gt, gte only exist in actual BigNumber values.

You could do something like
```
return {
   raw: BigNumber.from(dataFromLocalStorage)
}
```

Or even better, remove all local storage logic, just do this:

return {
   raw: BigNumber.from(1000),
   ...other values that were supposed to be returned according to BalanceStructOutput with hardcoded value
}

chrisling-dev (approved):
I think a lot of the code won't even be close to actual implementation (e.g. BigNumber.from(someBigNumberValue)) but has to be this way in this mocked feature. 

I'll approve for now since it won't be user facing although many things might break if we try to "force" the code to actual implementation. There's still a lot of things in this PR worth preserving tho.

Please ignore my comments for now so when we do actual implementation I can refer to the problems I found that will only apply to actual implementation

apps/frontend/core-dapp/src/features/ppo/stake/PPOStakingStore.ts
```diff
@@ -116,6 +169,118 @@ export class PPOStakingStore extends Erc20Store {
...
+
+      const { address } = this.root.web3Store.signerState
+      if (!address) {
+        return {
+          success: false,
+          error: 'wallet is not connected',
+        }
+      }
+      // const { hash, wait } = await this.sendTransaction<StartCooldown>('startCooldown', [address, amount])
+      // Once SC is implemented uncomment above line, and remove lines below
```
chrisling-dev, line 185: Add TODO: infront of comment

apps/frontend/core-dapp/src/features/ppo/stake/PPOStakingStore.ts
```diff
@@ -116,6 +169,118 @@ export class PPOStakingStore extends Erc20Store {
...
+
+      const { address } = this.root.web3Store.signerState
+      if (!address) {
+        return {
+          success: false,
+          error: 'wallet is not connected',
+        }
+      }
+      // const { hash, wait } = await this.sendTransaction<EndCooldown>('endCooldown', [])
+      // Once SC is implemented uncomment above line, and remove lines below
```
chrisling-dev, line 243: ditto

chrisling-dev (changes requested):
will review again after test is fixed, can't check latest preview now

apps/frontend/core-dapp/src/features/ppo/stake/UnstakeStore.ts
```diff
@@ -24,6 +30,22 @@ export class UnstakeStore {
...
+    error?: string | undefined
+  }> {
+    const { balanceData } = this.root.ppoStakingStore
+    if (balanceData === undefined || balanceData.raw.lt(this.currentUnstakingValue)) {
+      return Promise.resolve({ success: false })
+    }
+    return this.root.ppoStakingStore.startCooldown(this.currentUnstakingValue)
+  }
+
+  private getBalance(): number {
```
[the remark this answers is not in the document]
georgesmithishere: merge error, we don't even need this method

chrisling-dev (approved):
Great job! LGTM!

PR #242, opened 2022-07-22
feat: adding shared Pausable with setters

Nilay27:
* Adding files for the generic shared pausable contracts to be used by other contracts for setting `paused` and using the inherited modifier `whenNotPaused`.
* To add a test for the modifier, I had to create a wrapper around `Pausable.sol`. The wrapper `PausableWrapper.sol` only has one function, `testWhenNotPaused` with the modifier `whenNotPaused`.
* The contract `Pausable.sol` itself inherits from `SafeOwnable` without any constructor arguments, so by default, the `nominee` for `Pausable.sol` will be `ZERO_ADDRESS`, and the `owner` will be `deployer`.  But when the contract is inherited by any other contract, the `nominee` and `owner` will be of the inheriting contracts.

### DIFF
--- a/packages/prepo-shared-contracts/test/Pausable.test.ts
+++ b/packages/prepo-shared-contracts/test/Pausable.test.ts
@@ -1,0 +1,109 @@
+import { expect } from 'chai'
+import { ethers } from 'hardhat'
+import { SignerWithAddress } from '@nomiclabs/hardhat-ethers/dist/src/signer-with-address'
+import { ZERO_ADDRESS, JUNK_ADDRESS } from 'prepo-constants'
+import { pausableWrapperFixture } from './fixtures/PausableWrapperFixture'
+import { PausableWrapper } from '../types/generated'
+
+describe('Token', () => {
+  let deployer: SignerWithAddress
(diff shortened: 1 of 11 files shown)


packages/prepo-shared-contracts/test/Pausable.test.ts
```diff
@@ -0,0 +1,109 @@
+import { expect } from 'chai'
+import { ethers } from 'hardhat'
+import { SignerWithAddress } from '@nomiclabs/hardhat-ethers/dist/src/signer-with-address'
+import { ZERO_ADDRESS, JUNK_ADDRESS } from 'prepo-constants'
+import { pausableWrapperFixture } from './fixtures/PausableWrapperFixture'
+import { PausableWrapper } from '../types/generated'
+
+describe('Token', () => {
```
XavierEkkel, line 8:
```suggestion
describe('Pausable', () => {
```

XavierEkkel (approved):
LGTM after nit

ramenforbreakfast (approved):
LGTM!

PR #244, opened 2022-07-22
feat: switching to shared pausable contract

Nilay27:
* Switching to shared `Pausable.sol` for token contracts.
* Also removed tests for `setPaused()`


XavierEkkel (approved):
LGTM

ramenforbreakfast (approved):
LGTM, I also went ahead and removed `revertReason` and usage of the old chai matchers library.

PR #248, opened 2022-07-24
feat: adding modification of beforeTokenTransfer

Nilay27:
* Adding modification to `_beforeTokenTransfer` to include a call to transfer hook.
* The hook called depends on what we have set in the `setTransferHook`
* Since `_beforeTokenTransfer` is also called if `transfer` and `transferFrom` are called. Therefore, I have added tests for `transfer` and `transferFrom` to check for intended behaviours.
* I have added tests for `transfer` and `transferFrom` in this PR only as they are fairly similar to `burn` and `burnFrom`

**Note:** 

I had to use import `chai` and use `chai.use(smock.matchers)` because otherwise I was getting the error (`Error: Invalid Chai property: calledWith`)  in trying to test for revert for hooks. 

This was also happening if I used `expect(mockBlocklistTransferHook.hook).to.have.been.called` instead of
`expect(mockBlocklistTransferHook.hook).to.have.been.calledWith(ZERO_ADDRESS, user1.address, 1)`.


ramenforbreakfast:
> * Adding modification to `_beforeTokenTransfer` to include a call to transfer hook.
> * The hook called depends on what we have set in the `setTransferHook`
> * Since `_beforeTokenTransfer` is also called if `transfer` and `transferFrom` are called. Therefore, I have added tests for `transfer` and `transferFrom` to check for intended behaviours.
> * I have added tests for `transfer` and `transferFrom` in this PR only as they are fairly similar to `burn` and `burnFrom`
> 
> **Note:**
> 
> I had to use import `chai` and use `chai.use(smock.matchers)` because otherwise I was getting the error (`Error: Invalid Chai property: calledWith`) in trying to test for revert for hooks.
> 
> This was also happening if I used `expect(mockBlocklistTransferHook.hook).to.not.have.been.called` instead of `expect(mockBlocklistTransferHook.hook).to.have.been.calledWith(ZERO_ADDRESS, user1.address, 1)`.

Yeah, if a suite needs smock contracts, it needs to use the custom smock matchers, i'll add this clarification to the manual

apps/smart-contracts/token/test/PPO.test.ts
```diff
@@ -108,7 +121,33 @@ describe('=> PPO', () => {
...
+    })
+
+    it('reverts if transfer hook not set', async () => {
+      await ppo.connect(owner).setTransferHook(ZERO_ADDRESS)
+
+      await expect(ppo.connect(owner).mint(user1.address, 1)).revertedWith('Transfer hook not set')
+    })
+
+    it('reverts if blocklist transfer hook reverts', async () => {
+      mockBlocklistTransferHook.hook.reverts()
```
ramenforbreakfast, line 134: Just going to clarify here for other reviewers. The reason why @Nilay27 is mocking a generic revert rather than one with a specific message is that specific message revert mocking doesn't work and the team has already spent a significant amount of time investigating this problem.

apps/smart-contracts/token/test/PPO.test.ts
```diff
@@ -108,7 +121,33 @@ describe('=> PPO', () => {
...
+
+      await expect(ppo.connect(owner).mint(user1.address, 1)).to.be.reverted
+      expect(mockBlocklistTransferHook.hook).to.have.been.calledWith(ZERO_ADDRESS, user1.address, 1)
```
ramenforbreakfast, line 137: Verification that the hook has been called with the correct parameters should be in a separate testcase.

`calls configured hook with correct parameters`

Also I don't think we need testcases for both `RestrictedTransferHook` and `UnrestrictedTransferHook` since this would imply we'd need to add a testcase for every `TransferHook`. We should simply test the `TransferHook` itself in its own tests and only include in this test suite that the configured hook is called correctly, regardless of the hook configured.

We just need to ensure that any `TransferHook` is called with the right parameters and that reverts propagate upwards.

apps/smart-contracts/token/test/PPO.test.ts
```diff
@@ -201,10 +240,36 @@ describe('=> PPO', () => {
...
+      await ppo.connect(owner).setTransferHook(ZERO_ADDRESS)
+
+      await expect(ppo.connect(user1).burn(1)).revertedWith('Transfer hook not set')
+    })
+
+    it('reverts if blocklist transfer hook reverts', async () => {
+      mockBlocklistTransferHook.hook.reverts()
+
+      await expect(ppo.connect(user1).burn(1)).to.be.reverted
+      expect(mockBlocklistTransferHook.hook).to.have.been.calledWith(user1.address, ZERO_ADDRESS, 1)
```
ramenforbreakfast, line 257: Same things I mentioned in the `mint` test suite apply here.

apps/smart-contracts/token/test/PPO.test.ts
```diff
@@ -361,4 +457,236 @@ describe('=> PPO', () => {
         .withArgs(user1.address, ZERO_ADDRESS, 1)
     })
   })
+
+  describe('# transfer', () => {
```
ramenforbreakfast, line 461: @Nilay27 I think all this effort to re-test ERC20 functions we already know work is a perfect example of why we might still want to use wrapper test contracts.

Previously, I would have just exposed `_beforeTokenTransfer` via an `external` function in a testing contract, which would remove the need to have all these redundant tests.

@XavierEkkel I will have you chime in here on whether or not you think this is an acceptable case to use a wrapper contract for testing purposes (we used to call them "Mock" contracts, but thats not really an accurate name). 

By using a wrapper test contract, we would only need two tests to ensure ` _beforeTokenTransfer` reverts when the hook reverts, and that it calls the hook with the right parameters, rather than adding tests for every single function that calls `_beforeTokenTransfer` because `_beforeTokenTransfer` is not externally exposed in the production contract.

ramenforbreakfast (commented):
Holding off on approval due to concern with unnecessary test bloat.

apps/smart-contracts/token/test/PPO.test.ts
XavierEkkel: I agree that we don't need to fully re-test ERC20 functions.
However, I don't think testing functions that expose private/internal functions (in this case `_beforeTokenTransfer`) would be sufficient, as that still makes the assumption that it's actually being called in other functions (in this case `transfer` and `transferFrom`).
So I think the only tests we need per function where we expect `_beforeTokenTransfer` are:
-reverts if transfer hook not set
-reverts if hook reverts (don't need a test for each hook implementation)

XavierEkkel (commented):
will wait for existing comments to be addressed first!

Nilay27:
Addressed the comments and removed unnecessary tests.  Now we only test for the following tests in all the functions where `_beforeTokenTransfer ` is involved:
- `reverts if transfer hook not set`
- `reverts if transfer hook reverts`
- `calls transfer hook with correct parameters`

apps/smart-contracts/token/test/PPO.test.ts
```diff
@@ -23,6 +28,12 @@ describe('=> PPO', () => {
     await ppo.connect(owner).acceptOwnership()
   }
 
+  const setupPPOAndMockTransferHook = async (): Promise<void> => {
+    await setupPPO()
+    mockTransferHook = await smock.fake('BlocklistTransferHook')
```
XavierEkkel, line 33: fakes aren't mocks, we should really be calling this `fakeTransferHook` to make that clear

apps/smart-contracts/token/test/PPO.test.ts
```diff
@@ -23,6 +28,12 @@ describe('=> PPO', () => {
     await ppo.connect(owner).acceptOwnership()
   }
 
+  const setupPPOAndMockTransferHook = async (): Promise<void> => {
```
XavierEkkel, line 31:
```suggestion
  const setupPPOAndFakeTransferHook = async (): Promise<void> => {
```

apps/smart-contracts/token/test/PPO.test.ts
```diff
@@ -361,4 +431,61 @@ describe('=> PPO', () => {
...
+
+    it('reverts if transfer hook reverts', async () => {
+      mockTransferHook.hook.reverts()
+      await ppo.connect(user1).approve(user2.address, 1)
+
+      await expect(ppo.connect(user2).transferFrom(user1.address, user2.address, 1)).to.be.reverted
+    })
+
+    it('calls transfer hook with correct parameters', async () => {
+      await ppo.connect(user1).approve(user2.address, 1)
```
XavierEkkel, line 485:
```suggestion
      await ppo.connect(user1).approve(user2.address, 1)
      
```

apps/smart-contracts/token/test/PPO.test.ts
```diff
@@ -272,10 +319,33 @@ describe('=> PPO', () => {
...
+
+    it('reverts if transfer hook reverts', async () => {
+      mockTransferHook.hook.reverts()
+      await ppo.connect(user1).approve(user2.address, 1)
+
+      await expect(ppo.connect(user2).burnFrom(user1.address, 1)).to.be.reverted
+    })
+
+    it('calls transfer hook with correct parameters', async () => {
+      await ppo.connect(user1).approve(user2.address, 1)
```
XavierEkkel, line 343:
```suggestion
      await ppo.connect(user1).approve(user2.address, 1)
      
```

XavierEkkel (approved):
LGTM after comments

ramenforbreakfast (approved):
LGTM after Xavier's changes are introduced and you rebase this branch with the new changes from main.

PR #250, opened 2022-07-25
feat: add ERC20Permit to PPO

ramenforbreakfast:
To add support for signature permits for token transactions, we use `ERC20PermitUpgradeable` from OpenZeppelin's library.

Because `ERC20PermitUpgradeable` already implements a defined interface `IERC20PermitUpgradeable` unlike `ERC20BurnableUpgradeable`, we don't need to add an interface definition ourselves to `IPPO.sol`.

For `ERC20BurnableUpgradeable`, since OZ provided no interface, we needed to insert function interface definitions in `IPPO.sol` as well. This created a problem because now there was a definition for the functions `burn` and `burnFrom`, in both `IPPO.sol` and `ERC20BurnableUpgradeable.sol`. Solidity forced us to re-define the function, specify an override on both instances, and call the function via `super` since we didn't want to change the function anyway.

Since technically we re-defined the function in our own contract even though we just called it with `super`, we had to rewrite our own tests for something we already knew worked.

So we really don't need to rewrite tests for these functions, and only need to ensure that the `DOMAIN_SEPARATOR` used for messages is generated correctly in our initializer.

For the purposes of this, I added a few utilities that might be helpful in the future as well that were taken from OpenZeppelin's test suite.


apps/smart-contracts/token/test/PPO.test.ts
```diff
@@ -67,11 +67,23 @@ describe('=> PPO', () => {
     })
 
     it('sets owner token balance to zero', async () => {
-      await ppo.balanceOf(deployer.address)
+      expect(await ppo.balanceOf(deployer.address)).to.eq(0)
```
XavierEkkel, line 70: 👍

XavierEkkel (approved):
LGTM after nit!

PR #251, opened 2022-07-25
feat: adding files for generic withdraw and implemented withdraw ERC20

Nilay27:
* Adding files for shared withdraw token contracts.
* Also added `ERC20Mintable.sol` in the shared folder of packages so that it can be mocked and used by other future package contracts.
* Tests for `batchWithdrawERC20` are similar to our usual `withdrawERC20` with only adding cases of array length mismatch and withdrawal of multiple ERC20 tokens.


packages/prepo-shared-contracts/contracts/WithdrawTokens.sol
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+pragma solidity =0.8.7;
+
+import "./interfaces/IWithdrawTokens.sol";
+import "./SafeOwnable.sol";
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import "@openzeppelin/contracts/security/ReentrancyGuard.sol";
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
```
ramenforbreakfast, line 8:
```suggestion
```
This is a redundant import as `SafeERC20.sol` already imports this. Have confirmed it works by compiling without this line.

packages/prepo-shared-contracts/test/WithdrawTokens.test.ts
```diff
@@ -0,0 +1,129 @@
...
+        .connect(owner)
+        .batchWithdrawERC20([firstMockERC20.address], [contractFirstERC20BalanceBefore.sub(1)])
+
+      expect(await firstMockERC20.balanceOf(withdrawTokens.address)).to.eq(1)
+      expect(await firstMockERC20.balanceOf(owner.address)).to.eq(
+        ownerFirstERC20BalanceBefore.add(contractFirstERC20BalanceBefore.sub(1))
+      )
+    })
+
+    it('withdraws if multiple ERC20 tokens', async () => {
```
ramenforbreakfast, line 104:
```suggestion
    it('withdraws different amounts of multiple ERC20 tokens', async () => {
```
You should also use `.sub(2)` for the second amount to ensure that different amounts are recognized.

ramenforbreakfast (dismissed):
LGTM, besides minor nits.

packages/prepo-shared-contracts/contracts/WithdrawTokens.sol
```diff
@@ -0,0 +1,26 @@
...
+
+contract WithdrawTokens is IWithdrawTokens, SafeOwnable, ReentrancyGuard {
+  using SafeERC20 for IERC20;
+
+  constructor() {}
+
+  function batchWithdrawERC20(address[] calldata _erc20Tokens, uint256[] calldata _amounts)
```
ramenforbreakfast, line 15:
```suggestion
  function batchWithdrawERC20(address[] calldata _erc20Contracts, uint256[] calldata _amounts)
```

I notice that in `ERC721` as well, we use the `Token` terminology. Since ERC721s are technically not tokens and are not referred to as such, lets change this naming to `Contracts` across the board.

packages/prepo-shared-contracts/test/WithdrawTokens.test.ts
Nilay27: I have re-written the test to ensure that different amounts are withdrawn, using `const ERC20AmountArray = [parseEther('1'), parseEther('2')]`

ramenforbreakfast (approved):
LGTM, tests look good!

packages/prepo-shared-contracts/contracts/WithdrawTokens.sol
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+pragma solidity =0.8.7;
+
+import "./interfaces/IWithdrawTokens.sol";
+import "./SafeOwnable.sol";
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import "@openzeppelin/contracts/security/ReentrancyGuard.sol";
+
+contract WithdrawTokens is IWithdrawTokens, SafeOwnable, ReentrancyGuard {
```
XavierEkkel, line 9: we don't need to group all the different types into one token because we might only need one type of withdrawal. so we can just have separate contracts for each

XavierEkkel, line 9: will it cause issues if the token that inherits from WithdrawTokens is already using SafeOwnable via some other contract it also inherits from?

packages/prepo-shared-contracts/test/WithdrawTokens.test.ts
```diff
@@ -0,0 +1,137 @@
...
+      await expect(
+        withdrawTokens
+          .connect(owner)
+          .batchWithdrawERC20([firstMockERC20.address], [contractFirstERC20BalanceBefore.add(1)])
+      ).revertedWith('ERC20: transfer amount exceeds balance')
+    })
+
+    it('withdraws if amount = contract balance', async () => {
+      const contractFirstERC20BalanceBefore = await firstMockERC20.balanceOf(withdrawTokens.address)
+      const ownerFirstERC20BalanceBefore = await firstMockERC20.balanceOf(owner.address)
```
XavierEkkel, line 86: check that the balance is >0, and then separate test for if amount = 0

packages/prepo-shared-contracts/test/WithdrawTokens.test.ts
```diff
@@ -0,0 +1,137 @@
...
+        .connect(owner)
+        .batchWithdrawERC20([firstMockERC20.address], [contractFirstERC20BalanceBefore.sub(1)])
+
+      expect(await firstMockERC20.balanceOf(withdrawTokens.address)).to.eq(1)
+      expect(await firstMockERC20.balanceOf(owner.address)).to.eq(
+        ownerFirstERC20BalanceBefore.add(contractFirstERC20BalanceBefore.sub(1))
+      )
+    })
+
+    it('withdraws different amounts of multiple ERC20 tokens', async () => {
```
XavierEkkel, line 112: add test for withdrawing same token multiple times

XavierEkkel (approved):
LGTM after comments

packages/prepo-shared-contracts/contracts/WithdrawTokens.sol
Nilay27: It will not cause issues in such scenarios. I also cross-checked by inheriting it in the Vesting contract and then running the tests accordingly. Vesting inherits Pausable, which also inherits SafeOwnable.

PR #256, opened 2022-07-26
feat: added natspec docs for IPPO and IAccountList

ramenforbreakfast:
I have finished writing the natspec copy for `IPPO.sol` and `IAccountList.sol`. I will submit the copy for `ITransferHook.sol`, `IUnrestrictedTransferHook.sol` and `IRestrictedTransferHook.sol` later.


apps/smart-contracts/token/contracts/interfaces/IAccountList.sol
```diff
@@ -1,13 +1,43 @@
...
+   * @notice Removes every address from the set. Atomically includes any
+   * addresses in `newIncludedAccounts`.
+   * @dev Only callable by `owner()`.
+   * @param newIncludedAccounts Addresses to include after reset
+   */
   function reset(address[] calldata newIncludedAccounts) external;
 
+  /**
+   * @dev This contract does not delete anything when `reset()` is called.
+   * Instead, an index is incremented which points to a new set.
```
Nilay27, line 33: I think it's better to add this in `reset()` instead.

apps/smart-contracts/token/contracts/interfaces/IPPO.sol
```diff
@@ -3,15 +3,45 @@ pragma solidity =0.8.7;
 
 import "./ITransferHook.sol";
 
-//TODO: add all natspecs at the end
+/**
+ * @notice PPO is an ERC-20 asset that serves as the governance token for the
+ * prePO platform. PPO token-holders can stake their tokens for voting on
```
Nilay27, line 8:
```suggestion
 * prePO platform. PPO holders can stake their tokens for voting on
```

apps/smart-contracts/token/contracts/interfaces/IPPO.sol
```diff
@@ -3,15 +3,45 @@ pragma solidity =0.8.7;
...
+ * community proposals and receiving rewards.
+ */
 interface IPPO {
+  /**
+   * @notice Sets the hook to be called before a token transfer.
+   * @dev This contract will call the `hook` function and pass along the
```
Nilay27, line 14:
```suggestion
   * @dev This contract will call the `hook` function of the transfer hook and pass along the
```

Also, we should add somewhere that the `hook` function is called inside `_beforeTokenTransfer`

apps/smart-contracts/token/contracts/interfaces/IPPO.sol
```diff
@@ -3,15 +3,45 @@ pragma solidity =0.8.7;
...
+  /**
+   * @notice Burns `amount` tokens from `account`.
+   * @dev The caller's allowance with the `account` must be >= `amount` and
+   * will be deducted by `amount`.
+   * @param account Address to burn `PPO` from
+   * @param amount Amount of `PPO` to be burned
+   */
   function burnFrom(address account, uint256 amount) external;
 
+  /// @return Address of the configured `ITransferHook` contract
```
Nilay27, line 45: We are not just returning the address, we are returning the entire instance of `ITransferHook`. This method can be used by another contract to call functions like `getBlockList()`, `getToken()` etc of transferHook

Nilay27 (approved):
LGTM after nits

apps/smart-contracts/token/contracts/interfaces/IAccountList.sol
ramenforbreakfast: I could see a reason, for it happening in either place, I originally put it here because I saw no reason to inform anyone of the actual mechanics.

With the getter, you kind of need to explain it since it would be confusing why there'd be an index.

apps/smart-contracts/token/contracts/interfaces/IPPO.sol
ramenforbreakfast: I have taken this advice and included detail that it is called within `_beforeTokenTransfer`

apps/smart-contracts/token/contracts/interfaces/IPPO.sol
ramenforbreakfast: Technically speaking, all an external contract sees is an address, only an internal contract would understand to wrap the address in an interface.

An external entity would have to rewrap this in the interface to be able to interact with its methods.

apps/smart-contracts/token/contracts/interfaces/IPPO.sol
```diff
@@ -3,15 +3,46 @@ pragma solidity =0.8.7;
...
+  /**
+   * @notice Burns `amount` tokens from the caller.
+   * @param amount Amount of `PPO` to be burned
+   */
   function burn(uint256 amount) external;
 
+  /**
+   * @notice Burns `amount` tokens from `account`.
+   * @dev The caller's allowance with the `account` must be >= `amount` and
+   * will be deducted by `amount`.
```
XavierEkkel, line 40:
```suggestion
   * will be decreased by `amount`.
```

XavierEkkel (approved):
LGTM after comments!

apps/smart-contracts/token/contracts/interfaces/IAccountList.sol
```diff
@@ -1,13 +1,43 @@
...
+   * @return Index of the current set
+   */
   function getResetIndex() external view returns (uint256);
```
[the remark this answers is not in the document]
ramenforbreakfast: Gotcha, I'll leave a TODO in the `AccountList.sol` to remove it.

apps/smart-contracts/token/contracts/interfaces/IAccountList.sol
ramenforbreakfast: This is now irrelevant since `AccountList.sol` will have its index removed. Previously it was requested that `Blocklist.sol`, the precursor to `AccountList.sol`, have its index not be externally exposed, but when we cancelled `Blocklist.sol`, we forgot to carry over that request. I am leaving a TODO to remove it in `AccountList.sol`

PR #260, opened 2022-07-29
refactor: accommodate for local important address stores

ramenforbreakfast:
While writing the `Vesting` deployment script, I noticed a problem that I wanted to solve now that we have our `prepo-constant` address store.

I found myself having to copy and paste code that throws and error if a specific address definition is undefined or blank for a network, so I created a `getImportantAddressForNetwork` function that does this automatically whenever a developer needs an `IMPORTANT_ADDRESS`. 

In addition, I wanted to solve the problem of when a SC developer wants to do local testing, they need to keep important addresses locally stored in their `.env` file of the specific project they're working on. I didn't want to have to keep re-creating the same code over and over to default to a local environment variable if I was testing locally.

So you will notice that `getImportantAddressForNetwork` takes in a `localEnvFilePath` since the `prepo-constant` package needs to be able to reference whatever `.env` file in the project folder you are working on and the local important addresses in your `.env` would differ in the `core` project and the `token` project. 

This parameter is entirely optional so if you know you won't be using local networks, you don't have to provide it and the function will throw an error if you forget.

I have also taken this chance to move `dotenv` to the root package.json since it will be used by multiple folders.


ramenforbreakfast:
Putting this review on hold since there seems to be problems with TS GitHub Actions checks. I have not been able to solve this and am prioritizing getting a deployment out for Chris to use.

ramenforbreakfast:
Chris has resolved the problem preventing us from using the `fs` library in `prepo-constants`. He has merged the fix into this branch so that the GitHub Actions do not fail anymore. The PR is now ready for review.

apps/smart-contracts/token/deploy/PPO.ts
```diff
@@ -16,16 +17,9 @@ const ensureProxyOwnedByGovernance = async function (
...
-  if (governanceAddress !== undefined) {
-    if (adminOwner !== governanceAddress) {
-      console.log(
-        'ProxyAdmin not owned by governance, transferring ownership to',
-        governanceAddress
-      )
-      await upgrades.admin.transferProxyAdminOwnership(governanceAddress)
-    }
-  } else {
-    throw new Error('Governance not defined for the current network')
```
XavierEkkel, line 28: don't we still want to throw this if there's no governance address for the current network?

apps/smart-contracts/token/deploy/PPO.ts
```diff
@@ -53,7 +47,11 @@ const deployFunction: DeployFunction = async function ({
    * This can be temporarily removed if deploying to prod.
    */
   assertIsTestnetChain(currentChain)
-  const governanceAddress = IMPORTANT_ADDRESSES.GOVERNANCE[currentNetwork.name]
+  const governanceAddress = getImportantAddressForNetwork(
+    'GOVERNANCE',
+    currentNetwork.name,
+    path.join(__dirname, '../.env')
```
XavierEkkel, line 53: won't this always be defined regardless of whether they're deploying locally or on testnet? and if so the argument isn't really optional, and we could just move this logic within the function itself?

XavierEkkel (approved):
LGTM after questions and if LGT @liqwiz

packages/prepo-constants/src/addresses.ts
```diff
@@ -1,15 +1,38 @@
...
 }
 
 export const IMPORTANT_ADDRESSES: ImportantAddresses = {
-    GOVERNANCE: {
+    "GOVERNANCE": {
         arbitrumTestnet: "0x054CcD68A2aC152fCFB93a15b6F75Eea53DCD9E0"
     }
 }
+
+export function getImportantAddressForNetwork(keyword: string, network: string, localEnvFilePath?: string): string {
```
liqwiz, line 19: `ImportantAddress` is kind of ambiguous, it's not clear to me as a developer what important means here. maybe `PpoContractAddress` is more descriptive?

packages/prepo-constants/src/addresses.ts
```diff
@@ -1,15 +1,38 @@
...
+            throw new Error("A local network was detected, but no envfile path was provided to retrieve the important address")
+        }
+    } else {
+        addressForKeyword = IMPORTANT_ADDRESSES[keyword][supportedNetwork]
+    }
+    if (addressForKeyword === undefined || addressForKeyword === "") {
+        throw new Error(keyword + " not defined for the " +  supportedNetwork + " network")
+    }
+    return getAddress(addressForKeyword)
+}
```
liqwiz, line 38: the advised usage of `dotenv` is to call `.config` just once as early as possible in the entry point of your application, it makes stuff unnecessarily complicated injecting env vars arbitrarily in function calls, and it only needs to be called once

for this use case `PPO.ts` is the entrypoint, so what I suggest is

1. inject env vars right after imports inside `PPO.ts`

```js
dotenv.config({
  path: '../.env',
})
```

2. change `getImportantAddressForNetwork(` (or whatever it gets renamed to) to something like this
```suggestion
export function getImportantAddressForNetwork(
  addressLabel: string,
  network: SupportedNetworks,
  localAddress?: string
): string {
  // Handle local network
  if (network === 'hardhat' || network === 'ganache') {
    if (!localAddress) throw new Error('A localAddress must be provided for local networks')
    return getAddress(localAddress)
  }

  // Handle remote network
  if (!IMPORTANT_ADDRESSES[addressLabel] || !IMPORTANT_ADDRESSES[addressLabel][network]) {
    throw new Error(`${addressLabel} not defined for the ${network} network`)
  }
  return getAddress(IMPORTANT_ADDRESSES[addressLabel][network])
}
```

and then call it in `PPO.ts` something like this
```js
const governanceAddress = getImportantAddressForNetwork(
    'GOVERNANCE',
    currentNetwork.name,
    process.env["GOVERNANCE"]  
)
```

apps/smart-contracts/token/deploy/PPO.ts
ramenforbreakfast: This has been moved to the local address fetching function `getImportantAddressForNetwork`

packages/prepo-hardhat/src/smart-contracts-utils.ts
```diff
@@ -1,4 +1,4 @@
-import { BigNumber, providers, Contract } from 'ethers'
+import { BigNumber, providers, Contract, ContractTransaction } from 'ethers'
```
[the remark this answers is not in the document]
ramenforbreakfast: nope, good catch, will remove this.

apps/smart-contracts/token/deploy/PPO.ts
ramenforbreakfast: `__dirname` is specific to when its being accessed, thus i cannot move it to the function itself since it is in the `prepo-constants` package.

packages/prepo-constants/src/addresses.ts
```diff
@@ -1,15 +1,38 @@
 import { SupportedNetworks } from "./networks"
```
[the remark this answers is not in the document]
ramenforbreakfast: Have discussed with Merlin on enabling automatic eslint formatting in VSCode

packages/prepo-constants/src/addresses.ts
```diff
@@ -1,15 +1,38 @@
...
+import dotenv from "dotenv"
+import { getAddress } from "ethers/lib/utils"
 
 export type ImportantAddress = {
     [key in SupportedNetworks]?: string
 }
 
 export type ImportantAddresses = {
-    GOVERNANCE: ImportantAddress
+    [key: string]: ImportantAddress
```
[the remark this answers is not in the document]
ramenforbreakfast: As recommended, I have defined an `ImportantKeywords` type  instead of an ambiguous `string` type.

packages/prepo-constants/src/addresses.ts
ramenforbreakfast: I have decided to keep the generic data structures labeled with the `Important...` prefix. The function and constant storing these addresses I have renamed to `PREPO_ADDRESSES`, since these are not necessarily contract addresses and can include EOAs.

apps/smart-contracts/token/deploy/PPO.ts
ramenforbreakfast: This is now irrelevant as I have now applied Merlin's recommendation to access the environment variable in the deployment script, rather than trying to do it inside the package.

PR #261, opened 2022-07-29
feat: added Vesting deployment script

ramenforbreakfast:
At the request of @chrisling-dev , we need to deploy an instance of the `Vesting` contract. This deployment script follows the format of the old `core` deployment scripts since this is not an upgradeable contract and does not need most of the new OZ Defender specific features.

I have made this not a dependency of `PPO` since I don't want to lock their deployments together, but it will throw an error if one is not detected.


ramenforbreakfast:
Putting this review on hold since there seems to be problems with TS GitHub Actions checks. I have not been able to solve this and am prioritizing getting a deployment out for Chris to use.

ramenforbreakfast:
The deployment script for the Vesting contract has been modified with changes to deployment scripts made in PR #260 and is ready for review.

apps/smart-contracts/token/deploy/Vesting.ts
```diff
@@ -0,0 +1,64 @@
...
+  const { deploy, getOrNull } = deployments
+  const deployer = (await ethers.getSigners())[0]
+  console.log('Running Vesting deployment script with', deployer.address, 'as the deployer')
+  const currentChain = Number(await getChainId()) as ChainId
+  const currentNetwork = getNetworkByChainId(currentChain)
+  /**
+   * Make sure this script is not accidentally targeted towards a production environment.
+   * This can be temporarily removed if deploying to prod.
+   */
+  assertIsTestnetChain(currentChain as unknown as ChainId)
```
XavierEkkel, line 30: why the double `as`?

XavierEkkel (approved):
LGTM

apps/smart-contracts/token/deploy/Vesting.ts
ramenforbreakfast: Yep, this isn't necessary, good catch.

PR #265, opened 2022-08-01
fix: bump ethers version

liqwiz:
Need to bump `ethers` to fix error when running `yarn add ...`

Needed to also bump `hardhat` to work with latest ethers


ramenforbreakfast:
Due to the upgrade, I had to change multiple things:
- We cannot directly cast the `JsonRpcProvider` type returned by `hardhat-ethers` into a `Web3Provider` which was the only type made available by previous libraries. The new `ethers` version makes this `JsonRpcProvider` type available and I have fixed our `getLastTimestamp` and `setNextTimestamp` to directly take in the `JsonRpcProvider` type, so we no longer need to do re-castings of our provider throughout our tests.

- Previously, there was a bug in hardhat that made smart contract queries of `block.chainId` resolve to 0 while testing, now it correctly returns 31337 during testing.

- Needed to upgrade `@nomiclabs/hardhat-chai-matchers` library as the old one was incompatible with the new hardhat.

ramenforbreakfast (approved):
Feel free to merge if everything I did looks okay.

PR #253, opened 2022-07-25
feat: adding ERC1155 batch withdraw method

Nilay27:
* Adding method for `batchWithdrawERC1155` with relevant tests.
* Added file for `ERC1155Mintable.sol` in shared contracts.
* Also added tests for `onERC1155Received` and `onERC1155BatchReceived`.

PR #267, opened 2022-08-01
feat: adding generic ERC1155 withdraw contract

Nilay27:
* Opening this PR instead of #253  due to the decision to separate each withdrawal function into separate contracts.
* Also added ERC1155Mintable.sol in the shared folder of packages so that it can be mocked and used by other future package contracts.
* Tests for batchWithdrawERC1155 are similar to our usual withdrawERC1155 with only adding cases of array length mismatch and withdrawal of multiple ERC1155 tokens.
* Also resolved @ramenforbreakfast comments on #253


packages/prepo-shared-contracts/test/WithdrawERC1155.test.ts
```diff
@@ -0,0 +1,360 @@
...
+      )
+      expect(await firstMockERC1155.balanceOf(withdrawERC1155.address, ERC1155IdArray[1])).to.eq(
+        contractFirstERC1155Id2BalanceBefore.sub(ERC1155AmountArray[1])
+      )
+      expect(await firstMockERC1155.balanceOf(owner.address, ERC1155IdArray[1])).to.eq(
+        ownerFirstERC1155Id2BalanceBefore.add(ERC1155AmountArray[1])
+      )
+    })
+
+    it('withdraws different amounts of multiple ERC1155 tokens', async () => {
```
XavierEkkel, line 276:
```suggestion
    it('withdraws different amounts of multiple ids of different contracts', async () => {
```

packages/prepo-shared-contracts/contracts/WithdrawERC1155.sol
```diff
@@ -0,0 +1,51 @@
...
+    uint256[] calldata _amounts
+  ) external override onlyOwner nonReentrant {
+    require(
+      _erc1155Tokens.length == _amounts.length && _ids.length == _amounts.length,
+      "Array length mismatch"
+    );
+    for (uint256 i; i < _erc1155Tokens.length; ++i) {
+      IERC1155(_erc1155Tokens[i]).safeTransferFrom(
+        address(this),
+        owner(),
```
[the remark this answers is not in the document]
XavierEkkel: let's leave a TODO for this for later

XavierEkkel (approved):
LGTM after comments

packages/prepo-shared-contracts/contracts/WithdrawERC1155.sol
```diff
@@ -0,0 +1,51 @@
...
+        _ids[i],
+        _amounts[i],
+        ""
+      );
+    }
+  }
+
+  function onERC1155Received(
```
ramenforbreakfast, line 32: Just want to be clear we still need this because standard `ERC1155` implementation does not have a non `safeTransferFrom` function.

packages/prepo-shared-contracts/contracts/WithdrawERC1155.sol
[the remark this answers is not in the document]
ramenforbreakfast: Have confirmed on the SC end that Gnosis Safe does not support receiving ERC1155s.

ramenforbreakfast (approved):
LGTM, besides Xaviers comments and my nits.

packages/prepo-shared-contracts/contracts/WithdrawERC1155.sol
```diff
@@ -0,0 +1,51 @@
...
+
+import "./interfaces/IWithdrawERC1155.sol";
+import "./SafeOwnable.sol";
+import "@openzeppelin/contracts/token/ERC1155/IERC1155.sol";
+import "@openzeppelin/contracts/security/ReentrancyGuard.sol";
+
+contract WithdrawERC1155 is IWithdrawERC1155, SafeOwnable, ReentrancyGuard {
+  constructor() {}
+
+  function withdrawERC1155(
```
XavierEkkel, line 12:
```suggestion
  function withdraw(
```

ramenforbreakfast:
Summarizing my discussion with @Nilay27 

Let's leave out `onERC1155Received` as well for this. It should not be the responsibility of this `withdraw...` module to ensure a contract can receive ERC1155's. For testing purposes, let's just have a wrapper contract that has `onERC1155Received` so that we can actually send ERC1155 tokens to it for testing.

This is the same philosophy we agreed upon for ERC721, a contract should only inherit this module with the expectation that they will be able to withdraw tokens sent to them, but it shouldn't be its responsibility to implement functions that enable it to receive those tokens, that should be completely up to the inheriting contract.

Also ensure `TokenShop` has `onERC...` functions defined and test them.

XavierEkkel:
> Summarizing my discussion with @Nilay27
> 
> Let's leave out `onERC1155Received` as well for this. It should not be the responsibility of this `withdraw...` module to ensure a contract can receive ERC1155's. For testing purposes, let's just have a wrapper contract that has `onERC1155Received` so that we can actually send ERC1155 tokens to it for testing.
> 
> This is the same philosophy we agreed upon for ERC721, a contract should only inherit this module with the expectation that they will be able to withdraw tokens sent to them, but it shouldn't be its responsibility to implement functions that enable it to comply with `safeTransfer...` functions, that should be completely up to the inheriting contract.
> 
> Also ensure `TokenShop` has `onERC...` functions defined and test them.

I think it's different here to ERC721 - in ERC1155 it would be impossible for this contract to ever receive ERC1155 unless the `onERC1155Received` is included, so it makes sense to group them in this contract since they will always have to be together.

ramenforbreakfast:
> > Summarizing my discussion with @Nilay27
> > Let's leave out `onERC1155Received` as well for this. It should not be the responsibility of this `withdraw...` module to ensure a contract can receive ERC1155's. For testing purposes, let's just have a wrapper contract that has `onERC1155Received` so that we can actually send ERC1155 tokens to it for testing.
> > This is the same philosophy we agreed upon for ERC721, a contract should only inherit this module with the expectation that they will be able to withdraw tokens sent to them, but it shouldn't be its responsibility to implement functions that enable it to comply with `safeTransfer...` functions, that should be completely up to the inheriting contract.
> > Also ensure `TokenShop` has `onERC...` functions defined and test them.
> 
> I think it's different here to ERC721 - in ERC1155 it would be impossible for this contract to ever receive ERC1155 unless the `onERC1155Received` is included, so it makes sense to group them in this contract since they will always have to be together.

@Nilay27 let's keep it in then. I prefer to keep them separate, but I don't have a strong opinion on this.

Nilay27:
* Added `_recipients` argument to the `withdrawERC1155()` and not removing `onERC1155Received` and `onERC1155BatchReceived`  as per the decision made.

PR #273, opened 2022-08-05
chore: move tokenshop contracts to subfolder

ramenforbreakfast:
Moves `ITokenShop.sol`, `IPurchaseHook.sol`, `TokenShop.sol`, and `PurchaseHook.sol` to the `tokenshop` subfolder.


XavierEkkel:
changed base to main because the previous base was independent from this PR

PR #275, opened 2022-08-05
feat: adding files for MiniSales with initialization

Nilay27:
* Adding files for `MiniSales` along with tests for values of initial state variables.
* Implemented getters for `getSaleToken()` and  `getPaymentToken()` to test for the initial state values of `_saleToken`, `_paymentToken`.
* Also added function signature of all the methods in `IMiniSales` to avoid making the contract abstract.


XavierEkkel (approved):
LGTM after fixing to be compatible with interface changes in base PR

apps/smart-contracts/token/contracts/mini-sales/MiniSales.sol
```diff
@@ -0,0 +1,44 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+pragma solidity =0.8.7;
+
+import "./interfaces/IMiniSales.sol";
+import "prepo-shared-contracts/contracts/SafeOwnableUpgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/security/ReentrancyGuardUpgradeable.sol";
+
+contract MiniSales is IMiniSales, SafeOwnableUpgradeable, ReentrancyGuardUpgradeable {
+  IERC20 private _saleToken;
+  IERC20 private _paymentToken;
```
ramenforbreakfast, line 10: Are we not making these immutable anymore? I thought we all agreed in the design doc these would be immutable? Upgradeable contracts can have immutable variables, just need to define a constructor in addition to `initialize`.

Keep in mind, the fixture needs to be updated to pass in constructor args, please refer to the docs about `deployProxy` here https://docs.openzeppelin.com/upgrades-plugins/1.x/api-hardhat-upgrades

ramenforbreakfast (commented):
I notice that these do not have the new argument names from the recent change in the `IMiniSales` interface, will make these changes as well.

apps/smart-contracts/token/contracts/mini-sales/MiniSales.sol
ramenforbreakfast: Have confirmed with the team that this was unintentional and will make this change myself.

PR #277, opened 2022-08-05
feat: adding purchase for minisales

Nilay27:
* Adding `purchase()` method along with relevant tests.
* Also added initial file for `PurchaseHook.sol` for mocking purposes.
* Moreover, since now we have two `PurchaseHook` therefore we now have to specify the exact path to the PurchaseHook that we want to mock or getContractFactory for. Hence I went ahead and made the required changes in `PurchaseHookFixtures.ts`, `PurchaseHook.test.ts` and `TokenShop.test.ts`


ramenforbreakfast:
Closing because we need to add a sale token denominator and tests are going to need to be modified to take into account fixed point math. Its just going to be confusing for everyone if we leave this open and combine these changes into this PR or submit them afterwards.

PR #282, opened 2022-08-07
feat: adding allowlist purchase hook setters

Nilay27:
* Adding files for `AllowlistPurchaseHook` along with setters for `_allowedContract` and `_allowlist`.
* In the interface for `IAllowlistPurchaseHook`, event for `AllowlistChange` was not added, I have added it as per the requirement in the tech doc.
* Also added `setAllowedContract()` and `getAllowedContract()` in the interface as we only want the allowed contract to be able to call the hook.


ramenforbreakfast (approved):
LGTM, however I think I will submit a follow-up PR to consolidate `allowedContract` modifiers. `PPO` transfer hooks essentially use them as well and it would make sense to have them all share the same contract logic. That way if the team decides this is not the way to go, we won't hold up the PRs currently under review.

ramenforbreakfast:
> LGTM, however I think I will submit a follow-up PR to consolidate `allowedContract` modifiers. `PPO` transfer hooks essentially use them as well and it would make sense to have them all share the same contract logic. That way if the team decides this is not the way to go, we won't hold up the PRs currently under review.

Just noting here that we are removing the `onlyAllowedContract` and `onlyToken` from contracts since we are going to keep the `hook` external, but since effectively our current implementations are view-only, we don't need them in our hook contracts for now.

XavierEkkel (approved):
LGTM under the knowledge that `get/setAllowedContract` will be removed in https://github.com/prepo-io/prepo-monorepo/pull/284/

PR #284, opened 2022-08-07
feat: adding hook for allowlist purchase hook

Nilay27:
* adding implementation for `hook()` in `AllowlistPurchaseHook.sol`. 
* Currently, it just checks for whether the recipient is a part of the allowlist or not.


apps/smart-contracts/token/test/AllowlistPurchaseHook.test.ts
```diff
@@ -45,6 +57,46 @@ describe('AllowlistPurchaseHook', () => {
...
+
+    it('succeeds if recipient allowed', async () => {
+      allowedAccounts.isIncluded.whenCalledWith(recipient.address).returns(true)
+
+      await expect(
+        allowlistPurchaseHook
+          .connect(allowedContract)
+          .hook(purchaser.address, recipient.address, 1, 1)
+      ).to.not.reverted
+    })
```
ramenforbreakfast, line 97:
```suggestion
    it('reverts if recipient not allowed', async () => {
      allowedAccounts.isIncluded.whenCalledWith(recipient.address).returns(false)

      await expect(
        allowlistPurchaseHook
          .connect(allowedContract)
          .hook(purchaser.address, recipient.address, 1, 1)
      ).to.be.revertedWith('Recipient not allowed')
    })

    it('succeeds if recipient allowed', async () => {
      allowedAccounts.isIncluded.whenCalledWith(recipient.address).returns(true)

      await expect(
        allowlistPurchaseHook
          .connect(allowedContract)
          .hook(purchaser.address, recipient.address, 1, 1)
      ).to.not.reverted
    })
```

I think since we are using the smock library here, it is preferable that we simply verify that the `isIncluded` function is called with the correct parameters and that reverts are passed upward, rather than re-testing the underlying behavior.

ramenforbreakfast (approved):
LGTM besides on test suite modification.

apps/smart-contracts/token/test/AllowlistPurchaseHook.test.ts
ramenforbreakfast: Nilay and I have concluded that we should just add another test case verifying that the function is called with the correct arguments. Since we still need these tests to ensure `Recipient not allowed` is thrown (this belongs to the Hook and not AccountList, so we need to include it in this test suite).

Nilay27:
Removed `setAllowedContract()` , `getAllowedContract()` and `onlyAllowedContract()` modifier along with relevant tests.

ramenforbreakfast:
Approved! Same reasoning with the previous PR, decided to include removing `onlyAllowedContract` for this hook since it doesn't write to anything and this PR is already very small.

PR #285, opened 2022-08-08
feat: add reusable ProtectedHook contract

ramenforbreakfast:
After reviewing PR #282-284, I think it makes sense for us to have all of our contract hooks to inherit from a shared contract that stores an address which can call the hook.

This has been submitted separately with the intention of not creating conflicts with the current aforementioned PRs, and if accepted, would be integrated in a later PR.

For the revert message of `onlyAllowedContract`, I followed the naming convention of our `SafeOwnable` contract, which uses the format of `<Contract Name>: message`. It also does not explicitly say `msg.sender` in its message, and instead uses the word `sender`. 

Do we want to stick with this naming convention? It seems there is a preference in recent PR reviews for other contracts to now explicitly say `msg.sender`, rather than words like `caller` or `sender`.


packages/prepo-shared-contracts/contracts/ProtectedHook.sol
```diff
@@ -0,0 +1,23 @@
...
+pragma solidity =0.8.7;
+
+import "./SafeOwnable.sol";
+import "./interfaces/IProtectedHook.sol";
+
+contract ProtectedHook is IProtectedHook, SafeOwnable {
+  address private _allowedContract;
+
+  modifier onlyAllowedContract() {
+    require(_msgSender() == _allowedContract, "ProtectedHook: sender must be allowed contract");
```
Nilay27, line 11:
```suggestion
    require(_msgSender() == _allowedContract, "ProtectedHook: msg.sender != allowed contract");
```

As we always describe why the error occurred rather than telling the solution

packages/prepo-shared-contracts/contracts/ProtectedHook.sol
```diff
@@ -0,0 +1,23 @@
...
+    _;
+  }
+
+  function setAllowedContract(address _newAllowedContract) external override onlyOwner {
+    _allowedContract = _newAllowedContract;
+    emit AllowedContractChange(_newAllowedContract);
```
Nilay27, line 17: Do we want to emit an event for changing the allowed contract? 
I had a discussion with @XavierEkkel on this, and since setting this will not change the state of the calling/main contract or affect its functionality, we decided not to emit events for such functions.

packages/prepo-shared-contracts/test/ProtectedHook.test.ts
```diff
@@ -0,0 +1,108 @@
...
+  let user2: SignerWithAddress
+  let protectedHook: ProtectedHookWrapper
+
+  const setupProtectedHook = async (): Promise<void> => {
+    ;[deployer, user1, user2] = await ethers.getSigners()
+    owner = deployer
+    protectedHook = await protectedHookWrapperFixture()
+  }
+
+  describe('initial state', () => {
```
Nilay27, line 22: Add test for `'sets allowed contract to zero address'`

packages/prepo-shared-contracts/test/ProtectedHook.test.ts
```diff
@@ -0,0 +1,108 @@
...
+    })
+
+    it('unpauses', async () => {
+      await protectedHook.connect(owner).setAllowedContract(JUNK_ADDRESS)
+      expect(await protectedHook.getAllowedContract()).to.eq(JUNK_ADDRESS)
+
+      await protectedHook.connect(owner).setAllowedContract(ZERO_ADDRESS)
+
+      expect(await protectedHook.getAllowedContract()).to.eq(ZERO_ADDRESS)
+    })
```
Nilay27, line 64: ??
The naming of tests is incorrect, it is not a test for pausing and unpausing, we should be checking for `sets to non-zero address` and `sets to zero address`

Nilay27 (approved):
LGTM after one question and a few test suite nits.

packages/prepo-shared-contracts/contracts/ProtectedHook.sol
ramenforbreakfast: Ok if @XavierEkkel made that call then I will just remove it.

packages/prepo-shared-contracts/test/ProtectedHook.test.ts
ramenforbreakfast: Yeah my bad, i copied a lot of the test suite from Pausable, will change it.

packages/prepo-shared-contracts/contracts/ProtectedHook.sol
ramenforbreakfast: I put an explanation for why I did it the other way, since I'm not sure since contracts like `SafeOwnable` have it written differently.

XavierEkkel: Nilay is correct, and we shouldn't base our approach on the different approaches of external contracts.
I would also add that our standard is to not use a contract name prefix:
```suggestion
    require(_msgSender() == _allowedContract, "msg.sender != allowed contract");
```

packages/prepo-shared-contracts/test/fixtures/ProtectedHookFixtures.ts
```diff
@@ -0,0 +1,7 @@
+import { ethers } from 'hardhat'
```
XavierEkkel, line 1: nit: this isn't a 'grouped' fixtures file, it's just specific to a particular contract so there will only be one fixture.
so we can just call the file `ProtectedHookFixture.ts` (singular)

XavierEkkel (approved):
LGTM after comments

packages/prepo-shared-contracts/contracts/ProtectedHook.sol
ramenforbreakfast: Ok!  I'll make the change as well to SafeOwnable

PR #288, opened 2022-08-08
fix: changed name of withdrawERC20 method to withdraw

Nilay27:
* Changed the name of `withdrawERC20()` to `withdraw()` as per the suggestions made for the WithdrawERC721 and WithdrawERC1155 contracts. 
* This contract was already merged before those suggestions were made.


ramenforbreakfast (approved):
LGTM! So i'm assuming the plan is if a contract inherits from multiple different `Withdraw...` contracts, we rely on `withdraw` being distinct based on their different parameters, rather than explicitly naming each function `withdrawERC<blah>...`.

XavierEkkel:
> LGTM! So i'm assuming the plan is if a contract inherits from multiple different `Withdraw...` contracts, we rely on `withdraw` being distinct based on their different parameters, rather than explicitly naming each function `withdrawERC<blah>...`.

Actually this is a good point, the ERC721 version would actually have the same signature: `function withdraw(address[] calldata _erc721Tokens, uint256[] calldata _ids)`, so I will just close this PR.

PR #292, opened 2022-08-09
chore: updated dino-docs

georgesmithishere:


chrisling-dev (changes requested):
Screenshot 2022-08-12 at 4 14 04 PM

Build seems to be failing because of this reason

PR #296, opened 2022-08-10
feat: change minisales to non-upgradeable

Nilay27:
* Changing the mini sales contract as per the decision to change it to Non-upgradeable.
* Removed SafeOwnable and Reentrancy guard imports as both are included in `WithdrawERC20.sol`.
* As per discussion with @ramenforbreakfast , making the `_saleTokenDecimals` immutable and setting it in the constructor itself.
* Also, as we are now not reading the decimals from `_saleToken.Decimals()` and instead just passing 18 in the argument, I added `getSaleTokenDecimals()` so that we can be sure that it was set correctly during initialization.


apps/smart-contracts/token/test/MiniSales.test.ts
```diff
@@ -65,14 +69,17 @@ describe('=> MiniSales', () => {
...
     })
 
-    it('sets payment token from initialize', async () => {
+    it('sets payment token from constructor', async () => {
       expect(await miniSales.getPaymentToken()).to.eq(paymentToken.address)
     })
 
+    it('sets sale token decimals from constructor', async () => {
+      expect(await miniSales.getSaleTokenDecimals()).to.eq(parseUnits('1', saleTokenDecimals))
+    })
```
ramenforbreakfast, line 82:
```suggestion
    })
    
```

ramenforbreakfast (approved):
LGTM besides nit

apps/smart-contracts/token/contracts/mini-sales/interfaces/IMiniSales.sol
```diff
@@ -34,4 +34,6 @@ interface IMiniSales {
   function getPrice() external view returns (uint256);
 
   function getPurchaseHook() external view returns (IPurchaseHook);
+
+  function getSaleTokenDecimals() external view returns (uint256);
```
XavierEkkel, line 38: I disagree with adding this - external callers can just read decimals from the sale token directly (under our assumption that it never changes), and we shouldn't add anything to the interface just for tests.

We can write our tests just to check that amounts are correct with different-decimal tokens without needing to expose this.

XavierEkkel (approved):
LGTM after removing getter

PR #294, opened 2022-08-09
feat: adding test for when claimer is not source allowed

Nilay27:
* This is part 2 of the split up draft PR #287. 
* It contains a nested `describe` block test for the scenarios when `if user claims and is not allowed source`.

PR #287, opened 2022-08-08
feat: adding draft for end to end testing for Vesting contract

Nilay27:
* The tests cover the scenarios related to the user claiming and then trying transferring the tokens. Hence I have only taken for cases where vesting is ongoing and did not include when vesting has ended as that is not affected by PPO and transfer hook.

* Also replaced `utils.setNextTimestamp` with `setNextTimestamp` by importing it

PR #298, opened 2022-08-10
test: vesting e2e test when claimer is not allowed source

Nilay27:
* Resubmitting PR #294 which was already LGTMd by Michael, also resolved his comments on the PR.
* This is part 2 of the split up draft PR #287. 
* It contains a nested `describe` block test for the scenarios when `if user claims and is not allowed source`.


ramenforbreakfast (approved):
LGTM!

ramenforbreakfast:
Closing due to irrelevance, will need to be heavily modified anyway if we want to include them because of how old this PR is.

PR #300, opened 2022-08-10
test: miniSales e2e test purchase and subsequent block

Nilay27:
* This PR contains end-to-end tests of MiniSales for following scenarios:
  1. two single test scenarios i.e `if the user is blocked` and `if the user is not blocked`
  2. One nested `describe` covering the transfers of purchased PPO by the user in the case `if user purchases, but is subsequently blocked`
* For clarity, I have written all the tests such that purchaser and recipient are different.
* I also had to split up `deployMiniSales` into `setupSignersAndMockTokens` + deployment so that we do not end up deploying MiniSales twice. As in the `deployMiniSales()` we deploy it with a `MockERC20` token, whereas for E2E testing, we need the actual `PPO` token.

PR #301, opened 2022-08-10
feat: minisales e2e test when user is not allowed source

Nilay27:
* This is part 2 of the End to end testing for MiniSales contract.
* This contains tests for nested `describe` covering the transfers of purchased PPO by the user in the case `if user purchases and is not allowed source`


ramenforbreakfast:
I think PR 301 and 302 should be closed because we should not forego permissioned sale testing just because a user is blocked or not an allowed source. A PR should follow PR #300 adding permissioned sale testing.

ramenforbreakfast (changes requested):
I think PR 301 and 302 should be closed because we should not forego permissioned sale testing just because a user is blocked or not an allowed source. A PR should follow PR https://github.com/prepo-io/prepo-monorepo/pull/300 adding permissioned sale testing.

Nilay27:
Closing as per the suggestion to include permissioned sale scenario within the `if user purchases and is not allowed source` and avoid unnecessary complex merge conflicts.

PR #302, opened 2022-08-10
feat: minisales e2e test user is allowed source and permissioned sale

Nilay27:
* This is part 3 (final part) of the End to end testing for MiniSales contract.
* This PR contains end-to-end tests of MiniSales for following scenarios:
  1. Nested `describe` when the sale is permissioned with cases if recipient is allowed or not.
  2. One nested `describe` covering the transfers of purchased PPO by the user in the case `if user purchasesand is allowed source`


ramenforbreakfast (changes requested):
I think PR 301 and 302 should be closed because we should not forego permissioned sale testing just because a user is blocked or not an allowed source. A PR should follow PR https://github.com/prepo-io/prepo-monorepo/pull/300 adding permissioned sale testing.

Nilay27:
Closing as per the suggestion to include permissioned sale scenario within the `if user purchases and is allowed source` and avoid unnecessary complex merge conflicts.

PR #303, opened 2022-08-11
fix: unify defender-relay-client and update to 1.27.1 due to old axios pkg

ramenforbreakfast:
Out of the 54 warnings(not all SC related) we received from our package audit, 6 of them were related to OZ defender's relay package using a vulnerable `axios` dependency. This is a package we most likely want used throughout repos when we migrate over to using relays, so I have unified all subfolders in our project to use the same `defender-relay-client`.


Nilay27 (approved):
LGTM!

XavierEkkel (approved):
LGTM

PR #304, opened 2022-08-11
fix: ethereum waffle vulnerabilities

ramenforbreakfast:
Prior to this PR, we had not forced our core repo to use the updated chai matchers for Hardhat. To refresh everyone's memory, we switched over because revert matching in the old matching library does inclusive matching rather than exact matching, not ideal for accurate revert tests.

So we had two SC chai matching libraries in parallel. Unfortunately, we can no longer put off the effort to force our core repo to use the new `@nomicfoundation/hardhat-chai-matchers` library because the older `ethereum-waffle` and `@nomiclabs/hardhat-matchers` packages have vulnerabilities. This is the source of 24 out of the ~50 vulnerabilities in the audit.

While it seems like the changes are extensive, the majority of these changes are due to the following things:
- Removal of `ethereum-waffle` and `@nomiclabs/hardhat-waffle` reducing the size of our `yarn.lock`
- Needing to remove usage of `chai.use(solidity)` from every single Core fixture and test suite (this used to be how we accessed Solidity specific chai matchers). There was also one we accidentally kept in in the `token` test suite.

Less verbose changes include:
- Instead of removing the `revertReason` workaround throughout every test case, I have made it just return `reason` to make these changes less verbose.
- In `Collateral.test.ts`, we used to check for panic codes by literally matching the error string, the newer chai matchers library requires using their custom panic code revert matchers.
-  In `apps/smart-contracts/token/utils/index.ts`, there is a type from the `eth-sig-util` library we were using before to generate signatures to test `ERC20Permit`. This library has been updated(not explicitly by us, but due to the removal and update of these packages) and the type was changed.
- Needed to explicitly update our `core` repo `chai` pkg to `4.3.4` to match our `token` repo. This is because for some reason the one we had before, `4.3.3`, breaks the matchers completely.


XavierEkkel (approved):
LGTM after removing the empty MiniSales.ts file that was committed

PR #306, opened 2022-08-11
docs: withdrawerc20 and safeownable interface copy

ramenforbreakfast:


packages/prepo-shared-contracts/contracts/interfaces/ISafeOwnable.sol
```diff
@@ -1,13 +1,31 @@
...
 
+  /**
+   * @notice Nominate an address to be owner of a contract.
+   * @dev Only callable by `owner()`.
+   * @param nominee The address that will be nominated
+   */
   function transferOwnership(address nominee) external;
 
+  /**
+   * @notice Accepts ownership nomination.
```
Nilay27, line 24:
```suggestion
   * @notice Allows nominated address to accept ownership.
```

packages/prepo-shared-contracts/contracts/interfaces/ISafeOwnable.sol
```diff
@@ -1,13 +1,31 @@
...
+   * @dev Only callable by the current nominee.
```
Nilay27, line 25: Can also add a dev comment that sets the new nominee to zero address

packages/prepo-shared-contracts/contracts/interfaces/ISafeOwnable.sol
```diff
@@ -1,13 +1,31 @@
...
+   * @dev Emitted via `transferOwnership()`.
+   * @param previousNominee The previous nominee
+   * @param newNominee The new nominee
+   */
   event NomineeUpdate(address indexed previousNominee, address indexed newNominee);
 
+  /**
+   * @notice Nominate an address to be owner of a contract.
+   * @dev Only callable by `owner()`.
+   * @param nominee The address that will be nominated
```
Nilay27, line 19: Since we are using "address" for the `address` parameter type here, we should be consistent with the same for other address parameters in all methods.

packages/prepo-shared-contracts/contracts/interfaces/ISafeOwnable.sol
```diff
@@ -1,13 +1,31 @@
...
 interface ISafeOwnable {
+  /**
+   * @dev Emitted via `transferOwnership()`.
+   * @param previousNominee The previous nominee
+   * @param newNominee The new nominee
+   */
   event NomineeUpdate(address indexed previousNominee, address indexed newNominee);
 
+  /**
+   * @notice Nominate an address to be owner of a contract.
```
Nilay27, line 17:
```suggestion
   * @notice Nominates an address to be owner of a contract.
```

packages/prepo-shared-contracts/contracts/interfaces/ISafeOwnable.sol
```diff
@@ -1,13 +1,31 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 pragma solidity =0.8.7;
 
-// TODO: add natspec comments
+/**
+ * @notice An extension of OpenZeppelin's `Ownable.sol` contract that requires
+ * an address to accept a nomination before ownership is transferred.
```
Nilay27, line 6: Actually, the 'acceptance' is for ownership of the contract. The address is nominated by by default through `_setNominee` which is called via `transferOwnership` internally. 
We can have something like this:
```suggestion
 * @notice An extension of OpenZeppelin's `Ownable.sol` contract that requires
 * an address to be nominated in order to accept ownership.
```

Nilay27 (approved):
LGTM after comments.

XavierEkkel (approved):
LGTM after nits

packages/prepo-shared-contracts/contracts/interfaces/ISafeOwnable.sol
XavierEkkel:
```suggestion
   * @notice Nominates an address to be owner of the contract.
```

PR #309, opened 2022-08-11
feat: adding deployment script for TransferHook contracts

Nilay27:
* Adding deployment scripts for `BlocklistTransferHook` and `RestrictedTransferHook`.
* Since these are not upgradeable contracts, hence the script follows the format of the core deployment scripts.


apps/smart-contracts/token/deploy/BlocklistTransferHook.ts
```diff
@@ -0,0 +1,62 @@
...
+    console.log('Deployed BlocklistTransferHook to', blocklistTransferHookAddress)
+  } else {
+    console.log('Existing BlocklistTransferHook at', blocklistTransferHookAddress)
+  }
+  console.log('')
+}
+
+export default deployFunction
+
+deployFunction.dependencies = ['AccounList']
```
ramenforbreakfast, line 60:
```suggestion
deployFunction.dependencies = ['AccountList']
```

Typo aside, we should leave this out as a dependency. Dependencies make sense when there is a series of linear contract dependencies where each contract is only meant to be deployed once. 

We are going to need multiple different `AccountList` so dependencies don't make sense here.

Instead, we should add another section before deploying `BlocklistTransferHook`, to see if there is an `AccountList` deployed already specifically for the blocklist, similarly to how we do with `PPO`. The name we would fetch would be a name that we'd deploy an `AccountList` under, so maybe we'd do `getOrNull('Blocklist-AccountList')`.

This is why also, I would suggest writing the `AccountList` deploy script before hand, since a reviewer wouldn't even be able to run this without a `AccountList` deployment script.

apps/smart-contracts/token/deploy/RestrictedTransferHook.ts
```diff
@@ -0,0 +1,62 @@
...
+    console.log('Deployed RestrictedTransferHook to', restrictedTransferHookAddress)
+  } else {
+    console.log('Existing RestrictedTransferHook at', restrictedTransferHookAddress)
+  }
+  console.log('')
+}
+
+export default deployFunction
+
+deployFunction.dependencies = ['AccounList']
```
ramenforbreakfast, line 60: Same here, get rid of this dependency, before the contract is deployed, fetch a `SourceAllowlist-AccountList` and `DestinationAllowlist-AccountList`, and fail if they don't exist.

ramenforbreakfast (changes requested):
Please close these PRs, since it makes no sense to submit this before we even have an `AccountList` deployment script.

XavierEkkel (commented):
will hold off on review until built off https://github.com/prepo-io/prepo-monorepo/pull/310

Nilay27:
Included @ramenforbreakfast's suggestion to check for Blocklist, Source and Destination allowlist first before deploying the respective transfer hooks. 

Also removed `transferOwnership()` from the constructor of `BlocklistTransferHook` and `RestrictedTransferHook` as per the decision to set it afterwards during the deployment.

ramenforbreakfast (approved):
LGTM, I have amended the error messages to be more clear and moved transfer of ownership to the very end.

PR #310, opened 2022-08-12
feat: add AccountList deployment script

ramenforbreakfast:
This is pretty much the same as most non-upgradeable deployment scripts, except the addition of a variable `nameOfAccountList` to store a deployment under, since there will be multiple `AccountList` instances and we need a unique identifying name of each one.

This needs to be changed for each new deployment, to retrieve it for later.


apps/smart-contracts/token/deploy/AccountList.ts
```diff
@@ -0,0 +1,63 @@
...
+const { assertIsTestnetChain, sendTxAndWait } = utils
+
+const deployFunction: DeployFunction = async function deployAccountList({
+  ethers,
+  deployments,
+  getChainId,
+}: HardhatRuntimeEnvironment): Promise<void> {
+  const { deploy, getOrNull } = deployments
+  const deployer = (await ethers.getSigners())[0]
+  console.log('Running AccountList deployment script with', deployer.address, 'as the deployer')
```
XavierEkkel, line 23: like previous comment, also consider putting nameOfAccountList here instead of just `AccountList`

XavierEkkel (approved):
LGTM after comments

apps/smart-contracts/token/deploy/AccountList.ts
ramenforbreakfast: I will do so when we inform the user of when something is deployed or existing, since that should be particular to the name of the specific deployment.

Here doesn't make much sense since this statement pertains to the script and not a particular deployment.

PR #312, opened 2022-08-13
chore: remove purchase hook contract and usage in MiniSales test

ramenforbreakfast:
`PurchaseHook.sol` was originally introduced to not conflict with `MiniSales` development. Now that the `MiniSales` contract is complete, we can remove this.


XavierEkkel (approved):
LGTM

PR #316, opened 2022-08-14
chore: remove revertReason from token repo

Nilay27:
* As per the backlog tasks, removing revertReason from the token repository.
* Will be adding missing `await` to the reverts in a followup PR to make changes in this PR less verbose.


ramenforbreakfast (approved):
LGTM, tests pass

PR #318, opened 2022-08-14
chore: changing license identifier for token repo and shared packages

Nilay27:
* Changing SPDX licence indentifier from `AGPL-3.0-or-later` to `AGPL-3.0` for token repo contracts (except for staking contracts).


ramenforbreakfast (approved):
Approved! I have also added AGPL-3.0 licenses to staking contracts

PR #319, opened 2022-08-14
chore: changed initial state test names and vesting inherit withdrawERC20

Nilay27:
* This PR contains two backlog items:
  1. Changing `describe` and test names for initial states. In many of the non-upgradeable contracts, in the tests for initial state, we had the names as `sets [argument] from initialize`, which is wrong given that non-upgradeable contracts do not have an `initialize` method. Hence I changed it to `sets [argument] from constructor` (as per our convention for non-upgradeable contracts).
  2. Making Vesting inherit shared `WithdrawERC20` and removed it from the contract and interface and the corresponding tests. Moreover, removed import of `SafeOwnable.sol` as it is already included in `WithdrawERC20`


apps/smart-contracts/token/contracts/vesting/Vesting.sol
```diff
@@ -4,10 +4,10 @@ pragma solidity =0.8.7;
 import "@openzeppelin/contracts/security/ReentrancyGuard.sol";
 import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 import "./interfaces/IVesting.sol";
-import "prepo-shared-contracts/contracts/SafeOwnable.sol";
 import "prepo-shared-contracts/contracts/Pausable.sol";
+import "prepo-shared-contracts/contracts/WithdrawERC20.sol";
 
-contract Vesting is IVesting, SafeOwnable, ReentrancyGuard, Pausable {
+contract Vesting is IVesting, ReentrancyGuard, Pausable {
```
ramenforbreakfast: You need to inherit `WithdrawERC20`, or it won't have access to the contract's functions. I will do this myself in a commit and add back in a sanity test to check that the function is accessible.

ramenforbreakfast (approved):
I have made a few additions:
- The `Vesting` contract needs to inherit `WithdrawERC20` or else it won't have access to the contract's methods, inheriting is not sufficient.
- For simplicity, I have added back two tests for the inherited `withdrawERC20` just to at least verify the function is there.

PR #320, opened 2022-08-15
fix: remove owner nomination in  safeOwnable contracts

Nilay27:
* removing owner nomination in constructors/initializations for SafeOwnable
* This PR contains all the removal of `transferOwnership()` for all the Token repo contracts, except for `BlocklistTransferHook` and `RestrictedTransferHook` as they are already covered in PR #309 
* I have also modified the test and deployment scripts accordingly.

Note: 
I did not add transfer of Ownership in the deployment for `PPO`; @ramenforbreakfast, could you confirm whether we need that in `PPO.ts` or not, as I think the `ensureProxyOwnedByGovernance()` method takes care of that?


ramenforbreakfast (approved):
Removed owner nominations from deployment scripts as well since this was forgotten.

Also owner nomination was removed in TransferHook contracts in PR 309, but I think it makes more sense to do it here.

PR #321, opened 2022-08-15
chore: token event naming rectification

Nilay27:
* Changing event names as per our convention for token contracts.
* Note: Vesting did not have a test for emitting `Claim`; I have added that also.


ramenforbreakfast (approved):
LGTM, squeezed in the missing AGPL conversion we spoke about earlier.

PR #322, opened 2022-08-15
test: adding tests for when variables are not set

Nilay27:
* Adding tests for when a variable is not set but is getting used inside other methods.

* I found three such cases, in AllowlistPurchaseHook, BlocklistTransferHook and RestrictedTransferHook, where we check for ` isIncluded()` in respective Accountlists, in such cases if the account list is not set or incorrectly set, the methods should revert.


ramenforbreakfast (approved):
LGTM, I have changed the testname `reverts if correct <list> not set`, since what this is actually testing for is if a contract that does not have a `hook` function that follows the `PurchaseHook` or `TransferHook` interface.

PR #323, opened 2022-08-16
feat: adding implementation for transferFromWithPermit in PPO

Nilay27:
* Adding implementation of `transferFromWithPermit()` in `PPO.sol` along with natspecs in `IPPO.sol`.
* Also added utility function `getPermitSignature()` to generate `v,r,s` to `transferFromWithPermit()`.
* Splitting this PR into two parts : 1 for implementation and another for tests to avoid making this PR too long.

PR #324, opened 2022-08-16
test: adding tests for transferFromWithPermit in PPO

Nilay27:
* Adding test for `transferFromWithPermit()`. 
* Although this PR is big in terms of numbers of lines, but that is mostly because of following multiline calls :
```
await getPermitSignature(
        user1,
        ppo,
        user2.address,
        user1PPOBalanceBefore,
        deadline
      )

and 

await ppo
        .connect(spender)
        .transferFromWithPermit(
          user1.address,
          user2.address,
          user1PPOBalanceBefore,
          deadline,
          v,
          r,
          s
        )
```

Otherwise, it is not dissimilar to tests for usual transfer/transferFrom methods with a few added scenarios. Hence I decided to submit it in one go.


apps/smart-contracts/token/test/PPO.test.ts
```diff
@@ -3,23 +3,27 @@ import { ethers } from 'hardhat'
...
+const { getLastTimestamp, setNextTimestamp } = utils
+
 chai.use(smock.matchers)
 
 describe('=> PPO', () => {
   let deployer: SignerWithAddress
   let owner: SignerWithAddress
   let user1: SignerWithAddress
   let user2: SignerWithAddress
+  let spender: SignerWithAddress
```
ramenforbreakfast, line 21: I think better to move this to `transferFromWithPermit`, and instead of having a unique signer as `spender`, let's just set `spender = user2` or something.

apps/smart-contracts/token/test/PPO.test.ts
```diff
@@ -445,6 +449,402 @@ describe('=> PPO', () => {
...
+      )
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      await setNextTimestamp(ethers.provider as any, timeAfterDeadlineExpiration)
+
+      await expect(
+        ppo
+          .connect(spender)
+          .transferFromWithPermit(
+            user1.address,
+            user2.address,
```
ramenforbreakfast, line 481:
```suggestion
            user2.address,
```
This is a bad test because you've generating a permit signature for `spender.address` to spend, yet passed in `user2.address`, so it would fail even with a good deadline.

This is why i keep emphasizing you need to write your tests in a way that makes it clear what inputs are changing if your tests change inputs alot, especially with a function that has this many parameters and ones that aren't obvious. Like `timeAfterDeadlineExpiration` is fine because its very clear that this time is past the deadline since we can see it is `deadline + BLOCK_DURATION_IN_SECONDS`, so there is also no need for an expect.

It's harder with an input like an account because you are juggling 3 accounts around all the time.

apps/smart-contracts/token/test/PPO.test.ts
```diff
@@ -445,6 +449,402 @@ describe('=> PPO', () => {
...
+            user1PPOBalanceBefore.add(1),
+            deadline,
+            v,
+            r,
+            s
+          )
+      ).revertedWith('ERC20: transfer amount exceeds balance')
+    })
+
+    it('transfers to recipient if recipient is not spender', async () => {
```
ramenforbreakfast, line 623: This is related to PR #323, but this is a perfect example of why it makes no sense. The recipient _is_ the spender. I am not sure in what scenario you would write a signature to approve a spend and then send tokens to someone else?

If you agree with this point, lets remove these `if recipient not spender` and `if recipient is spender` cases and leave it with just the `transfers to recipient...` cases.

apps/smart-contracts/token/test/PPO.test.ts
```diff
@@ -445,6 +449,402 @@ describe('=> PPO', () => {
...
+      const tx = await ppo
+        .connect(spender)
+        .transferFromWithPermit(user1.address, user2.address, BigNumber.from(0), deadline, v, r, s)
+
+      await expect(tx)
+        .to.emit(ppo, 'Transfer(address,address,uint256)')
+        .withArgs(user1.address, user2.address, 0)
+    })
+
+    it('emits transfer if amount = 0 and recipient is spender', async () => {
```
ramenforbreakfast, line 779:
```suggestion
    it('emits transfer if amount = 0', async () => {
```
Can remove the other test case if no security concerns with PR #323 suggestion. Please do this for other cases.

ramenforbreakfast (changes requested):
I think this test suite needs some more changes before I can approve.

apps/smart-contracts/token/test/PPO.test.ts
Nilay27: I don't see how you concluded that the test will fail even with a good deadline when it doesn't (the test does not revert if you pass in `deadline` to `setNextTimestamp`. 

As per the current implementation, a user (`user1`) approves an account to spend its tokens and then the approved account can transfer the tokens to a recipient (other than user1 also), here the approved account is `spender` hence the spender can now transfer those tokens to another user (`user2`)

ramenforbreakfast: Ah yes this is my confusion, I forgot that the `spender` can send to another address, this is my bad.

ramenforbreakfast (approved):

ramenforbreakfast:
I for some reason had a brain fart and thought the spender and recipient had to be the same for `transferFrom`, I have dismissed my comments and have approved the test suite.

apps/smart-contracts/token/test/PPO.test.ts
```diff
@@ -445,6 +449,402 @@ describe('=> PPO', () => {
...
+            user1PPOBalanceBefore,
+            deadline,
+            v,
+            r,
+            s
+          )
+      ).revertedWith('ERC20Permit: expired deadline')
+    })
+
+    it('reverts if invalid signature', async () => {
```
XavierEkkel, line 491: i think this should be split into:
- `reverts if from address different from signature`
- `reverts if to address different from signature`
- `reverts if deadline different from signature`
- `reverts if invalid v`
- `reverts if invalid r`
- `reverts if invalid s`

XavierEkkel (approved):
LGTM after comments

PR #326, opened 2022-08-17
feat: add data payload to purchase hook

ramenforbreakfast:
This PR adds the data payload field to `AllowlistPurchaseHook`'s `hook` function. 

- I made it a `bytes32` because if it were to be a dynamic `bytes` parameter, we would need to make our contract use the experimental ABIEncoderV2 in order for the `hook` to be usable.

- Since the compiler forces me to have complete interface descriptions for the interface definitions that already existed and were just modified, I decided to just leave all the interface copy in this PR since there isn't much to begin with.

Majority of the verboseness of this PR is attributed to the following things:

- Updated the `AllowlistPurchaseHook.test.ts` to take into account the new parameter and verify that the `Referral` is only emitted when the data payload is not empty.

- Updated the `MiniSales.test.ts` to take into the account the new parameter and verify that it passes it along correctly to the `hook`.


apps/smart-contracts/token/test/AllowlistPurchaseHook.test.ts
```diff
@@ -102,18 +104,41 @@ describe('AllowlistPurchaseHook', () => {
...
+        allowlistPurchaseHook,
+        'Referral(address,uint256,uint256,bytes32)'
+      )
+    })
+
+    it('emits Referral when data is not empty', async () => {
+      allowedAccounts.isIncluded.whenCalledWith(recipient.address).returns(true)
+      const tx = await allowlistPurchaseHook
+        .connect(allowedContract)
+        .hook(purchaser.address, recipient.address, 1, 2, formatBytes32String('P'))
```
Nilay27, line 136: Something like `DUMMY_DATA` or `JUNK_DATA` would be better than an arbitrary alphabet. 
Same for other places.

Nilay27 (approved):
LGTM beside nit

apps/smart-contracts/token/test/AllowlistPurchaseHook.test.ts
```diff
@@ -102,18 +104,41 @@ describe('AllowlistPurchaseHook', () => {
...
+        .hook(purchaser.address, recipient.address, 1, 1, ZERO_HASH)
 
       expect(allowedAccounts.isIncluded).to.have.been.calledWith(recipient.address)
     })
+
+    it("doesn't emit Referral when data is empty", async () => {
+      allowedAccounts.isIncluded.whenCalledWith(recipient.address).returns(true)
+      const tx = await allowlistPurchaseHook
+        .connect(allowedContract)
+        .hook(purchaser.address, recipient.address, 1, 2, ZERO_HASH)
```
XavierEkkel, line 124: i think instead of magic numbers ('1', '2'), we should use a amount+price consts for all these tests

apps/smart-contracts/token/test/AllowlistPurchaseHook.test.ts
```diff
@@ -102,18 +104,41 @@ describe('AllowlistPurchaseHook', () => {
...
       allowedAccounts.isIncluded.whenCalledWith(recipient.address).returns(true)
       await allowlistPurchaseHook
         .connect(allowedContract)
-        .hook(purchaser.address, recipient.address, 1, 1)
+        .hook(purchaser.address, recipient.address, 1, 1, ZERO_HASH)
 
       expect(allowedAccounts.isIncluded).to.have.been.calledWith(recipient.address)
     })
+
+    it("doesn't emit Referral when data is empty", async () => {
```
XavierEkkel, line 120: nit: use `if` not `when`

same for other tests

apps/smart-contracts/token/test/AllowlistPurchaseHook.test.ts
```diff
@@ -102,18 +104,41 @@ describe('AllowlistPurchaseHook', () => {
...
+        .connect(allowedContract)
+        .hook(purchaser.address, recipient.address, 1, 2, ZERO_HASH)
+
+      await expect(tx).to.not.emit(
+        allowlistPurchaseHook,
+        'Referral(address,uint256,uint256,bytes32)'
+      )
+    })
+
+    it('emits Referral when data is not empty', async () => {
```
XavierEkkel, line 132: nit:
```suggestion
    it('emits Referral when data not empty', async () => {
```

apps/smart-contracts/token/contracts/mini-sales/interfaces/IPurchaseHook.sol
```diff
@@ -9,15 +9,22 @@ pragma solidity =0.8.7;
 interface IPurchaseHook {
   /**
    * @notice Hook to be called when a user makes a `MiniSales` purchase.
+   * @dev A `data` field is provided for passing in arbitrary data to add
+   * additional functionality.
+   *
+   * As of now, this is used to record usage of referral codes for a purchase
+   * of PPO.
```
XavierEkkel, line 16: this is irrelevant to this interface - it should only be mentioned within a child interface

apps/smart-contracts/token/contracts/mini-sales/AllowlistPurchaseHook.sol
```diff
@@ -14,9 +14,13 @@ contract AllowlistPurchaseHook is IAllowlistPurchaseHook, SafeOwnable {
...
     address _recipient,
     uint256 _amount,
-    uint256 _price
+    uint256 _price,
+    bytes32 _data
   ) public virtual override {
     require(_allowlist.isIncluded(_recipient), "Recipient not allowed");
+    if (_data != "") {
+      emit Referral(_recipient, _amount, _price, _data);
+    }
```
XavierEkkel, line 23: i think instead of adding this to the `AllowlistPurchaseHook` and expanding its scope, we should instead have a child contract that inherits from AllowlistPurchaseHook and overrides the hook to add in this part

ramenforbreakfast:
I have applied multiple changes to the implementation after the initial review:
- Made the `bytes32` `_data` parameter a dynamic `bytes` parameter to allow greater flexibility for future expansions.
- It was decided to remove the separate `Referral` event and instead move the `_data` to the `Purchase` event emitted in `MiniSales.sol`. `indexed` event parameters must be fixed size types of up to 32 bytes, so we extract only the first 32 bytes and pass it along to the `Purchase` event. This is kept generic as we might want to use this data payload for something else besides referral codes.
- The entire `_data` payload is still passed along to the `hook`, for future contracts that implement `IPurchaseHook` to take advantage of. However, `AllowlistPurchaseHook` will not use this parameter.

ramenforbreakfast:
The data processing logic has now been moved entirely out of `MiniSales.sol`. Since currently the `AllowlistPurchaseHook` does not involve any data processing logic, this PR now only contains changes to `purchase` function calls in test suites since a new parameter is introduced.

The test verifying that the hook is called correctly now also verifies that a non-zero data payload is passed to it.

apps/smart-contracts/token/contracts/mini-sales/interfaces/IMiniSales.sol
```diff
@@ -57,11 +57,13 @@ interface IMiniSales {
    * @param recipient Address that sale token will be delivered to
    * @param amount Amount of sale token to be purchased
    * @param price Sale token price in terms of payment token
+   * @param data Arbitrary data payload
```
XavierEkkel, line 60:
```suggestion
   * @param data Data payload passed on to the hook to support additional functionality
```

apps/smart-contracts/token/contracts/mini-sales/interfaces/IPurchaseHook.sol
```diff
@@ -9,15 +9,19 @@ pragma solidity =0.8.7;
 interface IPurchaseHook {
   /**
    * @notice Hook to be called when a user makes a `MiniSales` purchase.
+   * @dev A `data` field is provided for passing in arbitrary data to add
+   * additional functionality.
    * @param purchaser Address that payment token was taken from
    * @param recipient Address that sale token was delivered to
    * @param amount Amount of sale token purchased
    * @param price Sale token price in terms of payment token
+   * @param data Arbitrary data payload
```
XavierEkkel, line 18:
```suggestion
   * @param data Data payload for supporting additional hook functionality
```

apps/smart-contracts/token/contracts/mini-sales/interfaces/IPurchaseHook.sol
```diff
@@ -9,15 +9,19 @@ pragma solidity =0.8.7;
 interface IPurchaseHook {
   /**
    * @notice Hook to be called when a user makes a `MiniSales` purchase.
+   * @dev A `data` field is provided for passing in arbitrary data to add
+   * additional functionality.
```
XavierEkkel, line 13:
```suggestion
```

XavierEkkel (approved):
LGTM after nits

PR #327, opened 2022-08-17
fix: changed prettierrc to fix solhint max length warning

Nilay27:
* While going through the solhint warning, I saw that in the token repo we had 53 warnings just related to `max-line-length` exceeded. As per our conventions, we follow the max length of 79.  
* While ideally it should have been taken care of by running `yarn sl` but apparently prettier was not wrapping up functions and codes even if they exceeded 79 chars, rather it was wrapping only when they exceeded 99 chars.
* After debugging it for a long time and discussing it with the FE team, we found the solution. It's just a single line change, we needed to explicitly add `"printWidth": 79` to the `.prettierrc` configuration for solidity. 
* There are no functional changes in this PR, it only contains wrapping up of codes which exceeded 79 char limit.

* **Note**: 
The warning still persisted for two cases where we were importing OZ upgradeable libraries in `PPO.sol`, as they are strings and cannot be wrapped. I added `//solhint-disable-next-line max-line-length` to remove those warnings.


ramenforbreakfast (approved):
LGTM, thanks for addressing this!

PR #331, opened 2022-08-18
chore: tokenshop inherit generic withdraw tokens

Nilay27:
* Making `TokenShop` inherit shared `WithdrawERC20`, `WithdrawERC721` and `WithdrawERC1155` contracts and removed them from the contract and interface and the corresponding tests. 
* Moreover, removed import of `SafeOwnable.sol` and `ReentrancyGuard` as it they are already included in generic contracts.
* I've kept minimal tests for generic withdraw codes and left onERC721Recieved as per the discussion in PR #267.
* Also there was one bug in the code, `getERC721PurchaseCount` wasn't actually returning anything, have resolved this issue too.


ramenforbreakfast (approved):
LGTM, I added a minor change to include a line to ensure balances we are using are greater than 0.

PR #333, opened 2022-08-18
feat: adding SafeOwnable caller contract implementation

Nilay27:
* Adding implementation for the `SafeOwnableCaller` contract.
* Had to also add `renounceOwnership()` in `ISafeOwnable` as for any child contract inheriting `SafeOwnableCaller`, to be able to renounce ownership of a contract, it needs to call upon `renounceOwnership()` of that contract.
* Also implemented `renounceOwnership` in both `SafeOwnable` and `SafeOwnableUpgradeable`. 
* Added test for `renounceOwnership` in this PR only as they are basically identical.

**Note**:
In the implementation of `renounceOwnership()` we are also calling `_setNominee(address(0))`. This is to avoid bypassing of the `renounceOwnership` functionality.

A scenario where it could be bypassed is if the current owner of the contract calls `transferOwnership()` before renouncing the ownership. In such case even if the `owner` is set to ZeroAddress, the `nominee` will not be zero-address and could accept ownership of the contract; thus defeating the purpose of `renounceOwnerShip()`.

Hence we also set nominee to zero and therefore ensure that ownership of the contract can never be reclaimed.


packages/prepo-shared-contracts/contracts/SafeOwnableCaller.sol
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: AGPL-3.0
+pragma solidity =0.8.7;
+
+import "./interfaces/ISafeOwnable.sol";
+import "./interfaces/ISafeOwnableCaller.sol";
+
+contract SafeOwnableCaller is ISafeOwnableCaller {
+  function transferOwnership(address _nominee, address _safeOwnableContract)
```
ramenforbreakfast, line 8:
```suggestion
  function transferOwnership(address _safeOwnableContract, address _nominee)
```

ramenforbreakfast (approved):
LGTM, besides a nit, but are we adding tests for `SafeOwnableCaller`

PR #334, opened 2022-08-18
feat: adding interface and contract for SafeAccessControl

Nilay27:
* Adding interface + implementation for `SafeAccessControl`. 
* Two critical methods which required nominate + accept flow were:
    1. `grantRole()`, which grants a particular `Role` to an `account`. This has been changed such that calling `grantRole()` will set a nominee for the role by calling `_setRoleNominee()`. The nominated address can then accept the role.
    2. `_setRoleAdmin()`, this sets a `Role` as an admin for another `Role`. Although `_setRoleAdmin()` is an internal method in the OZ contract, but in case if any future contracts inheriting `SafeAccessControl` implements `setRoleAdmin()` then it should ideally be through a nominate + accept flow. 
  
 **Example:**
 
With the current nominate + accept flow, lets say if `Role A` is admin of `Role B`, and someone from `Role A` decides that the `new Admin Role` for `Role B` should be a `Role C`, then it will only nominate `Role C` as the new admin role for `Role B`. An account from `Role C` will be required to accept the admin role for `Role B`.   
 
Only after `Role C` accepts the admin role for `Role B`, will it be set as the `new Admin` for `Role B`.

`Role A` will no longer be the admin of `Role B` anymore


ramenforbreakfast:
> * Adding interface + implementation for `SafeAccessControl`.
> * Two critical methods which required nominate + accept flow were:
>   
>   1. `grantRole()`, which grants a particular `Role` to an `account`. This has been changed such that calling `grantRole()` will set a nominee for the role by calling `_setRoleNominee()`. The nominated address can then accept the role.
>   2. `_setRoleAdmin()`, this sets a `Role` as an admin for another `Role`. Although `_setRoleAdmin()` is an internal method in the OZ contract, but in case if any future contracts inheriting `SafeAccessControl` implements `setRoleAdmin()` then it should ideally be through a nominate + accept flow. With the current nominate + accept flow, lets say if `Role A` is admin of `Role B`, and someone from `Role A` decides that the `new Admin Role` for `Role A` should be a `Role C`, then it will only nominate `Role C` as the new admin role for `Role A`. An account from `Role C` will be required to accept the admin role for `Role B`

If I understand correctly what you are trying to explain here is that if a role is an admin of another admin role, the role will not automatically change the admin of the "child" roles even though effectively, since they are the "parent" admin, they are _de-facto_ in control of all child roles.

If that is the case, that is OK.

ramenforbreakfast (commented):
So this confused me for awhile. For `Ownable` contracts, OpenZeppelin by default will initialize the `owner` to the deployer. 

For `AccessControl.sol`, it is a little more complicated, since by default, the `adminRole` for every `Role` is `0x00`. So in order for the contract to be usable, this `0x00` role needs to be initialized to an accessible account, most likely the deployer.

Perhaps for flexibility, OpenZeppelin doesn't automatically initialize the `0x00` role for `AccessControl`, you need to do it yourself or the contract becomes unusable.

I would suggest we do this by adding a constructor like so
```
constructor() {
    _setupRole(DEFAULT_ADMIN_ROLE, _msgSender());
}
```

ramenforbreakfast:
> * Adding interface + implementation for `SafeAccessControl`.
> * Two critical methods which required nominate + accept flow were:
>   
>   1. `grantRole()`, which grants a particular `Role` to an `account`. This has been changed such that calling `grantRole()` will set a nominee for the role by calling `_setRoleNominee()`. The nominated address can then accept the role.
>   2. `_setRoleAdmin()`, this sets a `Role` as an admin for another `Role`. Although `_setRoleAdmin()` is an internal method in the OZ contract, but in case if any future contracts inheriting `SafeAccessControl` implements `setRoleAdmin()` then it should ideally be through a nominate + accept flow. With the current nominate + accept flow, lets say if `Role A` is admin of `Role B`, and someone from `Role A` decides that the `new Admin Role` for `Role A` should be a `Role C`, then it will only nominate `Role C` as the new admin role for `Role A`. An account from `Role C` will be required to accept the admin role for `Role B`

Just want to be clear, you are trying to communicate here that the "parent" of a "child" role will not automatically be the admin of the "children" of that "child" role. Although in reality, since the "parent" can decide the "child" role, they are effectively in control of all descendant roles.

ramenforbreakfast:
I have updated the branch to include the constructor I spoke about earlier.

packages/prepo-shared-contracts/contracts/interfaces/ISafeAccessControl.sol
```diff
@@ -0,0 +1,28 @@
...
+interface ISafeAccessControl {
+  event RoleAdminNomineeUpdate(bytes32 role, bytes32 newRoleAdminNominee);
+
+  event RoleNomineeUpdate(
+    bytes32 role,
+    address account,
+    bool nominationStatus
+  );
+
+  function setRoleAdmin(bytes32 role, bytes32 adminRoleNominee) external;
```
XavierEkkel, line 14:
```suggestion
  function setRoleAdminNominee(bytes32 role, bytes32 adminRoleNominee) external;
```
can be clearer about what it's actually doing

packages/prepo-shared-contracts/contracts/interfaces/ISafeAccessControl.sol
```diff
@@ -0,0 +1,28 @@
...
+
+  function acceptRoleAdmin(bytes32 role) external;
+
+  function grantRole(bytes32 role, address account) external;
```
XavierEkkel, line 18: this is already in https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/access/IAccessControl.sol

packages/prepo-shared-contracts/contracts/interfaces/ISafeAccessControl.sol
```diff
@@ -0,0 +1,28 @@
+// SPDX-License-Identifier: AGPL-3.0
+pragma solidity =0.8.7;
+
+//TODO: add natspecs
+interface ISafeAccessControl {
```
XavierEkkel, line 5: this should inherit from https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/access/IAccessControl.sol

packages/prepo-shared-contracts/contracts/interfaces/ISafeAccessControl.sol
```diff
@@ -0,0 +1,28 @@
...
+  event RoleAdminNomineeUpdate(bytes32 role, bytes32 newRoleAdminNominee);
+
+  event RoleNomineeUpdate(
+    bytes32 role,
+    address account,
+    bool nominationStatus
```
XavierEkkel, line 11: whats nomination status for?
I think we should use the same 'previous' and 'new' format as: https://github.com/prepo-io/prepo-monorepo/blob/main/packages/prepo-shared-contracts/contracts/interfaces/ISafeOwnable.sol#L15

packages/prepo-shared-contracts/contracts/SafeAccessControl.sol
```diff
@@ -0,0 +1,100 @@
...
+    _grantRole(_role, _msgSender());
+    _setRoleNominee(_role, _msgSender(), false);
+  }
+
+  function getRoleAdminNominee(bytes32 _role)
+    public
+    view
+    virtual
+    override
+    returns (bytes32)
```
XavierEkkel, line 66: i think this should be returning address?

packages/prepo-shared-contracts/contracts/SafeAccessControl.sol
```diff
@@ -0,0 +1,100 @@
+// SPDX-License-Identifier: AGPL-3.0
+pragma solidity =0.8.7;
+
+import "@openzeppelin/contracts/access/AccessControl.sol";
+import "./interfaces/ISafeAccessControl.sol";
+
+contract SafeAccessControl is ISafeAccessControl, AccessControl {
+  mapping(bytes32 => bytes32) private _roleToRoleAdminNominee;
```
XavierEkkel, line 8: shouldnt this be a mapping to address?

packages/prepo-shared-contracts/contracts/SafeAccessControl.sol
```diff
@@ -0,0 +1,100 @@
+// SPDX-License-Identifier: AGPL-3.0
+pragma solidity =0.8.7;
+
+import "@openzeppelin/contracts/access/AccessControl.sol";
+import "./interfaces/ISafeAccessControl.sol";
+
+contract SafeAccessControl is ISafeAccessControl, AccessControl {
```
XavierEkkel, line 7: lets inherit from https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/access/AccessControlEnumerable.sol instead

packages/prepo-shared-contracts/contracts/SafeAccessControl.sol
```diff
@@ -0,0 +1,100 @@
...
+  mapping(bytes32 => bytes32) private _roleToRoleAdminNominee;
+  mapping(bytes32 => mapping(address => bool)) private _roleToNominee;
+
+  constructor() {
+    _setupRole(DEFAULT_ADMIN_ROLE, _msgSender());
+  }
+
+  modifier onlyRoleNominee(bytes32 _role) {
```
XavierEkkel, line 15: this is only used for one specific function, so we can just move the require statement into that function instead

packages/prepo-shared-contracts/contracts/SafeAccessControl.sol
```diff
@@ -0,0 +1,100 @@
+// SPDX-License-Identifier: AGPL-3.0
+pragma solidity =0.8.7;
+
+import "@openzeppelin/contracts/access/AccessControl.sol";
+import "./interfaces/ISafeAccessControl.sol";
+
+contract SafeAccessControl is ISafeAccessControl, AccessControl {
+  mapping(bytes32 => bytes32) private _roleToRoleAdminNominee;
+  mapping(bytes32 => mapping(address => bool)) private _roleToNominee;
```
XavierEkkel, line 9:
```suggestion
  mapping(bytes32 => mapping(address => bool)) private _roleToAccountToNominated;
```

packages/prepo-shared-contracts/contracts/SafeAccessControl.sol
```diff
@@ -0,0 +1,100 @@
...
+    public
+    virtual
+    override
+    onlyRole(_roleToRoleAdminNominee[_role])
+  {
+    _setRoleAdmin(_role, _roleToRoleAdminNominee[_role]);
+    _setRoleAdminNominee(_role, 0x00);
+  }
+
+  function grantRole(bytes32 _role, address _account)
```
XavierEkkel, line 42: we also need a way to revoke a nomination for a particular account, because nominating a different account doesn't override the old one

packages/prepo-shared-contracts/contracts/SafeAccessControl.sol
```diff
@@ -0,0 +1,100 @@
...
+    public
+    view
+    virtual
+    override
+    returns (bytes32)
+  {
+    return _roleToRoleAdminNominee[_role];
+  }
+
+  function isRoleNominee(bytes32 _role, address _account)
```
XavierEkkel, line 71:
```suggestion
  function isNominated(bytes32 _role, address _account)
```

ramenforbreakfast:
- Added `revokeNomination` to set nomination of an account to false for a specific role.
- Made interfaces inherit each other so there are no duplicate definitions.
- Use `SafeAccessControlEnumerable` rather than `SafeAccessControl`

packages/prepo-shared-contracts/contracts/SafeAccessControlEnumerable.sol
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  function acceptRoleAdmin(bytes32 _role)
+    public
+    virtual
+    override
+    onlyRole(_roleToRoleAdminNominee[_role])
+  {
+    _setRoleAdmin(_role, _roleToRoleAdminNominee[_role]);
+    _setRoleAdminNominee(_role, 0x00);
```
XavierEkkel, line 35: would prefer that we switch the order here (remove old, and then set new)

packages/prepo-shared-contracts/contracts/SafeAccessControlEnumerable.sol
```diff
@@ -0,0 +1,104 @@
...
+    _setRoleNominee(_role, _account, true);
+  }
+
+  function acceptRole(bytes32 _role) public virtual override {
+    require(
+      _roleToAccountToNominated[_role][_msgSender()],
+      "msg.sender != nominee for role"
+    );
+    _grantRole(_role, _msgSender());
+    _setRoleNominee(_role, _msgSender(), false);
```
XavierEkkel, line 53: lets also switch the order here

packages/prepo-shared-contracts/contracts/interfaces/ISafeAccessControlEnumerable.sol
```diff
@@ -0,0 +1,31 @@
...
+  );
+
+  function setRoleAdminNominee(bytes32 role, bytes32 adminRoleNominee)
+    external;
+
+  function acceptRoleAdmin(bytes32 role) external;
+
+  function acceptRole(bytes32 role) external;
+
+  function revokeNomination(bytes32 _role, address _account) external;
```
XavierEkkel, line 23:
```suggestion
  function revokeNomination(bytes32 role, address account) external;
```

packages/prepo-shared-contracts/contracts/SafeAccessControlEnumerable.sol
```diff
@@ -0,0 +1,104 @@
...
+    override
+    onlyRole(getRoleAdmin(_role))
+  {
+    _setRoleNominee(_role, _account, true);
+  }
+
+  function acceptRole(bytes32 _role) public virtual override {
+    require(
+      _roleToAccountToNominated[_role][_msgSender()],
+      "msg.sender != nominee for role"
```
XavierEkkel, line 50:
```suggestion
      "msg.sender != role nominee"
```

packages/prepo-shared-contracts/contracts/interfaces/ISafeAccessControlEnumerable.sol
```diff
@@ -0,0 +1,31 @@
...
+import "@openzeppelin/contracts/access/IAccessControlEnumerable.sol";
+
+//TODO: add natspecs
+interface ISafeAccessControlEnumerable is IAccessControlEnumerable {
+  event RoleAdminNomineeUpdate(bytes32 role, bytes32 newRoleAdminNominee);
+
+  event RoleNomineeUpdate(
+    bytes32 role,
+    address account,
+    bool nominationStatus
```
XavierEkkel, line 13:
```suggestion
    bool nominated
```

XavierEkkel (approved):
LGTM after comments

PR #348, opened 2022-08-23
fix: removing initialize from SafeAccessControlEnumerableUpgradeable

Nilay27:
Removing `initialize` from SafeAccessControlUpgradeable so that inheriting contracts can call their own `initialize` method.
Reasoning for this can be found in PR #211 .

Also, setting up the admin for `DEFAULT_ADMIN_ROLE` is necessary during the initialization; hence we defined an internal function `__safeAccessControlEnumerable_init` that needs to be called by the inheriting contracts during initialization.


ramenforbreakfast (approved):
Please move this to the top where the `initialize` function used to be

ramenforbreakfast:
Applied multiple changes
- Moved `internal` initialization function to the top and made it not camel-case to conform with Upgradeable nomenclature for initializers.
- A test had the wrong revert message, probably from a contract nit that was merged in parallel.
- Reorder `acceptRoleAdmin` for the Upgradeable contract.

ramenforbreakfast:
Added in fix to test suite to prevent non-deterministic failure of tests

PR #350, opened 2022-08-24
test: safe ownable caller test transfer and accept ownership

Nilay27:
* Adding test for `transferOwnership()` and `acceptOwnership()` for `SafeOwnableCaller`.
* Added a wrapper contract `SafeOwnableCallerTest` which overrides all the non-view methods of `SafeOwnableCaller` with `onlyOwner` modifier.
* As both `SafeOwnableCaller` and `SafeOwnableCallerTest` have same method names thus causing function overloading, hence in the tests we need to call the methods in the following format:
`await contract.['methodName(argument1Type,argument2Type)'](argument1, argument2)`.
* Also, I have used `MockContract` instead of `FakeContract` for the owned contract because even if `FakeContract` would have been sufficient for testing `transferOwnership()` and `acceptOwnership()` but for testing `getNominee()` (to be submitted in the followup PR) we would need to fetch the values of the nominee in the owned contract. Hence we would need the access to the state variable `_nominee` of the Owned contract.


ramenforbreakfast (approved):
LGTM besides nits

packages/prepo-shared-contracts/test/SafeOwnableCaller.test.ts
```diff
@@ -0,0 +1,123 @@
...
+
+    it('reverts if not owner', async () => {
+      expect(await safeOwnableCaller.owner()).to.not.eq(user1.address)
+
+      await expect(
+        safeOwnableCaller.connect(user1)['acceptOwnership(address)'](mockOwnedContract.address)
+      ).revertedWith('Ownable: caller is not the owner')
+    })
+
+    it("reverts if owned contract's acceptOwnership reverts", async () => {
```
ramenforbreakfast, line 101:
```suggestion
    it("reverts if contract's acceptOwnership reverts", async () => {
```
No need for specifying owned contract, since any contract can be passed in and we'd still expect a revert to be passed upwards.

packages/prepo-shared-contracts/test/SafeOwnableCaller.test.ts
```diff
@@ -0,0 +1,123 @@
...
+      mockOwnedContract.acceptOwnership.reverts()
+
+      await expect(
+        safeOwnableCaller.connect(owner)['acceptOwnership(address)'](mockOwnedContract.address)
+      ).to.be.reverted
+      expect(mockOwnedContract.acceptOwnership).to.have.been.called
+    })
+
+    it("succeeds if owned contract's acceptOwnership doesn't revert", async () => {
```
ramenforbreakfast, line 110:
```suggestion
    it("succeeds if contract's acceptOwnership doesn't revert", async () => {
```
Same here

packages/prepo-shared-contracts/test/SafeOwnableCaller.test.ts
```diff
@@ -0,0 +1,123 @@
...
+      await safeOwnableCaller.connect(owner)['acceptOwnership(address)'](mockOwnedContract.address)
+
+      expect(mockOwnedContract.acceptOwnership).to.not.have.been.reverted
+      expect(mockOwnedContract.acceptOwnership).to.have.been.called
+    })
+
+    it('calls acceptOwnership of owned contract with correct parameters', async () => {
```
ramenforbreakfast, line 117:
```suggestion
    it('calls acceptOwnership of contract with correct parameters', async () => {
```
Same here

packages/prepo-shared-contracts/test/SafeOwnableCaller.test.ts
```diff
@@ -0,0 +1,123 @@
...
+      expect(await safeOwnableCaller.owner()).to.not.eq(user1.address)
+
+      await expect(
+        safeOwnableCaller
+          .connect(user1)
+          ['transferOwnership(address,address)'](mockOwnedContract.address, user1.address)
+      ).revertedWith('Ownable: caller is not the owner')
+    })
+
+    it("reverts if owned contract's transferOwnership reverts", async () => {
```
ramenforbreakfast, line 58: Please remove owned for all the other testcases as well, since it is redundant and every testcase has it.

PR #353, opened 2022-08-29
feat: add view methods from AccessControl

ramenforbreakfast:
These were left out of `SafeAccessControlEnumerableCaller.sol` and need to be added back in.


Nilay27:
* Closing as per the decision to not have view methods in caller contracts.

PR #354, opened 2022-08-29
fix: removing view methods from caller contracts

Nilay27:
* Removing view methods from `SafeOwnableCaller` and `SafeAccessControlEnumerableCaller` contracts.


ramenforbreakfast (approved):
LGTM

PR #355, opened 2022-08-29
fix: audit mitigation M-02, adding purchase price argument

Nilay27:
* Adding `purchasePrices` parameter in `purchase()` to avoid race condition and ensure that tokens are not purchased at higher prices than the user intended (in case of change of price of token to an higher amount).
* Since we are fetching price from `_contractToIdToPrice[_tokenContracts[i]][_ids[i]]`, thus loading it(SLOAD)  three times now throughout the purchase function, hence I cached it to local variable `_price`. Therefore this PR also contains gas optimization suggested in G-04.


apps/smart-contracts/token/test/TokenShop.test.ts
```diff
@@ -255,7 +255,11 @@ describe('TokenShop', () => {
 
   describe('# purchase', () => {
     let fakePurchaseHook: FakeContract<Contract>
-
+    const purchasePrices = itemPrices
+    const erc1155Id1PurchasePrice = erc1155Id1Price
+    const erc1155Id2PurchasePrice = erc1155Id2Price
+    const erc721Id1PurchasePrice = erc721Id1Price
+    const erc721Id2PurchasePrice = erc721Id2Price
```
ramenforbreakfast, line 262: I think we should just remove these and directly reference their previously declared `...IdNPrice` variables.

The only time you change these prices is in the >/=/< purchase price tests, and there you are just subtracting/adding one from the reference prices directly.

It would only make sense to declare `let` purchase price variables ahead of time if we were changing them alot between test cases and not setting them until within the testcase.

Here as `const` values, its not really worth it since we only change them a few times and aren't setting them to different values within the testcase, they are all set to the reference prices at the very beginning, making them redundant.

ramenforbreakfast (approved):
LGTM besides my nit about the redundant purchase price declarations.

PR #356, opened 2022-08-29
fix: audit mitigation M-04, adding _gap in upgradeable contracts

Nilay27:
* Adding `uint256[50] private __gap` in abstract upgradeable smart contracts to ensure no storage shifting down storage in the inheritance chain of the inheriting contracts.


ramenforbreakfast (approved):
LGTM

PR #357, opened 2022-08-30
fix: audit gas optimization G-01 and G-02

Nilay27:
Gas optimization in MiniSales.sol: 
`G-01` : using `_purchasePrice` (memory variable) instead of `_price` (storage variable)
`G-02`: caching `_purchaseHook` (storage variable loaded twice)

-- | -- | --
79829 | 79428 | -401


ramenforbreakfast (approved):
LGTM

PR #358, opened 2022-08-30
fix: audit gas optimization G-03

Nilay27:
Gas optimization in TokenShop.sol: 
`G-03`: caching `_purchaseHook` (storage variable used in Loop!)

-- | -- | --
155748 | 155580 | -168 per iteration

Note:
G-04 was already resolved in #355


ramenforbreakfast (approved):
LGTM, yeah this would have been very costly!

PR #359, opened 2022-08-30
fix: audit gas optimization G-05

Nilay27:
Gas optimization in BlocklistTransferHook.sol: 
`G-03`: caching `_blocklist` (storage variable loaded twice)

-- | -- | --
28115 | 28017 | -98


ramenforbreakfast (approved):
LGTM

PR #360, opened 2022-08-30
fix: audit gas optimization G-06 and G-07

Nilay27:
Gas optimization in PurchaseHook.sol (TokenShop): 
`G-06`: caching `_tokenShop ` (storage variable loaded twice) in `hookERC1155()`
`G-07`: caching `_tokenShop ` (storage variable loaded twice) in `hookERC721()`

-- | -- | --
29587 | 29502 | -85


ramenforbreakfast (approved):
LGTM

PR #361, opened 2022-08-30
fix: audit gas optimizations G-08, G-10 and G12

Nilay27:
Gas optimization in Vesting.sol (TokenShop): 
`G-08`: caching `_token` (storage variable loaded twice) in `claim()`

`G-10`: using `unchecked` for calculation of `_newTotalAllocatedSupply` in `setAllocations()`(we already ensure no overflow through `if(_amount > _prevAllocatedAmount)`.  This is called in Loop!

`G-12`: using `!= 0` instead of `>0` in `claim()`

Gas Before (`setAllocations()`) | Gas After | Change
-- | -- | --
80582 | 80364 | -218 (in each interation)

Gas Before (`claim()`) | Gas After | Change
-- | -- | --
89861 | 89763 | -98


ramenforbreakfast (approved):
LGTM

PR #362, opened 2022-08-30
fix: audit gas optimization G-09

Nilay27:
Gas optimization in `WithdrawERC20.sol`: 
`G-08`: caching `owner()` (storage variable loaded in loop) in `withdrawERC20()`

Gas Before  | Gas After | Change
-- | -- | --
61495 | 61440 | -55 (in each interation)


ramenforbreakfast (approved):
LGTM

PR #363, opened 2022-08-30
fix: audit gas optimizations G-13

Nilay27:
`G-13`: Caching the array length in the stack to save around 3 gas per iteration.


ramenforbreakfast (approved):
LGTM!

PR #365, opened 2022-08-31
fix: audit gas optimisation G-11

Nilay27:
`G-11`: Using `unchecked{ ++i;}` in loop to save around 80 gas per iteration.


ramenforbreakfast (approved):
LGTM

PR #367, opened 2022-09-01
feat: added deployment name constants

ramenforbreakfast:
This PR adds a `DEPLOYMENT_NAMES` constant to the `prepo-constants` package. The purpose of this is to ensure developers, etiher when deploying contracts themselves or referencing past deployments, can easily recall/access deployment names without having to copy and paste strings and manually look them up.

I have also forced all deploy scripts to use `DEPLOYMENT_NAMES` instead of copy and pasted strings.

Below is a short demo of why this is useful in preventing errors during deployment when we need to deploy multiple instances of one contract.

https://user-images.githubusercontent.com/5270250/187851162-dbd55fa5-3547-4525-aa44-7ae7e90ef465.mov


Nilay27 (approved):
LGTM!

PR #368, opened 2022-09-01
feat: inject DEPLOYMENT_NAMES into hardhat console for usage with tasks

ramenforbreakfast:
When creating utilities to perform operations on contracts where multiple instances might exist, the easiest solution I could come up with back during the core testnet was having the developer pass in the contract address directly.

This can be very error prone and incredibly annoying, relying on manually copy and pasting addresses referenced from somewhere else.

When working on the token utilities, I realized we could utilize the constants submitted in the previous PR and the Hardhat console to make referring to specific contracts easier with tasks. This PR will inject `DEPLOYMENT_NAMES` into the `hardhat console` if launched from the `token` directory.

Below is a brief demonstration with a currently in-development simple `modify-account-list` utility that will add or remove members to the `AccountList` under `name`.

https://user-images.githubusercontent.com/5270250/187857150-399e8f9e-0243-4d4c-b08b-8ac0f5dcc7ab.mov


Nilay27 (approved):
LGTM! this will be really helpful

PR #369, opened 2022-09-01
feat: add initial Token harness for setting up contract stack

ramenforbreakfast:
This is a harness for the `Token` project to easily retrieve the entire contract stack when writing scripts or tasks.
Includes multiple improvements from the original core stack harness:
- Contracts are arranged in a hierarchical graph rather than the harness having every contract as a top level property.
- Contracts are retrieved in batches in parallel. I did not retrieve them all at once since if a contract were to be inserted somewhere later, all the array indices below it would have to be shifted. I have settled for batching contract retrieval with each top level item.


Nilay27 (commented):
LGTM

PR #371, opened 2022-09-02
feat: utility to add or remove accounts from an AccountList

ramenforbreakfast:
The first utility to be added. A follow up PR will be made to change the task to use proposals rather than direct transaction execution. Feel free to try it out and provide suggestions!


Nilay27 (approved):
LGTM!

PR #375, opened 2022-09-05
feat: switch over to proposals rather than direct execution

ramenforbreakfast:


Nilay27 (approved):
LGTM, tested the code, everything works fine!

PR #379, opened 2022-09-08
feat: redeployed PPO and MiniSales because gap adds storage slots in inheriting contracts

ramenforbreakfast:


apps/smart-contracts/token/deployments/arbitrumTestnet/MiniSales_A.json
```diff
@@ -1,5 +1,5 @@
 {
-  "address": "0xAd51B8C98bc07a5D4caeF8218537869D165078B3",
```
Nilay27, line 2: Mini Sales does not inherit PPO so ideally adding storage slots in PPO should not affect MiniSales right? 

And if it does then Vesting also uses PPO and should be redeployed?

Nilay27 (approved):
LGTM besides on question.

apps/smart-contracts/token/deployments/arbitrumTestnet/MiniSales_A.json
ramenforbreakfast: Problem was that MiniSales does not allow changing of token whereas Vesting does.

PR #380, opened 2022-09-08
feat: add support for separate prod and test OZ Defender api keys

ramenforbreakfast:
To allow us to manage our contracts via different OZ Defender portals, we can no longer rely on the `@openzeppelin/hardhat-defender` package. This is because it forces the use of a singular API credential through the `hardhat.config.ts`.

The more flexible and stripped `defender-admin-client` allows us to dynamically specify what credential we are using, so at the expense of some additional operations that `hardhat-defender` previously simplified, we can now switch between a PROD and TEST api key depending on the detected network.

Developers will now need to have two separate environment variables `DEFENDER_PROD_API_KEY`, `DEFENDER_PROD_API_SECRET` and `DEFENDER_TEST_API_KEY`, `DEFENDER_TEST_API_SECRET`.


Nilay27 (approved):
LGTM!

PR #388, opened 2022-09-12
chore: provided local compiler

georgesmithishere:
https://www.notion.so/prepo/FE-Project-Management-d3878c29e9534234a50f4f8838525bee?p=6a6b41087e0e417d99678dcffb7aeb6e&pm=s

let's check how it works (we have to update environment variable LOCAL_SOLC for pipeline)
I found similar issue here:
https://github.com/NomicFoundation/hardhat/issues/1280
added local solcjs binary as shown here:
https://github.com/fvictorio/hardhat-examples/blob/master/custom-solc/hardhat.config.js
list of all binaries
https://etherscan.io/solcversions


ramenforbreakfast:
For testing purposes, I have rebased this from a fresh `main` , and rebased PR #391 and #392 off it to see if the checks run properly. The problem is non-deterministic so hard to easily tell whether this has actually fixed anything,

ramenforbreakfast (approved):
I don't see any issues, good solution to only use the local compiler for the GitHub actions. However, we still need these added for the token project as well as core projects.

chrisling-dev (approved):

chrisling-dev (approved):
Yes I think this is a good solution, good finding and thanks for the relevant links!

PR #391, opened 2022-09-13
feat: adding contract for mini sales flag

Nilay27:
* Adding contract for FE query to check whether a sale has started or not.
* Added all files at once as it is a simple safe-ownable contract with just a boolean flag.


ramenforbreakfast (approved):
LGTM, I will merge this in.

PR #408, opened 2022-09-19
fix: address multiple task bugs identified during usage on arbitrum

ramenforbreakfast:
This PR forces tasks to use new defender name constants rather than assuming the infura endpoint was consistent with OZ network naming. They were the same for Arbitrum Rinkeby, and it turns out they are not the same for Arbitrum mainnet.

In addition, add await to proposal creation to ensure we wait for the proposal to be submited in case there is an error returned by the API.


Nilay27 (approved):
LGTM!

PR #421, opened 2022-09-29
chore: change function name to differentiate between partial and full withdrawal

ramenforbreakfast:
The full withdrawal function will be submitted in a follow-up PR


Nilay27 (approved):
LGTM! The names of the tests were not changed; I have changed them in the follow-up PR.

XavierEkkel (commented):
Name change unnecessary - can just keep it as withdrawERC20 for both (they won't clash due to different arguments).

PR #422, opened 2022-09-29
fix: fix path to compiler

ramenforbreakfast:
Local compiler was pointing to the wrong path for `prepo-shared-contracts`


ramenforbreakfast:
Hold off on approving this, still seems to be failing, also in PR #421 which is based off this, core tests are failing instead of `prepo-shared-contracts`.

ramenforbreakfast:
Ok looks good now, the issue is that in the GitHub Actions YAML, we were setting the local env `LOCAL_SOLC` to `TRUE`, but when actually evaluated, would return a boolean `true`. Thus instead of using `LOCAL_SOLC===TRUE`, we are just performing a boolean check.

georgesmithishere:
> Ok looks good now, the issue is that in the GitHub Actions YAML, we were setting the local env `LOCAL_SOLC` to `TRUE`, but when actually evaluated, would return a boolean `true`. Thus instead of using `LOCAL_SOLC===TRUE`, we are just performing a boolean check.

I was testing it locally - it resolves as a string, thank you for fixing it

chrisling-dev (approved):

PR #435, opened 2022-10-11
Feat/portfolio filter

georgesmithishere:
https://www.notion.so/prepo/FE-Project-Management-d3878c29e9534234a50f4f8838525bee?p=0b2cd82c9a6c489fa28e98815c870a1a&pm=s


apps/frontend/core-dapp/src/stores/graphs/CoreGraphStore.ts
```diff
@@ -7,18 +7,28 @@ import { RootStore as CoreRootStore } from '../../../generated/mst-gql/core-dapp
 import { PositionCostBasis } from '../../types/user.types'
 import { HistoricalEvents } from '../../features/history/history.types'
 
+export type HistoryEventFilter = {
```
chrisling-dev, line 10: for consistency purpose could we rename this to `HistoricalEventsFilter`
```suggestion
export type HistoricalEventsFilter = {
```

apps/frontend/core-dapp/src/stores/graphs/CoreGraphStore.ts
```diff
@@ -7,18 +7,28 @@ import { RootStore as CoreRootStore } from '../../../generated/mst-gql/core-dapp
...
-    makeObservable(this, { positionsCostBasis: observable })
+    makeObservable(this, { positionsCostBasis: observable, historicalEvents: observable })
   }
 
   positionsCostBasis = (address: string): PositionCostBasis | undefined =>
     this.query<PositionCostBasis>(userPositionsQueryString, { address: address.toLowerCase() })
       ?.data
 
-  historicalEvents = (address: string): HistoricalEvents | undefined =>
+  historicalEvents = (filter: HistoryEventFilter): HistoricalEvents | undefined =>
```
chrisling-dev, line 30:
```suggestion
  historicalEvents = (filter: HistoricalEventsFilter): HistoricalEvents | undefined =>
```

chrisling-dev (changes requested):
Overall the behaviour is okay except for 1 thing:

- The filter shouldn't be applied until they click "Confirm", currently if I select a tx type for example, without clicking "Confirm" it already applies in the background which feels very weird. 
 
The solution should be that, on the filter modal (or wherever the UI is), we keep track of a local state, and these local states only get applied to PortfolioStore when confirm is clicked. And if I close or cancel the filter, it should reset to the states on PortfolioStore

georgesmithishere:
> Overall the behaviour is okay except for 1 thing:
> 
> * The filter shouldn't be applied until they click "Confirm", currently if I select a tx type for example, without clicking "Confirm" it already applies in the background which feels very weird.
> 
> The solution should be that, on the filter modal (or wherever the UI is), we keep track of a local state, and these local states only get applied to PortfolioStore when confirm is clicked. And if I close or cancel the filter, it should reset to the states on PortfolioStore

I totally agree with you, and I showed that behavior at sync, created separate ticket for it

chrisling-dev (approved):
LGTM

PR #437, opened 2022-10-12
feat: adding functionality to support splitting of whitelisting in groups

Nilay27:
* Adding functionality for splitting up lists into smaller groups and creating multiple proposals for the same.
* The modification to the existing modify-account-list task also allows us to specify a modification to the title of a proposal.

Example to run the command: 
`await run('modify-account-list', { name: DEPLOYMENT_NAMES.ppo.restrictedTransferHook.sourceAllowlist.name, groupSize: '7', add: './example-accounts-to-include.txt' , remove: './example-accounts-to-remove.txt'})`

The additional optional parameter `groupSize: '7'` will create multiple groups, each containing 7 addresses; it automatically takes care of the last group size in case of uneven splitting. 

For example, if total addresses = 10, group size = 7, then group 1 will have 7 addresses, and group 2 will have 3 addresses.


ramenforbreakfast (approved):
LGTM

PR #438, opened 2022-10-12
feat: confirm state

georgesmithishere:
https://www.notion.so/prepo/FE-Project-Management-d3878c29e9534234a50f4f8838525bee?p=0a550c2eaa064e02a3933bff1b638c1b&pm=s


chrisling-dev (changes requested):
Almost there, could we make it so that when we close the modal without confirming (e.g. click close icon or click outside of modal), it calls the `filterStore.resetFilters()`, currently it's a bit weird because if let's say I deselect a type, then close without confirm, nothing changes but when I open the modal the 'unapplied' filter is still there

Pretty sure you just have to call resetFilters in `FilterModal`'s `onClose`

chrisling-dev (approved):
LGTM!

PR #451, opened 2022-10-19
feat: initial PreUSD interface with no natspec

ramenforbreakfast:
Initial PreUSD interface with no commenting.

This interface will still be called `ICollateral`, because the methods on this contract will most likely be shared by all collateral types if we want to support more in the future. Also because no functions on this interface are PreUSD specific, and because we are using the generic "collateral" naming convention throughout the stack, it makes sense to not call this `IPreUSD`, but `ICollateral` instead.

Also in this PR, the `collateral` state in every hook is now wrapped with this `ICollateral` interface, rather than just being an address. This is because hooks such as `ManagerWithdrawHook` will need to interact with `PreUSD` beyond just checking if the right address is calling the hook. I figure since they all share this variable, we just have any contract inheriting `IHook`, wrap it with `ICollateral`.


ramenforbreakfast:
Closing to reduce PR train size and will re-open as a separate PR.

PR #452, opened 2022-10-19
feat: initial deposit hook and withdraw hook interfaces

ramenforbreakfast:
I decided to split out `CollateralDepositRecord` functions into a separate interface to reduce replication.

Along with the decision to wrap the `collateral` variable all hooks have with the `ICollateral` interface, **this now means `ManagerWithdrawHook.sol` will not need its own interface** since `ManagerWithdrawHook.sol` just needs to implement `IDepositRecordHook.sol`.


ramenforbreakfast:
Closing to reduce PR train size

PR #455, opened 2022-10-22
feat: add ICollateral skeleton

ramenforbreakfast:
Initial PreUSD interface with no commenting.

This interface will still be called `ICollateral`, because the methods on this contract will most likely be shared by all collateral types if we want to support more in the future. Also because no functions on this interface are PreUSD specific, and because we are using the generic "collateral" naming convention throughout the stack, it makes sense to not call this `IPreUSD`, but `ICollateral` instead.

Also in this PR, the collateral state in every hook is now wrapped with this `ICollateral` interface, rather than just being an address. This is because hooks such as `ManagerWithdrawHook` will need to interact with `PreUSD` beyond just checking if the right address is calling the hook. I figure since they all share this variable, we just have any contract inheriting `IHook` wrap it with `ICollateral`.


apps/smart-contracts/core/contracts/DepositHook.sol
```diff
@@ -1,21 +1,20 @@
-// SPDX-License-Identifier: UNLICENSED
+// SPDX-License-Identifier: AGPL-3.0
 import "./interfaces/IHook.sol";
-import "./interfaces/IAccountAccessController.sol";
 import "./interfaces/ICollateralDepositRecord.sol";
 import "@openzeppelin/contracts/access/Ownable.sol";
 
 pragma solidity =0.8.7;
```
XavierEkkel, line 6: nit: new styling (used throughout token contracts) is to place this line should go under license, with an empty line before any of the imports

same for other core contracts

apps/smart-contracts/core/contracts/interfaces/ICollateral.sol
```diff
@@ -0,0 +1,45 @@
+// SPDX-License-Identifier: AGPL-3.0
+
```
XavierEkkel, line 2: nit: no empty line between license and solidity version (or at least, that's how it was done in the token contracts)
```suggestion
```

same for other files

apps/smart-contracts/core/contracts/interfaces/ICollateral.sol
```diff
@@ -0,0 +1,45 @@
...
+pragma solidity =0.8.7;
+
+import "./IHook.sol";
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts-upgradeable/token/ERC20/IERC20Upgradeable.sol";
+
+interface ICollateral is IERC20Upgradeable {
+  // SETTERS
```
XavierEkkel, line 10: nit: this is self-evident + not accurate for the deposit/withdraw functions
```suggestion
```

apps/smart-contracts/core/contracts/interfaces/ICollateral.sol
```diff
@@ -0,0 +1,45 @@
...
+
+  function setDepositHook(IHook newHook) external;
+
+  function setWithdrawHook(IHook newHook) external;
+
+  function setManagerWithdrawHook(IHook newHook) external;
+
+  function setMinReservePercentage(uint256 newMinReservePercentage) external;
+
+  // GETTERS
```
XavierEkkel, line 31: nit: this is self-evident
```suggestion
```

apps/smart-contracts/core/contracts/interfaces/ICollateral.sol
```diff
@@ -0,0 +1,45 @@
...
+
+interface ICollateral is IERC20Upgradeable {
+  // SETTERS
+  function deposit(uint256 amount) external;
+
+  function withdraw(uint256 amount) external;
+
+  function managerWithdraw(uint256 amount) external;
+
+  function setBaseToken(IERC20 newBaseToken) external;
```
XavierEkkel, line 17: remove as discussed

XavierEkkel (approved):
LGTM after comments

apps/smart-contracts/core/contracts/interfaces/ICollateral.sol
```diff
@@ -0,0 +1,45 @@
...
+
+  function getWithdrawFee() external view returns (uint256);
+
+  function getDepositHook() external view returns (IHook);
+
+  function getWithdrawHook() external view returns (IHook);
+
+  function getManagerWithdrawHook() external view returns (IHook);
+
+  function getMinReservePercentage() external view returns (uint256);
```
XavierEkkel, line 44: since `getMinReservePercentage` is only ever called by `ManagerWithdrawHook.sol`, consider moving this to be in the hook instead

XavierEkkel (approved):

PR #459, opened 2022-10-24
feat: added WithdrawHook getters, setters, variables, and role constants

ramenforbreakfast:
This PR contains the initial state variables as defined in the technical document and their appropriate roles.

Getters and empty setters are provided so that the `WithdrawHook` can implement `IWithdrawHook` and have access to events later on.

In addition, a slight fix to the `setDepositRecord` function has been made to ensure it follows the consistent pattern of immediately wrapping the parameter with the interface.


XavierEkkel (approved):
LGTM

PR #460, opened 2022-10-24
feat: add setters to WithdrawHook and prepare tests for hook testing

ramenforbreakfast:
This PR contains multiple changes, but mainly focuses on adding setters for all the variables added in the previous PR.

In addition to setters, a minor correction to the `_initialAmount` and `_finalAmount` parameters was made to the `hook`.

The test suite has also had two somewhat-minor modifications to prepare it for hook testing (since the WithdrawHook has more involved logic than the others)

- The `TEST_AMOUNT_ONE` and `TEST_AMOUNT_TWO` made no sense because all we cared about was that the correct values were being passed to the `CollateralDepositRecord`. Since we will need to do actual testing of withdraw limits being met, I have renamed these to `TEST_AMOUNT_BEFORE_FEE` and `TEST_AMOUNT_AFTER_FEE` and changed their values to something more reasonable.

- There is one test case for checking if `CollateralDepositRecord` was called correctly. This test was written before we used `smock` and I have simplified it to use `smock`.


ramenforbreakfast:

apps/smart-contracts/core/contracts/WithdrawHook.sol
```diff
@@ -42,10 +42,10 @@ contract WithdrawHook is IWithdrawHook, SafeAccessControlEnumerable {
 
   function hook(
     address _sender,
-    uint256 _initialAmount,
-    uint256 _finalAmount
+    uint256 _amountBeforeFee,
+    uint256 _amountAfterFee
   ) external override onlyCollateral {
-    _depositRecord.recordWithdrawal(_sender, _finalAmount);
+    _depositRecord.recordWithdrawal(_sender, _amountAfterFee);
```
XavierEkkel, line 48: I think `_amountBeforeFee` should be used here for a more accurate global net deposit amount.

Simple example: if someone deposits $10 and then withdraws it immediately (but there's a 10% fee), we'd still expect global net deposits to be $0 again rather than $1.

Remember to update tests for this.

apps/smart-contracts/core/test/WithdrawHook.test.ts
```diff
@@ -195,4 +231,279 @@ describe('=> WithdrawHook', () => {
...
+      await withdrawHook.connect(deployer).setWithdrawalsAllowed(true)
+
+      expect(await withdrawHook.withdrawalsAllowed()).to.eq(true)
+
+      await withdrawHook.connect(deployer).setWithdrawalsAllowed(true)
+
+      expect(await withdrawHook.withdrawalsAllowed()).to.eq(true)
+    })
+
+    it('should emit a WithdrawalsAllowedChange event', async () => {
```
XavierEkkel, line 275:
```suggestion
    it('emits WithdrawalsAllowedChange', async () => {
```
same for other event-emitting tests

XavierEkkel (approved):
LGTM after comments

PR #461, opened 2022-10-25
feat: modify DepositHook to conform to IDepositHook

ramenforbreakfast:
Similar to the WithdrawHook, I have fixed the `setDepositRecord` function to wrap the parameter with the `ICollateralDepositRecord` interface, and fixed the `initialAmount` and `finalAmount` parameters in `hook`.

I did not go through the effort of changing the `TEST_AMOUNT_ONE` and `TEST_AMOUNT_TWO` in this test suite because there is no significant logic with amounts for `DepositHook`.


ramenforbreakfast:

apps/smart-contracts/core/contracts/DepositHook.sol
```diff
@@ -25,10 +28,11 @@ contract DepositHook is IHook, SafeAccessControlEnumerable {
...
   function hook(
     address _sender,
-    uint256 _initialAmount,
-    uint256 _finalAmount
+    uint256 _amountBeforeFee,
+    uint256 _amountAfterFee
   ) external override onlyCollateral {
-    _depositRecord.recordDeposit(_sender, _finalAmount);
+    require(_depositsAllowed, "deposits not allowed");
+    _depositRecord.recordDeposit(_sender, _amountAfterFee);
```
XavierEkkel, line 35: Agree on using `_amountAfterFee` here, as only the after-fee amount contributes to TVL.

XavierEkkel (approved):
LGTM after nit

PR #462, opened 2022-10-25
feat: add global withdraw limit logic

ramenforbreakfast:
We did not take the actual period tracking logic into account for the document. For the purposes of tracking logic, we need to store when the last time a period was reset, via `_lastGlobalPeriodReset` and `_lastUserPeriodReset`.

To make testing easier and cleaner, I have modified the limit so only one additional user needs to deposit the test amount to reach the global cap.

Originally I was going to submit both the global and user limit logic together, but the PR became too big and I have now split them up.


ramenforbreakfast:

apps/smart-contracts/core/contracts/WithdrawHook.sol
```diff
@@ -45,6 +46,21 @@ contract WithdrawHook is IWithdrawHook, SafeAccessControlEnumerable {
...
+    ) {
+      _lastGlobalPeriodReset = block.timestamp;
+      _globalAmountWithdrawnThisPeriod = 0;
+    }
+    require(
+      _globalAmountWithdrawnThisPeriod + _amountAfterFee <=
+        _globalWithdrawLimitPerPeriod,
+      "global withdraw limit exceeded"
+    );
+    _globalAmountWithdrawnThisPeriod += _amountAfterFee;
```
XavierEkkel, line 63: it's inefficient to have a case that will only ever be used once
we can also reduce the number of operations in the reset case
and don't forget to change to use `_amountBeforeFee` instead
```suggestion
    if (
      _lastGlobalPeriodReset + _globalPeriodLength < block.timestamp
    ) {
      _lastGlobalPeriodReset = block.timestamp;
      _globalAmountWithdrawnThisPeriod = _amountBeforeFee;
    } else {
      require(
        _globalAmountWithdrawnThisPeriod + _amountBeforeFee <=
          _globalWithdrawLimitPerPeriod,
        "global withdraw limit exceeded"
      );
      _globalAmountWithdrawnThisPeriod += _amountBeforeFee;
    }
```

apps/smart-contracts/core/test/WithdrawHook.test.ts
```diff
@@ -134,6 +144,157 @@ describe('=> WithdrawHook', () => {
...
+        await withdrawHook
+          .connect(vault)
+          .hook(user.address, TEST_AMOUNT_BEFORE_FEE, TEST_AMOUNT_AFTER_FEE)
+
+        expect(await withdrawHook.getLastGlobalPeriodReset()).to.eq(
+          await getLastTimestamp(ethers.provider)
+        )
+      })
+
+      it('sets last global reset to current time if global period has passed (>)', async () => {
```
XavierEkkel, line 161: nit:
```suggestion
      it('sets last global reset to current time if global period passed', async () => {
```
similarly for others

apps/smart-contracts/core/test/WithdrawHook.test.ts
```diff
@@ -134,6 +144,157 @@ describe('=> WithdrawHook', () => {
...
+        await withdrawHook
+          .connect(vault)
+          .hook(user.address, differentAmountToWithdraw, differentAmountToWithdraw)
+
+        expect(await withdrawHook.getGlobalAmountWithdrawnThisPeriod()).to.eq(
+          differentAmountToWithdraw
+        )
+      })
+
+      it("doesn't update last global reset if global period was just reached (==)", async () => {
```
XavierEkkel, line 203: nit:
```suggestion
      it("doesn't update last global reset if global period exactly reached", async () => {
```

similarly for others

apps/smart-contracts/core/test/WithdrawHook.test.ts
```diff
@@ -134,6 +144,157 @@ describe('=> WithdrawHook', () => {
...
+          ethers.provider,
+          previousResetTimestamp + TEST_GLOBAL_PERIOD_LENGTH + 1
+        )
+
+        await withdrawHook
+          .connect(vault)
+          .hook(user.address, TEST_AMOUNT_BEFORE_FEE, TEST_AMOUNT_AFTER_FEE)
+
+        const currentResetTimestamp = await getLastTimestamp(ethers.provider)
+        expect(await withdrawHook.getLastGlobalPeriodReset()).to.eq(currentResetTimestamp)
```
XavierEkkel, line 177: also assert that it's greater than the previous timestamp

apps/smart-contracts/core/test/WithdrawHook.test.ts
```diff
@@ -134,6 +144,157 @@ describe('=> WithdrawHook', () => {
...
+        await withdrawHook
+          .connect(vault)
+          .hook(user.address, TEST_AMOUNT_BEFORE_FEE, TEST_AMOUNT_AFTER_FEE)
+
+        expect(await withdrawHook.getGlobalAmountWithdrawnThisPeriod()).to.eq(
+          previousGlobalAmountWithdrawn.add(TEST_AMOUNT_AFTER_FEE)
+        )
+      })
+
+      it('reverts if global withdraw limit exceeded during a period', async () => {
```
XavierEkkel, line 275: nit: 
```suggestion
      it('reverts if global withdraw limit exceeded for period', async () => {
```

XavierEkkel, line 275: what about a test for when it reaches the limit exactly?

XavierEkkel (approved):
LGTM after comments

PR #463, opened 2022-10-25
feat: add user withdraw limit logic

ramenforbreakfast:
This continues on the previous PR, but with user withdraw limits.


ramenforbreakfast:

XavierEkkel (commented):
please apply the same comments from the global limit PR to this one, and then re-request review!

apps/smart-contracts/core/contracts/WithdrawHook.sol
```diff
@@ -58,6 +59,17 @@ contract WithdrawHook is IWithdrawHook, SafeAccessControlEnumerable {
       );
       _globalAmountWithdrawnThisPeriod += _amountBeforeFee;
     }
+    if (_lastUserPeriodReset + _userPeriodLength < block.timestamp) {
+      _lastUserPeriodReset = block.timestamp;
```
XavierEkkel, line 63: Please add as a dev natspec comment in this implementation contract on this function why we reset to block.timestamp rather than 'last period + period length'
I think the reasoning would be that otherwise the limit logic wouldn't work as intended if multiple periods were missed (until the last period reset timestamp catches up to the current period).

apps/smart-contracts/core/test/WithdrawHook.test.ts
```diff
@@ -316,6 +316,188 @@ describe('=> WithdrawHook', () => {
...
+        )
+        const previousResetTimestamp = await getLastTimestamp(ethers.provider)
+        await setNextTimestamp(
+          ethers.provider,
+          previousResetTimestamp + TEST_USER_PERIOD_LENGTH + 1
+        )
+
+        await withdrawHook
+          .connect(vault)
+          .hook(user.address, differentAmountToWithdraw, differentAmountToWithdraw)
```
XavierEkkel, line 368: in a separate PR, could you add a test / tests that the 3rd argument (amountAfterFee) is ignored?
can just leave a TODO at the top of this file

apps/smart-contracts/core/contracts/interfaces/IWithdrawHook.sol
```diff
@@ -39,6 +39,8 @@ interface IWithdrawHook is IDepositRecordHook {
 
   function getLastGlobalPeriodReset() external view returns (uint256);
 
+  function getLastUserPeriodReset() external view returns (uint256);
```
XavierEkkel, line 42: add test that the initial state of this is 0

apps/smart-contracts/core/test/WithdrawHook.test.ts
```diff
@@ -316,6 +316,188 @@ describe('=> WithdrawHook', () => {
         )
       })
     })
+
+    describe('user withdraw limit testing', () => {
```
XavierEkkel, line 320: add test that it doesn't revert if amountBeforeFee = 0
similarly for global section above

XavierEkkel (approved):
LGTM after comments

apps/smart-contracts/core/contracts/WithdrawHook.sol
```diff
@@ -58,6 +59,17 @@ contract WithdrawHook is IWithdrawHook, SafeAccessControlEnumerable {
       );
       _globalAmountWithdrawnThisPeriod += _amountBeforeFee;
     }
+    if (_lastUserPeriodReset + _userPeriodLength < block.timestamp) {
```
XavierEkkel, line 62: shouldn't this be `>`? same for above with global reset

apps/smart-contracts/core/test/WithdrawHook.test.ts
ramenforbreakfast: This is already tested by the fact that except for this particular section you pointed out where we are just withdrawing again, I on purposely use different values for `amountBeforeFee` and `amountAfterFee`. In addition there is a testcase that verifies `recordWithdrawal` ignores `amountAfterFee`.

What I can do here is also make this particular section use a different `amountAfterFee`

PR #466, opened 2022-10-26
feat: modify WithdrawHook to use a smock contract

ramenforbreakfast:
The old WithdrawHook tests uses an actual CollateralDepositRecord contract and attempts to verify written values, which is problematic given that the component is being changed.

To ensure we can keep submitting things in parallel, I have separated this out into its own PR for future PRs related to the `CollateralDepositRecord` or `WithdrawHook` to base themselves on.


XavierEkkel (approved):
LGTM

PR #469, opened 2022-10-30
feat: adding FixedPriceOracle

ramenforbreakfast:
In addition for this PR, I had to remove a few lines in some configuration files in the `prepo-shared-contracts` folder that we do not need.


packages/prepo-shared-contracts/contracts/FixedPriceOracle.sol
```diff
@@ -0,0 +1,19 @@
+// SPDX-License-Identifier: AGPL-3.0
+pragma solidity =0.8.7;
+
+import "./SafeOwnable.sol";
+
+contract FixedPriceOracle is SafeOwnable {
```
XavierEkkel, line 6: lets use an interface for this like we do with other contracts
(assuming we might use this contract again in future)

packages/prepo-shared-contracts/hardhat.config.ts
```diff
@@ -51,23 +51,6 @@ const config: HardhatUserConfig = {
           },
         },
       },
-      {
-        version: '0.8.6',
```
XavierEkkel, line 55: 👍

XavierEkkel (approved):
LGTM after using an interface

PR #470, opened 2022-10-30
feat: add record for FixedPriceOracle deployment

ramenforbreakfast:
This deployment script is essentially 99% the same as the rest of our scripts, except with the name changed.

Other changes in this PR involve removing minor technical debt in our `prepo-shared-contracts` folder. We have never deployed a contract from this folder, so I had to fix some configuration files and packages to allow us to do so.

Previously it was using its own `hardhat-etherscan`, a much older version, which was preventing me from performing verification. I have removed this reference and it now uses the shared etherscan package.


XavierEkkel (approved):
LGTM

PR #471, opened 2022-10-31
feat: ppo price feed api

chrisling-dev:
https://www.notion.so/prepo/FE-Project-Management-d3878c29e9534234a50f4f8838525bee?p=0886682bdf7549759880a1b3f3c4aabf&pm=s

### DIFF
--- a/apps/frontend/core-dapp/src/pages/api/ppo-price.ts
+++ b/apps/frontend/core-dapp/src/pages/api/ppo-price.ts
@@ -1,0 +1,32 @@
+import { ethers } from 'ethers'
+import { formatUnits } from 'ethers/lib/utils'
+import { NextApiRequest, NextApiResponse } from 'next'
+import { NETWORKS } from 'prepo-constants'
+
+const ORACLE_ABI = ['function getFixedPrice() view returns (uint256)']
+const ORACLE_ADDRESS = '0x503aef858e26C136C29e40483EA0297160d2BDC0'
+const PRICE_DECIMALS = 6
+
+const handler = async (_: NextApiRequest, res: NextApiResponse): Promise<void> => {
+  try {
+    const { rpcUrls, chainId } = NETWORKS.arbitrumOne
+
+    const provider = new ethers.providers.JsonRpcProvider(rpcUrls[0], chainId)
+
+    const fixedPriceContract = new ethers.Contract(ORACLE_ADDRESS, ORACLE_ABI, provider)
+
+    const priceBN = await fixedPriceContract.getFixedPrice()
+
+    if (priceBN) {
+      res.status(200).send(formatUnits(priceBN, PRICE_DECIMALS))
+      res.end()
+      return
+    }
+
+    throw new Error()
+  } catch (e) {
+    res.status(500).send('Internal server error')
+  }
+}
+
+export default handler


apps/frontend/core-dapp/src/pages/api/ppo-price.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    if (priceBN) {
+      res.status(200).send(formatUnits(priceBN, PRICE_DECIMALS))
+      res.end()
+      return
+    }
+
+    throw new Error()
+  } catch (e) {
+    res.status(500).send('Internal server error')
```
liqwiz, line 28:
```suggestion
  } catch (e) {
    console.error(e)
    res.status(500).send('Internal server error')
```

apps/frontend/core-dapp/src/pages/api/ppo-price.ts
```diff
@@ -0,0 +1,32 @@
...
+const ORACLE_ADDRESS = '0x503aef858e26C136C29e40483EA0297160d2BDC0'
+const PRICE_DECIMALS = 6
+
+const handler = async (_: NextApiRequest, res: NextApiResponse): Promise<void> => {
+  try {
+    const { rpcUrls, chainId } = NETWORKS.arbitrumOne
+
+    const provider = new ethers.providers.JsonRpcProvider(rpcUrls[0], chainId)
+
+    const fixedPriceContract = new ethers.Contract(ORACLE_ADDRESS, ORACLE_ABI, provider)
```
liqwiz, line 16: not sure how memory works in the Vercel functions.. but perhaps these declarations can be moved outside of the `handler` function and into the root scope like `PRICE_DECIMALS`, since they should not change each call?

apps/frontend/core-dapp/src/pages/api/ppo-price.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    const priceBN = await fixedPriceContract.getFixedPrice()
+
+    if (priceBN) {
+      res.status(200).send(formatUnits(priceBN, PRICE_DECIMALS))
+      res.end()
+      return
+    }
+
+    throw new Error()
```
liqwiz, line 26:
```suggestion
    throw new Error('getFixedPrice returned falsy value')
```

apps/frontend/core-dapp/src/pages/api/ppo-price.ts
```diff
@@ -0,0 +1,32 @@
...
+  try {
+    const { rpcUrls, chainId } = NETWORKS.arbitrumOne
+
+    const provider = new ethers.providers.JsonRpcProvider(rpcUrls[0], chainId)
+
+    const fixedPriceContract = new ethers.Contract(ORACLE_ADDRESS, ORACLE_ABI, provider)
+
+    const priceBN = await fixedPriceContract.getFixedPrice()
+
+    if (priceBN) {
```
liqwiz, line 20: can price ever be zero?

chrisling-dev: yes but BN in zero wouldn't be falsy (also don't think it will ever be zero after we set the price on contract)

ramenforbreakfast (approved):
LGTM, with one question.

apps/frontend/core-dapp/src/pages/api/ppo-price.ts
```diff
@@ -0,0 +1,32 @@
+import { ethers } from 'ethers'
+import { formatUnits } from 'ethers/lib/utils'
+import { NextApiRequest, NextApiResponse } from 'next'
+import { NETWORKS } from 'prepo-constants'
+
+const ORACLE_ABI = ['function getFixedPrice() view returns (uint256)']
```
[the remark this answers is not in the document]
chrisling-dev: yeap! An example app that uses this format is snapshot-strategies (https://github.com/snapshot-labs/snapshot-strategies/blob/master/src/strategies/mstable/index.ts#L7), it's also documented in ethers.js that we can use this format

So if we look into the type of `Contract`, it's second argument accepts ContractInterface, which could be a string, an array/json, or `Interface` type
Screenshot 2022-11-03 at 8 42 44 AM

the `Interface` type according to ethers.js could be https://docs.ethers.io/v5/api/utils/abi/interface/:
```
/ This interface is used for the below examples

const iface = new Interface([
  // Constructor
  "constructor(string symbol, string name)",

  // State mutating method
  "function transferFrom(address from, address to, uint amount)",

  // State mutating method, which is payable
  "function mint(uint amount) payable",

  // Constant method (i.e. "view" or "pure")
  "function balanceOf(address owner) view returns (uint)",

  // An Event
  "event Transfer(address indexed from, address indexed to, uint256 amount)",

  // A Custom Solidity Error
  "error AccountLocked(address owner, uint256 balance)",

  // Examples with structured types
  "function addUser(tuple(string name, address addr) user) returns (uint id)",
  "function addUsers(tuple(string name, address addr)[] user) returns (uint[] id)",
  "function getUser(uint id) view returns (tuple(string name, address addr) user)"
]);
```

chrisling-dev:
Closing this PR now as it's no longer relevant, the logic is directly implemented in bond-library https://github.com/prepo-io/bond-library/pull/1

PR #476, opened 2022-11-02
feat: trade as default route with query parameters

chrisling-dev:
https://www.notion.so/prepo/FE-Project-Management-d3878c29e9534234a50f4f8838525bee?p=f22c241a1a564004a774ed015eaf31a5&pm=s


chrisling-dev (commented):
Self reviewed

apps/frontend/core-dapp/src/pages/markets.tsx
```diff
@@ -1,16 +1,15 @@
...
+const Markets: NextPage = () => null
+
+// eslint-disable-next-line require-await
+export const getServerSideProps: GetServerSideProps = async () => ({
+  redirect: {
+    permanent: true, // https://nextjs.org/learn/seo/crawling-and-indexing/status-codes
+    destination: Routes.Trade,
+  },
+  props: {},
+})
```
chrisling-dev, line 13: I'm keeping the `markets` page as an empty page that will redirect to `/trade`

This is because most browsers will cache Redirection so those who have visited `app.prepo.io` and redirected to `/markets` before might still get redirected to `/markets` even if we delete it, and they will see a not found message

So it's safer to do another redirection to /trade incase they get redirected to /markets

chrisling-dev (commented):
Self reviewed

apps/frontend/core-dapp/src/lib/markets.ts
```diff
@@ -41,4 +41,7 @@ const prefaketoken: Market = {
 }
 
 export const markets = [prefakestock, prefaketoken]
-export const marketsMap: { [key in SupportedMarketID]: Market } = { prefakestock, prefaketoken }
+export const marketsMap: { [key in SupportedMarketID]: Market } = {
+  fakestock: prefakestock,
+  faketoken: prefaketoken,
```
XavierEkkel, line 46:
```suggestion
  fakestock: fakestock,
  faketoken: faketoken,
```

XavierEkkel (approved):
preview LGTM

will let @liqwiz review the FE code itself

apps/frontend/core-dapp/src/pages/markets.tsx
```diff
@@ -1,16 +1,15 @@
...
+import { GetServerSideProps, NextPage } from 'next'
+import { Routes } from '../lib/routes'
 
-const Markets: NextPage = () => (
-  <>
-    <SEO
-      title="Markets | prePO"
-      description="The decentralized Pre-IPO & Pre-Token trading platform"
-      ogImageUrl="/prepo-og-image.png"
-    />
```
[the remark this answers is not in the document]
chrisling-dev: For markets page no, because it's a deleted page, this file only exists to redirect to trade page because some browser might have cached the redirection to /markets.

So in case browser has cached the redirection, they will go `app.prepo.io -> app.prepo.io/markets -> app.prepo.io/trade`. Nothing we can do here to prevent this extra redirect because the caching is done on user's browser.

But in most cases they will just go `app.prepo.io -> app.prepo.io/trade`

PR #480, opened 2022-11-04
feat: added fee setters

ramenforbreakfast:


apps/smart-contracts/core/test/Collateral.test.ts
```diff
@@ -19,13 +19,31 @@ describe('=> Collateral', () => {
...
+      collateral,
+      deployer,
+      deployer,
+      await collateral.SET_DEPOSIT_FEE_ROLE()
+    )
+    await grantAndAcceptRole(
+      collateral,
+      deployer,
+      deployer,
+      await collateral.SET_WITHDRAW_FEE_ROLE()
```
XavierEkkel, line 43: deployer has the base admin role initially, but that base admin role doesn't mean they can call any function, correct?

XavierEkkel (approved):
LGTM after question

apps/smart-contracts/core/test/Collateral.test.ts
ramenforbreakfast: Technically correct, the base admin role has the ability to initially _assign_ anybody to _any_ role. But does not actually hold any roles unless they assign it to themselves.

PR #481, opened 2022-11-04
feat: added setters for hooks

ramenforbreakfast:


apps/smart-contracts/core/test/Collateral.test.ts
```diff
@@ -42,6 +42,24 @@ describe('=> Collateral', () => {
...
+      deployer,
+      deployer,
+      await collateral.SET_WITHDRAW_HOOK_ROLE()
+    )
+    await grantAndAcceptRole(
+      collateral,
+      deployer,
+      deployer,
+      await collateral.SET_MANAGER_WITHDRAW_HOOK_ROLE()
+    )
```
XavierEkkel, line 62: each of these should only be in the beforeEach for their respective functions
otherwise, it might be succeeding because of the other roles, not because of the specific role we enabled for that function's tests

XavierEkkel (approved):
LGTM after comment

PR #489, opened 2022-11-08
feat: add getReserve to Collateral

ramenforbreakfast:
As discussed in the design document, I decided to add a `getReserve` function to `Collateral` so that an external contract like `ManagerWithdrawHook` can easily retrieve the balance rather than having to import a ERC20 interface and pull the `baseToken` contract to read the balance from.

Also fixed the `Collateral` smock fixture that will be used later, it is an upgradeable contract so the constructor should not have any parameters.


apps/smart-contracts/core/test/Collateral.test.ts
```diff
@@ -384,4 +384,18 @@ describe('=> Collateral', () => {
...
+    beforeEach(async () => {
+      await setupCollateral()
+    })
+
+    it("returns contract's base token balance", async () => {
+      await baseToken.connect(deployer).mint(collateral.address, parseEther('1'))
+      const contractBalanceBefore = await baseToken.balanceOf(collateral.address)
+      expect(contractBalanceBefore).to.be.gt(0)
+
+      expect(await collateral.getReserve()).to.eq(contractBalanceBefore)
```
XavierEkkel, line 398:
```suggestion
      await baseToken.connect(deployer).mint(collateral.address, parseEther('1'))
      const contractBalance = await baseToken.balanceOf(collateral.address)
      expect(contractBalance).to.be.eq(parseEther('1'))

      expect(await collateral.getReserve()).to.eq(contractBalance)
```

XavierEkkel (approved):
LGTM after nit

PR #492, opened 2022-11-08
feat: add hook logic to ManagerWithdrawHook

ramenforbreakfast:
I ultimately decided to remove the `onlyCollateral` modifier to the `hook` here because it is only a view function. We cannot change it to a view function to abide to our interface, but it doesn't matter and the modifier can be added back in if we ever decide to perform write operations.

This also simplifies testing since if `onlyCollateral` was on there, I would need to introduce a wrapper contract for testing purposes. Usually our control modifiers reference something that is only used for checking access, but here the contract itself also uses the `collateral` contract to determine reserve requirements, and you cannot sign a transaction using a contract.

I decided to make `getRequiredReserve` a public function since I think it would be useful to obtain it externally.


apps/smart-contracts/core/test/ManagerWithdrawHook.test.ts
```diff
@@ -306,4 +306,73 @@ describe('=> ManagerWithdrawHook', () => {
...
+
+  describe('# getRequiredReserve', () => {
+    beforeEach(async () => {
+      await setupManagerWithdrawHook()
+    })
+
+    it('reverts if deposit record not configured', async () => {
+      await managerWithdrawHook.connect(deployer).setDepositRecord(ZERO_ADDRESS)
+
+      await expect(managerWithdrawHook.getRequiredReserve()).reverted
```
XavierEkkel, line 318: nice concise new style - let's use this going forward

apps/smart-contracts/core/test/ManagerWithdrawHook.test.ts
```diff
@@ -306,4 +306,73 @@ describe('=> ManagerWithdrawHook', () => {
...
+    })
+
+    it('returns correct reserve requirement', async () => {
```
XavierEkkel, line 321:
```suggestion
    it('returns required reserve', async () => {
```

apps/smart-contracts/core/test/ManagerWithdrawHook.test.ts
```diff
@@ -306,4 +306,73 @@ describe('=> ManagerWithdrawHook', () => {
...
+      const requiredReserve = TEST_GLOBAL_DEPOSIT_CAP.mul(TEST_MIN_RESERVE_PERCENTAGE).div(
+        RESERVE_DENOMINATOR
+      )
+      const amountToWithdraw = parseEther('1')
+      collateral.getReserve.returns(requiredReserve.add(amountToWithdraw).add(1))
+      expect(await collateral.getReserve()).to.be.gt(requiredReserve.add(amountToWithdraw))
+
+      await managerWithdrawHook
+        .connect(user)
+        .hook(user.address, amountToWithdraw.add(1), amountToWithdraw)
```
XavierEkkel, line 375: i think it's a bit confusing with the 2nd and 3rd arguments - can we make the variable for the ignored argument explicitly be a `IGNORED_ARGUMENT` variable?
same for elsewhere (could also do this in a separate PR)

apps/smart-contracts/core/test/ManagerWithdrawHook.test.ts
```diff
@@ -306,4 +306,73 @@ describe('=> ManagerWithdrawHook', () => {
         .withArgs(TEST_MIN_RESERVE_PERCENTAGE)
     })
   })
+
+  describe('# getRequiredReserve', () => {
+    beforeEach(async () => {
+      await setupManagerWithdrawHook()
+    })
+
+    it('reverts if deposit record not configured', async () => {
```
XavierEkkel, line 315:
```suggestion
    it('reverts if deposit record not set', async () => {
```

apps/smart-contracts/core/test/ManagerWithdrawHook.test.ts
```diff
@@ -306,4 +306,73 @@ describe('=> ManagerWithdrawHook', () => {
...
+      depositRecord.getGlobalNetDepositAmount.returns(TEST_GLOBAL_DEPOSIT_CAP)
+      const requiredReserve = TEST_GLOBAL_DEPOSIT_CAP.mul(TEST_MIN_RESERVE_PERCENTAGE).div(
+        RESERVE_DENOMINATOR
+      )
+      collateral.getReserve.returns(requiredReserve)
+      const amountToWithdraw = 1
+      expect(await collateral.getReserve()).to.be.lt(requiredReserve.add(amountToWithdraw))
+
+      await expect(
+        managerWithdrawHook.connect(user).hook(user.address, amountToWithdraw + 1, amountToWithdraw)
```
XavierEkkel, line 346: shouldnt it just be withdrawing amountToWithdraw? no +1 necessary because already exactly at required reserve beforehand

apps/smart-contracts/core/test/ManagerWithdrawHook.test.ts
```diff
@@ -306,4 +306,73 @@ describe('=> ManagerWithdrawHook', () => {
...
+    })
+
+    it('allows withdrawal down to exactly reserve minimum', async () => {
+      depositRecord.getGlobalNetDepositAmount.returns(TEST_GLOBAL_DEPOSIT_CAP)
+      const requiredReserve = TEST_GLOBAL_DEPOSIT_CAP.mul(TEST_MIN_RESERVE_PERCENTAGE).div(
+        RESERVE_DENOMINATOR
+      )
+      const amountToWithdraw = parseEther('1')
+      collateral.getReserve.returns(requiredReserve.add(amountToWithdraw))
+      expect(await collateral.getReserve()).to.be.gte(requiredReserve.add(amountToWithdraw))
```
XavierEkkel, line 357: confusing setup - can just do the same as above and ensure that reserve returns requiredReserve + 1
then just withdraw 1

apps/smart-contracts/core/test/ManagerWithdrawHook.test.ts
XavierEkkel, line 346:
```suggestion

      await expect(
        managerWithdrawHook.connect(user).hook(user.address, 1, IGNORED_ARGUMENT)
```

apps/smart-contracts/core/test/ManagerWithdrawHook.test.ts
XavierEkkel: similarly for one below, can just return requiredReserve + 2, then withdraw only 1

XavierEkkel (approved):
LGTM after comments

apps/smart-contracts/core/contracts/ManagerWithdrawHook.sol
```diff
@@ -24,16 +24,16 @@ contract ManagerWithdrawHook is
...
-  }
-
   function hook(
     address _sender,
     uint256 _amountBeforeFee,
     uint256 _amountAfterFee
-  ) external override {}
+  ) external override {
+    require(
+      _collateral.getReserve() >= getRequiredReserve() + _amountAfterFee,
```
XavierEkkel, line 33: nit: clearer way of writing
```suggestion
      _collateral.getReserve() - _amountAfterFee >= getRequiredReserve(),
```

apps/smart-contracts/core/contracts/ManagerWithdrawHook.sol
```diff
@@ -79,4 +79,10 @@ contract ManagerWithdrawHook is
   function getMinReservePercentage() external view override returns (uint256) {
     return _minReservePercentage;
   }
+
+  function getRequiredReserve() public view override returns (uint256) {
```
XavierEkkel, line 83:
```suggestion
  function getMinReserve() public view override returns (uint256) {
```
for consistency

apps/smart-contracts/core/test/ManagerWithdrawHook.test.ts
ramenforbreakfast: The + 1 is for the ignored argument, I think this sort of ties in to your previous argument where I should make it more clear using an IGNORED_ARGUMENT variable.

ramenforbreakfast: I think you meant to switch the arguments here (we ignore `amountBeforeFee`, not `amountAfterFee`), but understand what you mean.

PR #495, opened 2022-11-09
feat: add treasury and reserve denominator to Collateral

ramenforbreakfast:
This PR involves adding the treasury address and some elements to `Collateral` that were regrettably not recognized in our technical specification.

- Restating here from the technical doc, but the `treasury` address is needed since where to send fees to cannot be reliably determined from access controls, due to the contract not being held by a single entity.
- `IERC20` cannot be used since we need to know the base token decimals, and thus need to use `IERC20MetadataUpgradeable`. The base token decimal needs to be known because the base token might have a different fixed-precision than 18 and thus cannot just mint the exact same amount back to the user.
- Calculate base token denominator within the constructor.


XavierEkkel:
> * `IERC20` cannot be used since we need to know the base token decimals, and thus need to use `IERC20MetadataUpgradeable`. The base token decimal needs to be known because the base token might have a different fixed-precision than 18 and thus cannot just mint the exact same amount back to the user.

I think it doesn't matter if it's minted with identical decimals, only that it's minted in a consistent ratio.
The collateral tokens aren't using facing so I don't think it's worth making the contract more complicated just to maintain matching decimals.

apps/smart-contracts/core/contracts/Collateral.sol
```diff
@@ -92,10 +107,23 @@ contract Collateral is
...
+    view
+    override
+    returns (IERC20MetadataUpgradeable)
+  {
     return _baseToken;
   }
 
+  function getBaseTokenDenominator() external view override returns (uint256) {
+    return _baseTokenDenominator;
+  }
```
XavierEkkel, line 121: remove as discussed

apps/smart-contracts/core/contracts/Collateral.sol
```diff
@@ -12,13 +12,18 @@ contract Collateral is
   ERC20PermitUpgradeable,
   SafeAccessControlEnumerableUpgradeable
 {
-  IERC20 private _baseToken;
+  IERC20MetadataUpgradeable private _baseToken;
```
XavierEkkel, line 15: change back to `IERC20` as discussed

apps/smart-contracts/core/contracts/Collateral.sol
```diff
@@ -12,13 +12,18 @@ contract Collateral is
...
+  uint256 private _baseTokenDenominator;
```
XavierEkkel, line 16: make this immutable

apps/smart-contracts/core/contracts/Collateral.sol
```diff
@@ -33,12 +38,13 @@ contract Collateral is
   function initialize(
     string memory _name,
     string memory _symbol,
-    IERC20 _newBaseToken
+    IERC20MetadataUpgradeable _newBaseToken
```
XavierEkkel, line 41: add an arg after for `_baseTokenDecimals`

apps/smart-contracts/core/contracts/Collateral.sol
```diff
@@ -33,12 +38,13 @@ contract Collateral is
...
   ) public initializer {
     __SafeAccessControlEnumerable_init();
     __ERC20_init(_name, _symbol);
     __ERC20Permit_init(_name);
     _baseToken = _newBaseToken;
+    _baseTokenDenominator = 10**_newBaseToken.decimals();
```
XavierEkkel, line 47: change this to `10**_baseTokenDecimals`

apps/smart-contracts/core/contracts/Collateral.sol
XavierEkkel: note that similar to MiniSales.sol this will need to be tested indirectly

ramenforbreakfast:
**Note**
- Remove `IERC20Metadata`.
- Make `_baseToken` and add a `_baseTokenDenominator` as `immutable` variables initialized in the constructor.
- No need to dynamically calculate decimals, just directly pass them in.

XavierEkkel (approved):
LGTM after comments

PR #496, opened 2022-11-09
feat: add deposit logic to Collateral

ramenforbreakfast:
This PR contains the `deposit` logic and associated tests. Some items that might be needed as context for the reviewer:
- Added a dev comment clarifying why we do not use the old fee logic that tries to prevent a user from skipping fees.
- I decided to keep just the `amountAfterFee` and the `fee` in the event, since the amount minted is already communicated via `Transfer('0x0'...` and the `amountBeforeFee` can be determined from these values.
- Instead of reverting upon `_amount > _fee` as specified in our tech doc (this is how we did it in the old contract), I did some thinking and realized there is no need for this since math is already underflow safe and there is no situation where this could happen except when the fee is 100%. A deposit of 0 would have been caught by this statement, but its easier just to revert if 0.


ramenforbreakfast:
Updated branch to reflect `Collateral` changes merged in recent PRs.

apps/smart-contracts/core/contracts/Collateral.sol
```diff
@@ -49,7 +49,25 @@ contract Collateral is
...
   }
 
-  function deposit(uint256 _amount) external override {}
+  /**
+   * @dev Depending on how small the deposit fee is set, the fee might be
+   * zero. However, in a practical scenario, even a deposit fee as low as
+   * 0.1% would require a depositor to deposit less than $0.001 to not pay
+   * a fee. Such an attack would not be economically feasible and is not worth
+   * protecting against.
+   */
```
XavierEkkel, line 58: this is only true assuming the collateral is USDC, it may be economically feasible in other cases (or on other chains)
i think we should just have a check that _fee > 0 if depositFee > 0

apps/smart-contracts/core/contracts/Collateral.sol
```diff
@@ -49,7 +49,25 @@ contract Collateral is
...
+  function deposit(uint256 _amount) external override {
+    require(_amount > 0, "cannot deposit zero");
```
XavierEkkel, line 60:
```suggestion
    require(_amount > 0, "amount = 0");
```

apps/smart-contracts/core/contracts/Collateral.sol
```diff
@@ -49,7 +49,25 @@ contract Collateral is
...
+    uint256 _fee = (_amount * _depositFee) / FEE_DENOMINATOR;
+    uint256 _amountAfterFee = _amount - _fee;
```
XavierEkkel, line 62: only declare variables right before they are needed
so this should be moved above the if statement

apps/smart-contracts/core/contracts/Collateral.sol
```diff
@@ -49,7 +49,25 @@ contract Collateral is
...
+    _baseToken.transferFrom(msg.sender, address(this), _amount);
+    _baseToken.transfer(_treasury, _fee);
+    if (address(_depositHook) != address(0)) {
+      _depositHook.hook(msg.sender, _amount, _amountAfterFee);
+    }
+    _mint(msg.sender, (_amountAfterFee * 1e18) / _baseTokenDenominator);
```
XavierEkkel, line 68: why do we need to scale by 1e18 when amount should already be scaled by its decimals?

apps/smart-contracts/core/contracts/Collateral.sol
```diff
@@ -49,7 +49,25 @@ contract Collateral is
...
 
-  function deposit(uint256 _amount) external override {}
+  /**
+   * @dev Depending on how small the deposit fee is set, the fee might be
+   * zero. However, in a practical scenario, even a deposit fee as low as
+   * 0.1% would require a depositor to deposit less than $0.001 to not pay
+   * a fee. Such an attack would not be economically feasible and is not worth
+   * protecting against.
+   */
+  function deposit(uint256 _amount) external override {
```
XavierEkkel, line 59: add reentrancy guard
same for withdraw+managerWithdraw

apps/smart-contracts/core/contracts/interfaces/ICollateral.sol
```diff
@@ -6,6 +6,12 @@ import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import "@openzeppelin/contracts-upgradeable/token/ERC20/IERC20Upgradeable.sol";
 
 interface ICollateral is IERC20Upgradeable {
+  event Deposit(
+    address indexed depositor,
+    uint256 indexed amount,
```
XavierEkkel, line 11: let's call this `amountAfterFee` to be clearer

apps/smart-contracts/core/contracts/interfaces/ICollateral.sol
```diff
@@ -6,6 +6,12 @@ import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
...
+    uint256 indexed fee
```
XavierEkkel, line 12: why are we indexing on amounts?

apps/smart-contracts/core/test/Collateral.test.ts
```diff
@@ -463,4 +495,149 @@ describe('=> Collateral', () => {
...
+  describe('# deposit', () => {
+    beforeEach(async () => {
+      await setupCollateral()
+      await baseToken.mint(user1.address, parseEther('1'))
+      await baseToken.connect(user1).approve(collateral.address, parseEther('1'))
+      await collateral.connect(deployer).setTreasury(treasury.address)
+      await collateral.connect(deployer).setDepositFee(TEST_DEPOSIT_FEE)
+      await collateral.connect(deployer).setDepositHook(depositHook.address)
+    })
+    it('reverts if depositing zero', async () => {
```
XavierEkkel, line 508: nit: keep concise
```suggestion
    it('reverts if deposit 0', async () => {
```

apps/smart-contracts/core/test/Collateral.test.ts
```diff
@@ -463,4 +495,149 @@ describe('=> Collateral', () => {
...
+      expect(amountToDeposit).to.be.gt(0)
+      expect(await baseToken.allowance(user1.address, collateral.address)).to.be.eq(amountToDeposit)
+      const fee = amountToDeposit.mul(await collateral.getDepositFee()).div(FEE_DENOMINATOR)
+
+      await collateral.connect(user1).deposit(amountToDeposit)
+
+      expect(await baseToken.balanceOf(treasury.address)).to.eq(treasuryBTBefore.add(fee))
+    })
+
+    it('mints equivalent Collateral to depositor based on base token decimals', async () => {
```
XavierEkkel, line 571:
```suggestion
    it('mints to depositor adjusting for base token decimals', async () => {
```

XavierEkkel, line 571: add assertions that the decimals are different for base vs. collateral

XavierEkkel, line 571: add test for when base+collateral decimals are the same, and a test for when base decimals > collateral decimals

XavierEkkel, line 571: add test for when deposit fee is set to 0

apps/smart-contracts/core/contracts/Collateral.sol
ramenforbreakfast: This has been clarified in discord and will resolve this

ramenforbreakfast:
Submitted various fixes and amendments:
- Changed setup to allow for dynamic base token setups, which will allow us to cleanly test base tokens with different decimal precision.
- Changed revert case to revert if calculated fee is 0 and fee rate > 0 as stated in the review. Otherwise, check if the amount = 0 (if amount = 0 when fee rate > 0 then it will be caught be the first condition).
- Made only `depositor` in the `Deposit` event indexed, changed `amount` to `amountAfterFee`
- Added testcases for different base token decimals, and tests for the new revert logic, and ensuring deposits can still happen when fee = 0.

apps/smart-contracts/core/contracts/Collateral.sol
```diff
@@ -49,7 +51,22 @@ contract Collateral is
...
-  function deposit(uint256 _amount) external override {}
+  function deposit(uint256 _amount) external override nonReentrant {
+    uint256 _fee = (_amount * _depositFee) / FEE_DENOMINATOR;
+    if (_depositFee > 0) {
+      require(_fee > 0, "fee = 0");
+    } else {
+      require(_amount > 0, "amount = 0");
+    }
+    _baseToken.transferFrom(msg.sender, address(this), _amount);
+    _baseToken.transfer(_treasury, _fee);
```
XavierEkkel, line 62: this is a logic error (and is why we shouldn't use `treasury`, because it could refer to either collateral treasury or prePO treasury - the two should be kept separate)

ramenforbreakfast:
Rebased and edited to use the new `manager` instead of `treasury` variable, removed fee transfers and instead approve fee funds to the deposit hook.

apps/smart-contracts/core/test/Collateral.test.ts
```diff
@@ -568,4 +601,237 @@ describe('=> Collateral', () => {
...
+  describe('# deposit', () => {
+    beforeEach(async () => {
+      await setupCollateral()
+      await baseToken.mint(user1.address, parseUnits('1', USDC_DECIMALS))
+      await baseToken.connect(user1).approve(collateral.address, parseUnits('1', USDC_DECIMALS))
+      await collateral.connect(deployer).setDepositFee(TEST_DEPOSIT_FEE)
+      await collateral.connect(deployer).setDepositHook(depositHook.address)
+    })
+
+    it('reverts if deposit 0 and fee rate 0', async () => {
```
XavierEkkel, line 614: nit:
```suggestion
    it('reverts if deposit = 0 and deposit fee = 0%', async () => {
```

XavierEkkel: similarly for other tests

apps/smart-contracts/core/test/Collateral.test.ts
```diff
@@ -568,4 +601,237 @@ describe('=> Collateral', () => {
...
+      expect(await collateral.getDepositFee()).to.be.gt(0)
+
+      await expect(collateral.connect(user1).deposit(0)).to.be.revertedWith('fee = 0')
+    })
+
+    it('reverts if deposit > 0, fee 0, and fee rate > 0 ', async () => {
+      expect(await collateral.getDepositFee()).to.be.gt(0)
+      const amountToDeposit = BigNumber.from(1)
+      // expect fee to be zero
+      expect(amountToDeposit.mul(await collateral.getDepositFee()).div(FEE_DENOMINATOR)).to.eq(0)
```
XavierEkkel, line 630: nit: assign the argument to a feeAmount variable first for readability, and then can also remove the comment above

apps/smart-contracts/core/test/Collateral.test.ts
```diff
@@ -568,4 +601,237 @@ describe('=> Collateral', () => {
...
+    it('reverts if hook reverts', async () => {
+      const amountToDeposit = await baseToken.balanceOf(user1.address)
+      expect(amountToDeposit).to.be.gt(0)
+      expect(await baseToken.allowance(user1.address, collateral.address)).to.be.eq(amountToDeposit)
+      depositHook.hook.reverts()
+
+      await expect(collateral.connect(user1).deposit(amountToDeposit)).to.be.reverted
+    })
+
+    it('transfers amount to contract from depositor', async () => {
```
XavierEkkel, line 666: nit:
```suggestion
    it('transfers amount from depositor to contract', async () => {
```

apps/smart-contracts/core/test/Collateral.test.ts
```diff
@@ -568,4 +601,237 @@ describe('=> Collateral', () => {
...
+        .sub(fee)
+        .mul(parseEther('1'))
+        .div(GREATER_DECIMAL_DENOMINATOR)
+
+      await collateral.connect(user1).deposit(amountToDeposit)
+
+      expect(await collateral.balanceOf(user1.address)).to.eq(userCTBefore.add(expectedCT))
+    })
+
+    it('allows a deposit > 0 if fee rate 0', async () => {
```
XavierEkkel, line 765: nit:
```suggestion
    it('allows deposit > 0 if deposit fee = 0%', async () => {
```

apps/smart-contracts/core/test/Collateral.test.ts
```diff
@@ -568,4 +601,237 @@ describe('=> Collateral', () => {
...
+      await collateral.connect(deployer).setDepositFee(0)
+      const userBTBefore = await baseToken.balanceOf(user1.address)
+      const contractBTBefore = await baseToken.balanceOf(collateral.address)
+      const userCTBefore = await collateral.balanceOf(user1.address)
+      const amountToDeposit = userBTBefore
+      // expect fee to be zero
+      expect(amountToDeposit.mul(await collateral.getDepositFee()).div(FEE_DENOMINATOR)).to.eq(0)
```
XavierEkkel, line 772: nit: same comment here as before

apps/smart-contracts/core/test/Collateral.test.ts
```diff
@@ -568,4 +601,237 @@ describe('=> Collateral', () => {
...
+      expect(secondFee).to.not.eq(firstFee)
+      await baseToken.mint(user1.address, secondAmountToDeposit)
+      await baseToken.connect(user1).approve(collateral.address, secondAmountToDeposit)
+
+      await collateral.connect(user1).deposit(secondAmountToDeposit)
+
+      expect(await baseToken.allowance(collateral.address, depositHook.address)).to.eq(secondFee)
+    })
+
+    it('mints to depositor adjusting for when decimals > base token decimals', async () => {
```
XavierEkkel, line 711: nit:
```suggestion
    it('mints decimal-adjusted amount to depositor if decimals > base token decimals', async () => {
```

XavierEkkel: similarly for other tests

XavierEkkel (approved):
LGTM after nits

apps/smart-contracts/core/test/Collateral.test.ts
```diff
@@ -568,4 +601,237 @@ describe('=> Collateral', () => {
...
+      const secondFee = secondAmountToDeposit
+        .mul(await collateral.getDepositFee())
+        .div(FEE_DENOMINATOR)
+      expect(secondFee).to.not.eq(firstFee)
+      await baseToken.mint(user1.address, secondAmountToDeposit)
+      await baseToken.connect(user1).approve(collateral.address, secondAmountToDeposit)
+
+      await collateral.connect(user1).deposit(secondAmountToDeposit)
+
+      expect(await baseToken.allowance(collateral.address, depositHook.address)).to.eq(secondFee)
```
XavierEkkel, line 708: add assertion that fee balance stays in the collateral contract

PR #502, opened 2022-11-11
feat: add manager withdrawal logic to Collateral

ramenforbreakfast:


ramenforbreakfast:
Closing due to it being simpler to re open rather than rebase and fix conflicts since this will be a parallel PR.

PR #503, opened 2022-11-12
feat: add manager withdraw logic to Collateral

ramenforbreakfast:
There are a few duplicate items here that will be merged together with the deposit and withdraw PRs, mostly in the test suite setup.

Also to allow this to be submitted in parallel, instead of actually depositing into the `Collateral` contract, I just transfer over base token balance since all we need to do is verify smock behavior with the hook, we don't actually need to verify reserve requirements.


XavierEkkel (approved):
LGTM after nits

PR #512, opened 2022-11-15
fix: use base IERC20 and _saleTokenDenominator name in MiniSales

combatdave:
- [ ]  [https://github.com/prepo-io/prepo-monorepo/blob/main/apps/smart-contracts/token/contracts/mini-sales/MiniSales.sol](https://github.com/prepo-io/prepo-monorepo/blob/main/apps/smart-contracts/token/contracts/mini-sales/MiniSales.sol)
    - should use `IERC20` instead of `IERC20Metadata`
    - `saleTokenDenominator` rather than `saleTokenDecimals`


XavierEkkel (approved):
LGTM!

ramenforbreakfast (approved):
LGTM!

PR #514, opened 2022-11-15
refactor: rename deposit record

davidprepo:
change CollateralDepositRecord.sol to DepositRecord.sol


ramenforbreakfast:
Will continue reviewing this later, but seems like we can't change that `developer.js` file or it will break the frontend doc build due to the doc still being named `CollateralDepositRecord`. We will need to change these docs anyway later so lets leave that out of this PR.

apps/smart-contracts/core/contracts/interfaces/IDepositRecord.sol
```diff
@@ -52,7 +52,7 @@ interface ICollateralDepositRecord {
   function recordWithdrawal(uint256 amount) external;
 
   /**
-   * @notice Sets the global cap on assets backing Collateral in circulation.
+   * @notice Sets the global deposit cap on assets.
```
XavierEkkel, line 55:
```suggestion
   * @notice Sets the global net deposit cap.
```

XavierEkkel (approved):
LGTM after Michael's comment and after ensuring the final PR compiles + passes tests (can look in the `Checks` tab of the PR)

ramenforbreakfast (approved):
LGTM, but you need to rename the deployment script to `DepositRecord.ts`, validated tests are working.

apps/smart-contracts/core/contracts/interfaces/IDepositRecord.sol
davidprepo: Fixed in the latest commit.

PR #518, opened 2022-11-16
feat: renaming events to nouns and using chai event matcher

davidprepo:
Some of our old core tests utilize an events.ts where getters for every single event are defined to grab them. However, we want to refactor these old tests with the new way of doing things where instead of using events.ts, event emission is verified.


ramenforbreakfast (approved):
LGTM, but you should just remove `AccountAccessController`

davidprepo:
> LGTM, but you should just remove `AccountAccessController`

You want me to remove AccountAccessController.sol and all related files? This should be a separate PR right?

davidprepo:
Removing AccountAccessController in separate PR per call.

PR #520, opened 2022-11-16
refactor: removing underscores from core contracts

davidprepo:
Parent PR: https://github.com/prepo-io/prepo-monorepo/pull/519/


apps/smart-contracts/core/contracts/PrePOMarket.sol
```diff
@@ -46,8 +46,8 @@ contract PrePOMarket is IPrePOMarket, Ownable, ReentrancyGuard {
     address _newCollateral,
     ILongShortToken _newLongToken,
     ILongShortToken _newShortToken,
-    uint256 _newFloorLongPrice,
-    uint256 _newCeilingLongPrice,
+    uint256 _newFloorLongPayout,
```
ramenforbreakfast, line 49: I think this is due to the fact that main has been updated since this PR? Please rebase this and I would also resubmit this across multiple PRs, this PR is too large and I am pretty sure all of these changes do not need to exist within the same PR as they should be independent of one another.

Perhaps consider grouping them in groups of 4? I think that is a manageable size.

ramenforbreakfast (changes requested):

PR #521, opened 2022-11-16
feat: adding testcases for reserve exactly or exceeding global net de…

davidprepo:
Add 2 testcases for ManagerWithdrawHook to ensure that hook does not revert when getReserve is exactly or exceeds getGlobalNetDepositAmount


apps/smart-contracts/core/test/ManagerWithdrawHook.test.ts
```diff
@@ -374,5 +374,21 @@ describe('=> ManagerWithdrawHook', () => {
 
       await managerWithdrawHook.connect(user).hook(user.address, IGNORED_ARGUMENT, amountToWithdraw)
     })
+
+    it('allows withdrawal when reserve is higher than global net deposit amount', async () => {
```
ramenforbreakfast, line 378: Should use "if" wording and more concise wording for comparisons, so `allows withdrawal if reserve is > global net deposit amount`. Please do the same for the testcase below.

apps/smart-contracts/core/test/ManagerWithdrawHook.test.ts
```diff
@@ -374,5 +374,21 @@ describe('=> ManagerWithdrawHook', () => {
...
+      depositRecord.getGlobalNetDepositAmount.returns(TEST_GLOBAL_DEPOSIT_CAP)
+      collateral.getReserve.returns(TEST_GLOBAL_DEPOSIT_CAP.add(1))
+
+      const amountToWithdraw = 1
+      await managerWithdrawHook.connect(user).hook(user.address, IGNORED_ARGUMENT, amountToWithdraw)
+    })
+
+    it('allows withdrawal when reserve is exactly at global net deposit amount', async () => {
```
ramenforbreakfast, line 386: I would put this above the previous testcase.

ramenforbreakfast (approved):
LGTM besides some minor nits

apps/smart-contracts/core/test/ManagerWithdrawHook.test.ts
XavierEkkel: `is` is also a filler word we want to avoid, so this would be better:
`allows withdrawal if reserve > global net deposit amount`

davidprepo: Understood, will follow in future.

apps/smart-contracts/core/test/ManagerWithdrawHook.test.ts
davidprepo: rgr

PR #522, opened 2022-11-16
feat: removing depositRecord from hook constructors

davidprepo:
This ⇒ https://github.com/prepo-io/prepo-monorepo/pull/48/files#r867322655 involves making depositRecord not a constructor argument for DepositHook , WithdrawHook, and ManagerWithdrawHook . The end result should be that all 3 hooks should not have any constructor arguments. Will need to modify existing tests.


apps/smart-contracts/core/test/ManagerWithdrawHook.test.ts
```diff
@@ -310,6 +304,7 @@ describe('=> ManagerWithdrawHook', () => {
   describe('# getMinReserve', () => {
     beforeEach(async () => {
       await setupManagerWithdrawHook()
+      await managerWithdrawHook.connect(deployer).setDepositRecord(depositRecord.address)
```
ramenforbreakfast, line 307: I would put this in `setupManagerWithdrawHook`. `getSignersAndDeployHook` is meant to only perform deployments so that setter tests can test without the setter being called before (since it only performs deployment and configuration of external contracts). If a testcase is calling `setupManagerWithdrawHook`, that means that the testcase is not testing a setter and setters can be called to configure it ahead of time.

apps/smart-contracts/core/contracts/DepositHook.sol
```diff
@@ -32,7 +28,9 @@ contract DepositHook is IDepositHook, SafeAccessControlEnumerable {
     uint256 _amountAfterFee
   ) external override onlyCollateral {
     require(_depositsAllowed, "deposits not allowed");
-    _depositRecord.recordDeposit(_sender, _amountAfterFee);
+    if (address(_depositRecord) != address(0)) {
```
ramenforbreakfast, line 31: This is not necessary imo. This `DepositHook` is expected to be configured with one already and it's not necessary to waste effort checking this every time. If we expect otherwise we'd simply deploy a `DepositHook` that does not expect a `depositRecord`. 

At initialization, if for some reason `_depositRecord` is not configured, this transaction would simply revert since no method would exist on the zero address.

apps/smart-contracts/core/contracts/WithdrawHook.sol
```diff
@@ -78,7 +74,9 @@ contract WithdrawHook is IWithdrawHook, SafeAccessControlEnumerable {
       );
       _userToAmountWithdrawnThisPeriod[_sender] += _amountBeforeFee;
     }
-    _depositRecord.recordWithdrawal(_amountBeforeFee);
+    if (address(_depositRecord) != address(0)) {
```
ramenforbreakfast, line 77: Same here, lets not touch the implementation.

ramenforbreakfast (approved):
LGTM once my comments are addressed.

apps/smart-contracts/core/test/ManagerWithdrawHook.test.ts
davidprepo: Rgr, actually did it that way at first but then wasn't sure the level of granularity required from using or not using the setupManagerWithdrawHook method.

apps/smart-contracts/core/contracts/DepositHook.sol
davidprepo: rgr

apps/smart-contracts/core/contracts/WithdrawHook.sol
davidprepo: rgr

PR #523, opened 2022-11-16
docs: splitting readmes into sensible parenting

davidprepo:
README reorganization (have a README in the smart-contracts folder describing generic things that apply to all directories, e.g. visual studio code, generic setup. READMEs in subdirectories should only be things specific to that subdirectory. shared contracts folder should just link the main SC README)


ramenforbreakfast (approved):
LGTM!

PR #526, opened 2022-11-17
feat: markets slide up

chrisling-dev:
https://www.notion.so/prepo/FE-Project-Management-d3878c29e9534234a50f4f8838525bee?p=e5eb74d099c0486db8d087aea6cc0f20&pm=s

Scrollable demo with mocked data:
https://user-images.githubusercontent.com/81092286/202968279-256d2766-ffc4-4c0f-8982-66e7c1bcd5d8.mov


apps/frontend/core-dapp/src/features/trade/SlideUpCard.tsx
```diff
@@ -0,0 +1,99 @@
...
+const SlideUpCard: React.FC<Props> = ({ children, onClose, show, title }) => {
+  const handleClose = (): void => {
+    if (onClose) onClose()
+  }
+  return (
+    <Wrapper show={show}>
+      <RelativeWrapper>
+        <InnerWrapper>
+          <Flex justifyContent="space-between" mb={12}>
+            {typeof title === 'string' || title === undefined ? (
```
georgesmithishere, line 77: title === undefined and then we show title as text in Typography - it sounds wrong

georgesmithishere (approved):

apps/frontend/core-dapp/src/features/trade/SlideUpCard.tsx
chrisling-dev: Ahh I purposely had this so there's always something on the left to keep close icon on the right (because of space-between), otherwise close icon will come to the left when title is empty

XavierEkkel (approved):
LGTM from a UI perspective

but noting that in followup PRs we should:
-use the new tick design for the selected market
-allow the currently-selected market to be clickable

PR #528, opened 2022-11-17
refactor: remove underscores from hook contracts

davidprepo:
Related:
https://github.com/prepo-io/prepo-monorepo/pull/527
https://github.com/prepo-io/prepo-monorepo/pull/529

PR #527, opened 2022-11-17
refactor: no underscores in PrePOMarket.sol and PrePOMarketFactory.sol

davidprepo:
Related:
https://github.com/prepo-io/prepo-monorepo/pull/528
https://github.com/prepo-io/prepo-monorepo/pull/529

PR #529, opened 2022-11-17
refactor: underscores removed from remaining core contracts

davidprepo:
Related:
https://github.com/prepo-io/prepo-monorepo/pull/527
https://github.com/prepo-io/prepo-monorepo/pull/528


ramenforbreakfast (approved):
LGTM

PR #530, opened 2022-11-17
feat: fee reimbursement

davidprepo:
Head PR for train:

- [ ]  `PPO` fee reimbursement
    - `DepositHook.sol` calls a function on a new contract, `FeeReimbursement.sol` that will attribute the fee to a user.
    - `FeeReimbursement.sol` will have `PPO` deposited into it from governance for users to claim. Users will claim a reimbursement on the `USDC` fees attributed to them in `PPO` based on a fixed price read from a `MiniSales` contract.

My understanding is that USDC isn't to be moved to the FeeReimbursement contract and therefore the MiniSales purchase() function can't be used, rather that the FeeReimbursement contract is to transfer PPO from it's own reserves. Therefore the FeeReimbursement contract needs to know the correct amount of PPO to transfer for a given amount of USDC.

add getAmountOut to MiniSales


ramenforbreakfast (approved):
Makes sense to me! However, I would suggest naming it `getSaleForPayment`, to follow the same naming convention used throughout the contract.

PR #536, opened 2022-11-18
fix: false positive deploy action

chrisling-dev:
https://www.notion.so/prepo/FE-Project-Management-d3878c29e9534234a50f4f8838525bee?p=57bac553243147cbaa6de2e45b69bdef&pm=s


ramenforbreakfast:
LGTM, tested with a SC PR temporarily.

ramenforbreakfast (dismissed):

.github/workflows/deploy-preview-apps.yml
```diff
@@ -59,54 +59,55 @@ jobs:
     if: github.event_name == 'pull_request'
     outputs:
       matrix: ${{ steps.set-matrix.outputs.matrix }}
+      projectIds: ${{ join(fromJson(steps.set-matrix.outputs.matrix).include.*.projectId) }}
```
chrisling-dev, line 62: change 1: join all changed FE app's id into 'projectId1,projectId2,projectId3'

.github/workflows/deploy-preview-apps.yml
```diff
@@ -123,31 +124,23 @@ jobs:
...
-          echo $matrix
-          echo $matrix | jq .
-          echo $matrix | json2yaml
-          if [[ $matrix != '' ]] ; then echo "Matrix is not empty string!" ; else echo "Matrix is empty string!" ; fi
-          if [[ $matrix != '[]' ]] ; then echo "Matrix is not empty array!" ; else echo "Matrix is empty array!" ; fi
-          if [[ (needs.get-changed-apps.outputs.matrix != '' && needs.get-changed-apps.outputs.matrix != '[]') ]] ; then echo "Should run pipeline" ; else echo "Should not run pipeline" ; fi
-      
+          projectIdString='${{ needs.get-changed-apps.outputs.projectIds }}'
+          echo $projectIdString
+          if [[ ($projectIdString != '') ]] ; then echo "App changed, deploy!" ; else echo "No app changed." ; fi
```
chrisling-dev, line 129: change 2: instead of checking `needs.get-changed-apps.outputs.matrix === '' or ']'`, which will always be true because `matrix` is an object that looks like: `{ "include": [...changedProjects] }`, we should just check whether projectIdString is empty

projectIdString is derived from matrix.include, not matrix, which is the cause of problem

.github/workflows/deploy-preview-apps.yml
```diff
@@ -123,31 +124,23 @@ jobs:
...
+
       - name: Download changed apps artifact
         uses: actions/download-artifact@v2
         with:
           name: changed-apps-${{ needs.generate-artifact-key.outputs.artifact }}
 
-      - name: Print changed apps cache
-        run: cat changed_apps.json | jq -c -r .
-
```
chrisling-dev, line 141: change 3: this is unnecessary, we already print changed_apps.json in one of the previous steps

.github/workflows/deploy-preview-apps.yml
```diff
@@ -123,31 +124,23 @@ jobs:
...
   deploy-apps:
     name: Deploy apps
     needs: get-changed-apps
     runs-on: ubuntu-latest
-    if: |
-      (needs.get-changed-apps.outputs.matrix != '' && needs.get-changed-apps.outputs.matrix != '[]')
+    if: needs.get-changed-apps.outputs.projectIds != ''
```
chrisling-dev, line 140: change 4: only run "Deploy apps" if projectIds string is not empty

ramenforbreakfast (approved):
Tested via PR #591, no alert was sent when the PR was opened.

PR #538, opened 2022-11-18
refactor: Rename IHook to ICollateralHook

ramenforbreakfast:
The `IHook` contract contains references to collateral, which is typed as an `ICollateral` variable so that it can access things like reserve percentages, etc.

To differentiate Collateral related hooks from hooks for `PrePOMarket`s, I propose renaming `IHook` to `ICollateralHook`. 

Later on we should also organize contracts into `markets` and `collateral` folders, if the team agrees, we can add this as a backlog item. Right now I don't want to touch the directory structure as it will only slow down current development due to GitHub treating them as new files.


davidprepo (approved):
LGTM

XavierEkkel (approved):
LGTM

PR #539, opened 2022-11-18
feat: AllowlistMarketHook skeleton contract and interfaces

ramenforbreakfast:
I thought about this for a while, and I thought that it would be simpler given both our short-term needs and longer-term needs, to design the `PrePOMarketHook` as a shared hook amongst multiple `PrePOMarkets`.

It would be tedious to manage a hook for every single `PrePOMarket` launched, and the logic we plan to implement doesn't really need to be specific to one market. Most likely if we offer someone access to one market, we'd given them access to another market.

This architecture also provides flexibility in that you can still deploy multiple hooks for different markets if you wish, but most likely if we are just doing allowlist checking and possible fee reimbursement, I don't see why any of that needs to be an isolated contract from other markets.

This `AllowlistMarketHook` will be the contract that is used for both `mint` and `redeem` (separate instances) at first, until a `FeeReimbursementHook` is deployed later.

There is no need to implement pausing since this utilizes an access list.

This simplified architecture means only one contract needs to be written for the `mintHook` and `redeemHook`


davidprepo (approved):
LGTM

apps/smart-contracts/core/contracts/interfaces/IMarketHook.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity =0.8.7;
+
+interface IMarketHook {
+  event AllowedMarketsChange(address market, bool allowed);
```
XavierEkkel, line 5: nit:
```suggestion
  event AllowedMarketChange(address market, bool allowed);
```

apps/smart-contracts/core/contracts/AllowlistMarketHook.sol
```diff
@@ -0,0 +1,36 @@
+// SPDX-License-Identifier: UNLICENSED
```
XavierEkkel, line 1: wrong license for all these contracts

apps/smart-contracts/core/contracts/AllowlistMarketHook.sol
```diff
@@ -0,0 +1,36 @@
...
+  mapping(address => bool) private allowedMarkets;
+  IAccountList private allowlist;
+
+  function hook(
+    address _sender,
+    uint256 _amountBeforeFee,
+    uint256 _amountAfterFee
+  ) external override {}
+
+  function setAllowedMarket(address _market, bool _allowed)
```
XavierEkkel, line 17: what is this function for?

apps/smart-contracts/core/contracts/AllowlistMarketHook.sol
```diff
@@ -0,0 +1,36 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity =0.8.7;
+
+import "./interfaces/IAllowlistMarketHook.sol";
+import "prepo-shared-contracts/contracts/SafeOwnable.sol";
+
+contract AllowlistMarketHook is IAllowlistMarketHook, SafeOwnable {
+  mapping(address => bool) private allowedMarkets;
```
XavierEkkel, line 8: nit: wrong naming convention for mappings

ramenforbreakfast:
Closing since implementation has changed significantly.

PR #541, opened 2022-11-18
feat: PrePOMarket - adding hook setters, change to SafeOwnable, and make constants public

ramenforbreakfast:
Changes included in this PR are as follows:
- Changed `PrePOMarket` to use `SafeOwnable`, which meant also changing some parts of the test suite since ownership is not immediately transferred over.
- Due to the above changes, I also spotted one testcase where a setup was being repeated throughout the test and it should have just been put in the `beforeEach` of the testcase. I have done this now.
- Note that you will notice that there are probably better ways to organize some of the tests in `PrePOMarket.test.ts`. I do not want to include a refactoring effort in this PR but acknowledge that this should be done at some point, possibly after this is merged since we'll need to make further edits to `PrePOMarket`.
- Added setters for the `IMarketHook` contracts.
- In line with our `Collateral` contract, I have made the constants `public` and removed unnecessary getters for them when they were `private`.


davidprepo (commented):
LGTM but I don't think I'm competent enough here, maybe good to get a second opinion.

XavierEkkel (commented):
please split into independent single-purpose PRs

ramenforbreakfast:
Closing this, will reopen as separate PRs.

PR #544, opened 2022-11-19
feat: change deposit to accept recipient

ramenforbreakfast:
Due to the helper contract needing to call `deposit` on behalf of a user, we cannot directly limit credit deposits to `msg.sender` and must be more flexible. I do not see any reason why we need to do so for `redeem`, since we will most likely have UX designed in a way where users are not directly settling positions in `USDC`, but `preUSD`.

Because of this, I have added a `_recipient` to `deposit` and only take `baseToken` from `msg.sender` (the helper contract), but credit deposits to the `_recipient`.

This means that I have also modified the `deposit` testcases to have a separate `sender` and `recipient` to ensure this is differentiated. I have also added a testcase to validate the scenario where the funding address is the same as the `_recipient`.


apps/smart-contracts/core/contracts/Collateral.sol
```diff
@@ -58,7 +58,11 @@ contract Collateral is
    * (withdrawable by manager). Converts amount after fee from base token
    * units to collateral token units.
    */
-  function deposit(uint256 _amount) external override nonReentrant {
+  function deposit(uint256 _amount, address _recipient)
```
XavierEkkel, line 61: this approach won't help with the helper - see Discord

ramenforbreakfast: This has been discussed and clarified in a separate discussion, so will be re-requesting a review.

XavierEkkel, line 61: nit: let's put `_recipient` as the first argument, to be consistent with the order in `hook()`, `_mint()`, and the events

apps/smart-contracts/core/test/Collateral.test.ts
```diff
@@ -635,24 +635,33 @@ describe('=> Collateral', () => {
   })
 
   describe('# deposit', () => {
```
XavierEkkel, line 637: should add tests for:
-when caller ('sender') is the same as the recipient
-that (aside from above case) amount isnt minted to the sender

XavierEkkel (approved):
LGTM after comments

apps/smart-contracts/core/test/Collateral.test.ts
ramenforbreakfast: This is already achieved because in the `beforeEach` block, it is obvious and explicit that sender and recipient are initially setup differently, with an expect statement for extra clarity.

In addition, there is already a testcase for if `sender` is the same as `recipient`.

However, just to be extra sure, I will also add in all of our minting testcases, that collateral is not minted to the sender.

PR #545, opened 2022-11-19
feat: add IDepositTradeHelper and Uniswap packages

ramenforbreakfast:
Added some pre-emptive dev comments around `depositAndTrade` to provide context on function inputs.


apps/smart-contracts/core/package.json
```diff
@@ -48,6 +48,8 @@
         "@openzeppelin/contracts-upgradeable": "4.7.3",
         "@typechain/ethers-v5": "8.0.3",
         "@typechain/hardhat": "3.0.0",
+        "@uniswap/v3-core": "1.0.1",
```
XavierEkkel, line 51: will we need core as well? the interface only uses periphery

apps/smart-contracts/core/contracts/interfaces/IDepositTradeHelper.sol
```diff
@@ -0,0 +1,33 @@
...
+
+import "./ICollateral.sol";
+import "./IPrePOMarket.sol";
+import "@uniswap/v3-periphery/contracts/interfaces/ISwapRouter.sol";
+
+interface IDepositTradeHelper {
+  /**
+   * @dev `baseTokenAmount` will be taken from user and deposited to mint
+   * preUSD. This minted preUSD will be used entirely towards purchasing
+   * either Long or Short tokens from the corresponding UniswapV3 pool.
```
XavierEkkel, line 12:
```suggestion
   * @dev `baseTokenAmount` will be taken from msg.sender (assumes that msg.sender has already approved at least `baseTokenAmount` of the Collateral's base token) and then deposited to mint
   * Collateral. This minted Collateral will be used entirely towards purchasing
   * either Long or Short tokens from the corresponding UniswapV3 pool.
```

apps/smart-contracts/core/contracts/interfaces/IDepositTradeHelper.sol
```diff
@@ -0,0 +1,33 @@
...
+   * The UniswapV3 pool address does not need to be provided since the
+   * SwapRouter will automatically identify the pool based on token inputs.
+   * @param baseTokenAmount Base Token to be used towards position
+   * @param market PrePOMarket to purchase position for
+   * @param long Whether to purchase a long or short position
+   */
+  function depositAndTrade(
+    uint256 baseTokenAmount,
+    IPrePOMarket market,
+    bool long
```
XavierEkkel, line 23: why not just pass in the relevant pool directly instead? the relevant pool address can be retrieved off-chain rather than on-chain

apps/smart-contracts/core/contracts/interfaces/IDepositTradeHelper.sol
XavierEkkel, line 12: what value will be used for minAmountOut in the swap router?

apps/smart-contracts/core/package.json
ramenforbreakfast: We need the core because the `ISwapRouter` interface inherits from a contract in `v3-core`

ramenforbreakfast:
I have updated this based on our discussion here => https://www.notion.so/prepo/DepositTradeHelper-SC-Spec-37ec0e75d82f402d8c68ac49e1562a7e

Detailed descriptions of structs will be added here based on the information in this doc. Struct variable naming follows the naming convention used in the functions they will be used with (`ERC20Permit` and `SwapRouter`) to prevent confusion.

XavierEkkel (commented):
needs updating to reflect latest spec

ramenforbreakfast:
Updated with `getBaseToken` and removed immutable setters. Was going to upgrade to OZ 4.8.0 thinking they had removed the draft naming, but still has yet to be released.

apps/smart-contracts/core/contracts/interfaces/IDepositTradeHelper.sol
```diff
@@ -0,0 +1,50 @@
...
+    uint256 deadline;
+    uint8 v;
+    bytes32 r;
+    bytes32 s;
+  }
+
+  struct OffChainTradeParams {
+    address tokenOut;
+    address deadline;
+    address amountOutMinimum;
```
XavierEkkel, line 20:
```suggestion
    uint256 amountOutMinimum;
```

apps/smart-contracts/core/contracts/interfaces/IDepositTradeHelper.sol
```diff
@@ -0,0 +1,50 @@
...
+  function depositAndTrade(
+    uint256 baseTokenAmount,
+    Permit calldata baseTokenPermit,
+    Permit calldata collateralPermit,
+    OffChainTradeParams calldata tradeParams
+  ) external;
+
+  function getCollateral() external view returns (ICollateral);
+
+  function getBaseToken() external view returns (IERC20Permit);
```
XavierEkkel, line 47: nit: move above `getCollateral` to have consistent order with `depositAndTrade` arguments

apps/smart-contracts/core/contracts/interfaces/IDepositTradeHelper.sol
```diff
@@ -0,0 +1,50 @@
...
+    bytes32 s;
+  }
+
+  struct OffChainTradeParams {
+    address tokenOut;
+    address deadline;
+    address amountOutMinimum;
+    uint160 sqrtPriceLimitX96;
+  }
+
```
XavierEkkel, line 23: missing `POOL_FEE_TIER` constant

XavierEkkel (approved):
LGTM after comments

PR #556, opened 2022-11-23
feat: remove mintingFee and publicMinting

ramenforbreakfast:
While these two things could technically be separate PRs, they are so intertwined since they are both involved in test setup fixtures for both `PrePOMarket` and `PrePOMarketFactory` that the value of separating them out would not outweigh how annoying it would be to resolve conflicts aftewards.

Any of the `PrePOMarket` testcases I edited I also took the chance to modify their names to be more in line with our current naming scheme. But I do not want to turn this into a full blown test refactoring PR.


XavierEkkel (approved):
LGTM after nit

PR #557, opened 2022-11-23
feat: add return value to Collateral deposit and remove allowance

ramenforbreakfast:
Following in line with a safer approach in our `PrePOMarket` tech doc, I have also added a line to `Collateral`'s `deposit` to remove approvals given to the hook after calling.

As discussed, we need to add a return value to `deposit`, however, there is a mistake in the doc since we actually can't use `amountAfterFee` since that is the `baseToken` amount spent, not the actual `Collateral` minted. Hence why I now store the value since it's used twice.


ramenforbreakfast:
Have updated this to address test bug identified in #564

PR #563, opened 2022-11-23
feat: adding NFT access logic to DepositHook

davidprepo:


apps/smart-contracts/core/contracts/DepositHook.sol
```diff
@@ -65,4 +91,85 @@ contract DepositHook is IDepositHook, SafeAccessControlEnumerable {
...
+  function getCollectionScore(IERC721 _collection)
+    external
+    view
+    override
+    returns (uint256)
+  {
+    if (collectionScores.contains(address(_collection))) {
+      return collectionScores.get(address(_collection));
+    }
+    return 0;
```
ramenforbreakfast, line 157:
```suggestion
    return collectionScores.tryGet(address(_collection));
```
This can just be replaced by `tryGet`, no?

apps/smart-contracts/core/contracts/DepositHook.sol
```diff
@@ -65,4 +91,85 @@ contract DepositHook is IDepositHook, SafeAccessControlEnumerable {
...
+  }
+
+  function setCollectionScores(
+    IERC721[] memory _collections,
+    uint256[] memory _scores
+  ) external override onlyRole(SET_COLLECTION_SCORES_ROLE) {
+    require(
+      _collections.length == _scores.length,
+      "collections.length != scores.length"
+    );
```
ramenforbreakfast, line 124: So the document was edited slightly this morning, might have missed it, but we want to revert if score = 0, since we don't want to waste space adding a collection that is worth 0.

We revert, since if we were to add one with 0, we most likely screwed up the input and would want it to alert us.

https://www.notion.so/prepo/NFTAccessHook-SC-Mini-Spec-5bbf9f36b33f4005933272429b274740#c4651d6cdd6b4ef3b1a10144cd159cb2

apps/smart-contracts/core/test/DepositHook.test.ts
```diff
@@ -111,6 +154,39 @@ describe('=> DepositHook', () => {
 
       expect(mockDepositRecord.recordDeposit).to.be.calledWith(user.address, TEST_AMOUNT_AFTER_FEE)
     })
+
+    it('should revert if required score > 0 and user not on allow list', async () => {
```
ramenforbreakfast, line 158:
```suggestion
    it('should revert if required score > 0 and user not on allow list', async () => {
```

So multiple things:
- I know it might be confusing since some of these test suites use old naming conventions, but we don't use "should", can just say directly `reverts if...`

- This testcase naming is not accurate since what we really want to test is that if `requiredScore > 0` and user isn't included, then the user's score must pass the required threshold. So something like `reverts if user is not allowed and doesn't pass required score > 0`

- In addition to the naming, we should also add expects to verify that these conditions are met AND explicitly reflect the name of the testcase, for example in the testcase below, 

`expect(await depositHook.getRequireScore()).to.eq(1)` should be:
`expect(await depositHook.getRequiredScore()).to.be.gt(0)` since that is what the testcase explicitly says.

also, `expect(await depositHook.getAccountScore(user.address)).to.be.gte(await depositHook.getRequiredScore())` and `expect(await mockAllowlist.isIncluded(user.address)).to.eq(false)`

Since these are also conditions clearly stated in the testcase. 

My comments here pretty much apply to all the testcases for the hook modifications, so please apply them accordingly.

apps/smart-contracts/core/test/DepositHook.test.ts
```diff
@@ -257,4 +333,136 @@ describe('=> DepositHook', () => {
...
+      await expect(depositHook.connect(user).setCollectionScores([], [])).revertedWith(
+        `AccessControl: account ${user.address.toLowerCase()} is missing role ${await depositHook.SET_COLLECTION_SCORES_ROLE()}`
+      )
+    })
+
+    it('sets collection scores', async () => {
+      await depositHook.connect(deployer).setCollectionScores([testERC721.address], [1])
+
+      expect(await depositHook.getCollectionScore(testERC721.address)).to.eq(1)
+    })
```
ramenforbreakfast, line 399: You need to add testcases for if it can set multiple scores, this test block only checks for one element.

apps/smart-contracts/core/test/DepositHook.test.ts
```diff
@@ -111,6 +154,39 @@ describe('=> DepositHook', () => {
...
+    it('should revert if required score > 0 and user not on allow list', async () => {
+      await depositHook.connect(deployer).setRequiredScore(1)
+      expect(await depositHook.getRequiredScore()).to.eq(1)
+
+      await expect(
+        depositHook.connect(vault).hook(user.address, TEST_AMOUNT_BEFORE_FEE, TEST_AMOUNT_AFTER_FEE)
+      ).to.revertedWith('sender not allowed')
+    })
+
+    it('succeeds if required score > 0 and user on allow list', async () => {
```
ramenforbreakfast, line 167: This testcase should also expect that the user does not meet the score threshold.

There should also be a testcase to ensure that if the user is included and meets the threshold, that it does not revert.

apps/smart-contracts/core/test/DepositHook.test.ts
```diff
@@ -111,6 +154,39 @@ describe('=> DepositHook', () => {
...
+      expect(await depositHook.getRequiredScore()).to.eq(1)
+      await depositHook.connect(deployer).setCollectionScores([testERC721.address], [1])
+      expect(await depositHook.getCollectionScore(testERC721.address)).to.eq(1)
+
+      await testERC721.connect(deployer).mint(user.address)
+
+      await depositHook
+        .connect(vault)
+        .hook(user.address, TEST_AMOUNT_BEFORE_FEE, TEST_AMOUNT_AFTER_FEE)
+    })
```
ramenforbreakfast, line 189: You need to add testcases for if the user's score exceeds the collection score, not just meets it, I would also try to use a required score and collection score > 1, and not just rely on one collection, since we have no idea if the actual summation is working properly.

ramenforbreakfast (changes requested):
So the contract looks mostly fine to me besides some details missing that might have been missed from the document.

I think the tests need some work, particularly the hook logic testcases.

Also in the future, please do not submit all of these changes at once, I would try to split up the setters and getters (the more mundane logic), from the hook logic in a train, so that 

1. the easy stuff can be merged in first.

2. The reviewer doesn't have to also see stuff like test contracts and more simple tests along with the stuff that needs more attention.

apps/smart-contracts/core/contracts/DepositHook.sol
davidprepo: tryGet returns (bool success, uint value), so it could but the code would be similarly readable. Can change if it's a code style you'd prefer?

ramenforbreakfast: Ah i see, then I guess we'd have to isolate the `uint` in a separate line. We can keep it the way it is.

PR #564, opened 2022-11-24
feat: remove allowance from withdraw and fix approve tests

ramenforbreakfast:
Adds a line in `withdraw` to `approve` 0 after the `hook` is called.

So while writing the tests for this, I realized that I had not been `await`ing the expects for the `Approval` events, so actually some of the tests were false positives.

What was happening is that when we do the base token decimal specific tests, I had done the base setup for the new `Collateral` configured with the custom decimals base token specifically deployed for that test, but not reassigning the `withdrawHook`, so it was actually skipping the fee approval and those expects were false positives.

I have fixed this for the test cases where we change the base token decimals.


XavierEkkel (approved):
LGTM

PR #566, opened 2022-11-25
fix: add return value to deposit

ramenforbreakfast:
I have removed this from PR from the previous one it was included in and submitted it as a separate one since there are no potential conflicts that would necessitate keeping them together.


apps/smart-contracts/core/test/Collateral.test.ts
```diff
@@ -921,6 +921,20 @@ describe('=> Collateral', () => {
...
+      const amountToDeposit = await baseToken.balanceOf(sender.address)
+      expect(amountToDeposit).to.be.gt(0)
+      expect(await baseToken.allowance(sender.address, collateral.address)).to.be.eq(
+        amountToDeposit
+      )
+      const fee = amountToDeposit.mul(await collateral.getDepositFee()).div(FEE_DENOMINATOR)
+      const expectedCT = amountToDeposit.sub(fee).mul(parseEther('1')).div(USDC_DENOMINATOR)
+
+      expect(
+        await collateral.connect(sender).callStatic.deposit(recipient.address, amountToDeposit)
```
XavierEkkel, line 935: why callStatic?

XavierEkkel (approved):
LGTM after question

apps/smart-contracts/core/test/Collateral.test.ts
ramenforbreakfast: Because you cannot retrieve return values of a write transaction like you can with a view function.

PR #569, opened 2022-11-28
feat: change market fee logic to reflect simplified Collateral fee logic

ramenforbreakfast:
To make tests less verbose, I am operating under the assumption that the existing `setupMarketToEnd` fixture does indeed bring a market to its end by setting the `finalLongPayout`, only verifying that it has been set to a value less than MAX_PAYOUT. 

Added comments to explain custom long/short input values when custom ones are needed for edge cases.


apps/smart-contracts/core/contracts/PrePOMarket.sol
```diff
@@ -133,12 +133,12 @@ contract PrePOMarket is IPrePOMarket, Ownable, ReentrancyGuard {
...
-     * an amount large enough to pay a fee
-     */
-    uint256 _fee = (_collateralOwed * redemptionFee) / FEE_DENOMINATOR + 1;
-    require(_collateralOwed > _fee, "Redemption amount too small");
+    uint256 _fee = (_collateralOwed * redemptionFee) / FEE_DENOMINATOR;
+    if (redemptionFee > 0) {
+      require(_fee > 0, "fee = 0");
+    } else {
+      require(_collateralOwed > 0, "amount = 0");
+    }
```
XavierEkkel, line 141: we should write it like we write the one in `withdraw()`
i.e.:
-move these fee calculations + checks above the `burnFrom()` calls (so that as many checks as possible are done before any state-changing calls)
-rename `_collateralOwed` to just `_collateralAmount` (it being owed is implied since there is no other collateral amount to confuse it with)
-have a separate `_collateralAmountAfterFee` variable (rather than reassigning the same variable)
-(in a separate PR) have `_fee` as the 3rd argument of the `Redemption` event

ramenforbreakfast: Ok will finish addressing in 10 mins, need to re-test.

apps/smart-contracts/core/test/PrePOMarket.test.ts
```diff
@@ -465,6 +465,98 @@ describe('=> prePOMarket', () => {
       }
     })
 
+    it('reverts if redemption = 0, redemption fee = 0%, and redeeming before market end', async () => {
```
XavierEkkel, line 468: nit:
```suggestion
    it('reverts if redemption = 0, fee = 0%, and before market end', async () => {
```
'redemption'/'redeeming' is implied, same for tests below

apps/smart-contracts/core/test/PrePOMarket.test.ts
```diff
@@ -465,6 +465,98 @@ describe('=> prePOMarket', () => {
...
+      /**
+       * Given a test fee of 20 (0.002%), smallest redemption that would result in a
+       * fee(of 1) would be 50000 wei, so for fee = 0, redeem 49999.
+       */
+      const longToRedeem = BigNumber.from(49999)
+      const shortToRedeem = longToRedeem
+      expect(await longToken.balanceOf(user.address)).to.be.gte(longToRedeem)
+      expect(await shortToken.balanceOf(user.address)).to.be.gte(shortToRedeem)
+      // expect fee to be zero
+      const totalOwed = await calculateTotalOwed(longToRedeem, shortToRedeem, false)
```
XavierEkkel, line 510: also expect total owed to be > 0, same for similar below tests

apps/smart-contracts/core/test/PrePOMarket.test.ts
XavierEkkel: also change 'redemption = 0' to 'amounts = 0', similarly for below tets

apps/smart-contracts/core/test/PrePOMarket.test.ts
```diff
@@ -556,24 +648,69 @@ describe('=> prePOMarket', () => {
       expect(await collateralToken.balanceOf(user.address)).to.eq(totalOwed.sub(redeemFee))
     })
 
-    it('should not allow redemption amounts too small for a fee before market end', async () => {
+    it('allows a redemption > 0 if redemption fee = 0% and redeeming equal parts', async () => {
```
XavierEkkel, line 651:
```suggestion
    it('allows amounts > 0 if redemption fee = 0% and redeeming equal parts', async () => {
```
nit: no `a`
same for below

XavierEkkel (approved):
LGTM after comments!

PR #577, opened 2022-11-28
feat: adding nft access logic to deposithook

davidprepo:


ramenforbreakfast:
@davidprepo going to close this since its just confusing being here and out of date with the rest of the train.
Please submit this as a subsequent PR after #584

PR #578, opened 2022-11-28
feat: ITokenSender and IUintValue

davidprepo:


XavierEkkel (approved):
LGTM if remove all natspec and remove underscores from argument variables (do this for PRs going forward)

apps/smart-contracts/core/contracts/interfaces/ITokenSender.sol
```diff
@@ -0,0 +1,82 @@
...
+import "./IAccountList.sol";
+import "prepo-shared-contracts/contracts/interfaces/IUintValue.sol";
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
+/**
+ * @title ITokenSender
+ * @dev A generic TokenSender contract that receives an amount of a pre-conversion asset,
+ * and sends a converted amount of post-conversion asset to a recipient.
+ */
+interface ITokenSender {
```
XavierEkkel, line 13: run linter on this PR so that linting isn't in the next PR in the train

apps/smart-contracts/core/contracts/interfaces/ITokenSender.sol
```diff
@@ -0,0 +1,82 @@
...
+  /**
+   * @dev Emitted when the price is changed.
+   */
+  event PriceChange(IUintValue price);
+
+  /**
+   * @dev Emitted when the prive multiplier is changed.
+   */
+  event PriceMultiplierChange(uint256 priceMultiplier);
```
ramenforbreakfast, line 22:
```suggestion
  event PriceMultiplierChange(uint256 multiplier);
```
Lets be consistent with other event naming conventions where we do not make the entire variable name the event parameter.

apps/smart-contracts/core/contracts/interfaces/ITokenSender.sol
```diff
@@ -0,0 +1,82 @@
...
+
+  /**
+   * @dev Emitted when the scaled price lower bound is changed.
+   */
+  event ScaledPriceLowerBoundChange(uint256 scaledPriceLowerBound);
```
ramenforbreakfast, line 27:
```suggestion
  event ScaledPriceLowerBoundChange(uint256 scaledPrice);
```
I think `scaledPrice`, rather than `lowerBound` works better here, because the value itself is a scaled price, that is used to act as a lower bound. Similar to the above `priceMultiplier`, using `multiplier` since it is a multiplier being applied to a price.

ramenforbreakfast (approved):
LGTM besides nits

apps/smart-contracts/core/contracts/interfaces/ITokenSender.sol
```diff
@@ -0,0 +1,82 @@
...
+
+  /**
+   * @notice Sends tokens an account.
+   *
+   * @param _recipient is the reciever address
+   * @param _unconvertedAmount is the pre-conversion amount to send, e.g. the baseToken or Collateral fee extracted from deposit/withdraw/redeem()
+   */
+  function send(address _recipient, uint256 _unconvertedAmount) external;
```
ramenforbreakfast, line 35:
```suggestion
  function send(address recipient, uint256 unconvertedAmount) external;
```

Lets use the new convention since this is a new contract, same for the rest of this contract and `IUintValue.sol` (remember to rebase the train so you don't accidentally overwrite these changes).

PR #582, opened 2022-11-28
feat: generic slide up & disabled state, slide up positions

georgesmithishere:
https://www.notion.so/prepo/FE-Project-Management-d3878c29e9534234a50f4f8838525bee?p=88b084cd442f41489530bc0f741eb82a&pm=s

### DIFF
--- a/apps/frontend/core-dapp/src/features/portfolio/PortfolioStore.ts
+++ b/apps/frontend/core-dapp/src/features/portfolio/PortfolioStore.ts
@@ -51,8 +51,8 @@ export class PortfolioStore {
     const position = this.positions.find(
       (item) => item.market.urlId === marketId && direction === item.position
     )
-    if (position) {
-      this.setSelectedPosition(position)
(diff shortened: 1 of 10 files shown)


chrisling-dev (changes requested):
Could we make SlideUp component even more generic? We will be using it for few other things like trade page settings, simulator settings and etc so it should be able to take something else other than a list. Also anything component should be able to trigger the SlideUp and not just button. What I'm thinking is:

SlideUp component with `content` prop (ReactNode) that takes anything, children is the trigger (e.g. the button)

PositionsList -> With all the
```
 // ... selected position
 // ... list of other positions
```

MarketsList -> With all the
```
 // ... selected market
 // ... list of other markets
```

Then we would just use:
<SlideUp content={<MarketsList /> or <PositionsList />}>
  // ... Button or whatever
</SlideUp>

chrisling-dev:
Actually we could just use the SlideUpCard component, I think that would be generic enough

PR #587, opened 2022-11-29
feat: add fee to Redemption event

ramenforbreakfast:
Still using a filter for this event since we want to verify that the indexed parameters are searchable.


XavierEkkel (approved):
LGTM after comments

PR #589, opened 2022-11-29
refactor: make some more obvious changes to speed up Collateral tests

ramenforbreakfast:
I did not want to do this refactoring earlier when we were writing tests for `deposit` and `withdraw` since it would have required splitting out setup into multiple specific fixtures and other refactoring.

However, Collateral tests are now taking 16 minutes on my own machine, and on GitHub Actions i guarantee they are taking up the majority of the PR testing time.

These changes split up the singular monolith setup fixture functions to ensure only the specific components needed for each test are setup (with reason).

In addition, the `deposit` and `withdraw` tests use conditional setups that simplify them greatly and also reduce the number of times stacks are setup twice (we were setting them up twice before when we had to deploy one with a different baseToken).

These changes reduce the run time of Collateral tests by ~half, going down to 9 minutes on my machine.


XavierEkkel (approved):
LGTM!

PR #590, opened 2022-11-29
feat: remove redemption fee from PrePOMarket constructor

ramenforbreakfast:
I also took the liberty of correcting the `setRedemptionFee` test block since it was small and we are already "operating" on it.

In `PrePOMarket` had to add in setting the redemption fee since it was implied that it would start off non-zero since it was a constructor argument. Some of the code in that suite being changed is not ideal, but I don't think refactoring at this point is worth the effort unless it becomes necessary.


ramenforbreakfast:
Closing, will be opening PR off Chris's branch to test if when opened, email alerts will still show a false failure.

PR #592, opened 2022-11-29
feat: remove redemption fee from PrePOMarket constructor

ramenforbreakfast:
I also took the liberty of correcting the setRedemptionFee test block since it was small and we are already "operating" on it.

In PrePOMarket had to add in setting the redemption fee since it was implied that it would start off non-zero since it was a constructor argument. Some of the code in that suite being changed is not ideal, but I don't think refactoring at this point is worth the effort unless it becomes necessary.

This was based off Chris's false positive deploy action to verify that an SC run job would not trigger a false positive email. Will be rebased later, once we can confirm it does work.


apps/smart-contracts/core/test/PrePOMarket.test.ts
```diff
@@ -287,35 +263,46 @@ describe('=> prePOMarket', () => {
...
 
-    it('should be settable to zero', async () => {
+    it('sets to zero', async () => {
+      await prePOMarket.connect(treasury).setRedemptionFee(FEE_LIMIT)
+      expect(await prePOMarket.getRedemptionFee()).to.not.eq(0)
+
       await prePOMarket.connect(treasury).setRedemptionFee(0)
+
       expect(await prePOMarket.getRedemptionFee()).to.eq(0)
     })
```
XavierEkkel, line 302: add idempotency test

XavierEkkel (approved):
LGTM after comment

PR #584, opened 2022-11-28
feat: NFTAccess: logic and tests for removeCollections

davidprepo:
Parent: https://github.com/prepo-io/prepo-monorepo/pull/583

PR #583, opened 2022-11-28
feat: logic and test for setCollectionScores

davidprepo:

PR #576, opened 2022-11-28
feat: implementing INFTAccessHook and required score getter/setter

davidprepo:
https://github.com/prepo-io/prepo-monorepo/pull/583
https://github.com/prepo-io/prepo-monorepo/pull/584

PR #593, opened 2022-11-29
feat: nft access logic to DepositHook

davidprepo:
https://github.com/prepo-io/prepo-monorepo/pull/576
https://github.com/prepo-io/prepo-monorepo/pull/583
https://github.com/prepo-io/prepo-monorepo/pull/584


apps/smart-contracts/core/contracts/DepositHook.sol
```diff
@@ -158,9 +160,18 @@ contract DepositHook is
...
+  {
+    uint256 score = 0;
+    for (uint256 i = 0; i < collectionToScore.length(); i++) {
+      (address collection, uint256 collectionScore) = collectionToScore.at(i);
+      score += IERC721(collection).balanceOf(_account) > 0
+        ? collectionScore
+        : 0;
+    }
+    return score;
+  }
```
ramenforbreakfast, line 176: Change `score` to `_score`. For files that we are editing, lets keep the old convention because it will get confusing. Only go with the new way for files that are completely new being added.

apps/smart-contracts/core/test/DepositHook.test.ts
```diff
@@ -138,40 +175,103 @@ describe('=> DepositHook', () => {
       ).to.revertedWith('msg.sender != collateral')
     })
 
-    it('reverts if user not allow list and deposits not allowed', async () => {
-      mockAllowlist.isIncluded.returns(false)
-      expect(await mockAllowlist.isIncluded(user.address)).to.be.false
-      await depositHook.connect(deployer).setDepositsAllowed(false)
-      expect(await depositHook.depositsAllowed()).to.eq(false)
+    it('reverts if deposits not allowed, not on allowlist, score < required score', async () => {
+      await setupDepositPermissionForTest(false, false, 0, 1)
```
ramenforbreakfast, line 179: Nice!

apps/smart-contracts/core/test/DepositHook.test.ts
```diff
@@ -138,40 +175,103 @@ describe('=> DepositHook', () => {
...
         depositHook.connect(vault).hook(user.address, TEST_AMOUNT_BEFORE_FEE, TEST_AMOUNT_AFTER_FEE)
       ).to.revertedWith('deposits not allowed')
     })
 
-    it('reverts if user on allow list but deposits not allowed', async () => {
-      mockAllowlist.isIncluded.returns(true)
-      expect(await mockAllowlist.isIncluded(user.address)).to.be.true
-      await depositHook.connect(deployer).setDepositsAllowed(false)
-      expect(await depositHook.depositsAllowed()).to.eq(false)
+    it('reverts if deposits not allowed, on allowlist, score < required score', async () => {
```
ramenforbreakfast, line 186: Since testing every possible combination within here is going to be quite cumbersome (~20 testcases if we did), the team has decided to make some compromises to reduce the verbosity, should only be 6? testcases.

- No need to add additional deposits allowed revert tests, just keep the one that already exists.
- Move `requiredScore` check to the front, and then just check that `allowlist` is not called when `requiredScore` = 0. So no need to check any combinations other than that with `requiredScore` = 0.

apps/smart-contracts/core/contracts/DepositHook.sol
```diff
@@ -51,7 +51,9 @@ contract DepositHook is
     uint256 _amountAfterFee
   ) external override onlyCollateral {
     require(depositsAllowed, "deposits not allowed");
-    require(allowlist.isIncluded(_sender), "sender not allowed");
+    if (!allowlist.isIncluded(_sender) && requiredScore > 0) {
```
ramenforbreakfast, line 54:
```suggestion
    if (requiredScore > 0 && !allowlist.isIncluded(_sender)) {
```
As per discussion below, move `requiredScore` to the front.

ramenforbreakfast (changes requested):

apps/smart-contracts/core/contracts/DepositHook.sol
```diff
@@ -51,7 +51,9 @@ contract DepositHook is
...
+      require(getAccountScore(_sender) >= requiredScore, "sender not allowed");
+    }
```
ramenforbreakfast, line 56: Ignore my earlier comment about moving `requiredScore` to the front, since it is an `&&` condition, it will evaluate both every time.

ramenforbreakfast:
David and I have proposed utilizing preset arrays of test inputs, and then iterating them through a for loop similar to what we did with ACQR tests. Something like combinations of a tuple of (`shouldRevert`, `onAllowlist`, `accountScore`, `requiredScore`), `shouldRevert` is required since we need to dynamically append whether something will be a success or not.

PR #597, opened 2022-11-29
feat: fee rebate hook skeleton

davidprepo:

PR #598, opened 2022-11-29
feat: FeeRebate: treasury set/get for fee rebate hook

davidprepo:
https://www.notion.so/prepo/Fee-Rebate-SC-Spec-5723bb91a17c42e2a409ffd9af591085#e421f990845d4b9792f587392fc2546f

Parent PR: https://github.com/prepo-io/prepo-monorepo/pull/597


apps/smart-contracts/core/contracts/FeeRebateHook.sol
```diff
@@ -6,16 +6,21 @@ import "./interfaces/ITokenSender.sol";
 import "prepo-shared-contracts/contracts/SafeAccessControlEnumerable.sol";
 
 contract FeeRebateHook is IFeeRebateHook, SafeAccessControlEnumerable {
-  address private treasury;
+  address public treasury;
```
ramenforbreakfast, line 9:
```suggestion
  address private _treasury;
```
Why is this public? Also same with other PRs, lets follow the new convention for the rest of this contract.

apps/smart-contracts/core/contracts/FeeRebateHook.sol
```diff
@@ -6,16 +6,21 @@ import "./interfaces/ITokenSender.sol";
...
-    address /*treasury*/
-  ) external override onlyRole(SET_TREASURY_ROLE) {}
+  function setTreasury(address _newTreasury)
+    external
+    override
+    onlyRole(SET_TREASURY_ROLE)
+  {
+    treasury = _newTreasury;
+    emit TreasuryChange(_newTreasury);
+  }
```
ramenforbreakfast, line 23:
```suggestion
  function setTreasury(address treasury)
    external
    override
    onlyRole(SET_TREASURY_ROLE)
  {
    _treasury = treasury;
    emit TreasuryChange(treasury);
  }
```

apps/smart-contracts/core/test/FeeRebateHook.test.ts
```diff
@@ -32,6 +33,10 @@ describe('=> FeeRebateHook', () => {
   })
 
   describe('initial state', () => {
+    it('does not have treasury set', async () => {
```
ramenforbreakfast, line 36:
```suggestion
    it('does not set treasury', async () => {
```

apps/smart-contracts/core/test/FeeRebateHook.test.ts
```diff
@@ -32,6 +33,10 @@ describe('=> FeeRebateHook', () => {
...
+      expect(await feeRebateHook.treasury()).to.equal(ethers.constants.AddressZero)
```
ramenforbreakfast, line 37:
```suggestion
      expect(await feeRebateHook.treasury()).to.eq(ZERO_ADDRESS)
```
Not a big deal, but we already have ZERO_ADDRESS, and please be consistent with using `eq` instead of `equal`

ramenforbreakfast (approved):
LGTM, besides a few minor changes, and not sure why `treasury` is not `private`

apps/smart-contracts/core/contracts/FeeRebateHook.sol
davidprepo: This is public to as no getter was described in the SC documentation. The options are:

- Keep as public
- Implement getter function
- Don't test the setter
- Only test that this value is set correctly implicitly when testing behavior somehow

I opted for number 1 as it was least invasive and most complete.

ramenforbreakfast: Oh that might be because in the doc, theres a section that says if there is a `private` variable, and nothing is said about it, it is implied that there is a getter.

apps/smart-contracts/core/contracts/FeeRebateHook.sol
```diff
@@ -6,19 +6,26 @@ import "./interfaces/ITokenSender.sol";
...
 contract FeeRebateHook is IFeeRebateHook, SafeAccessControlEnumerable {
-  address private _treasury;
+  address public treasury;
   ITokenSender private _feeRebate;
-  
+
   bytes32 public constant SET_TREASURY_ROLE =
     keccak256("FeeRebateHook_setTreasury(address)");
   bytes32 public constant SET_TOKEN_SENDER_ROLE =
     keccak256("FeeRebateHook_setTokenSender(ITokenSender)");
```
ramenforbreakfast, line 15: Should have caught this earlier, but there should be no access controls enforced in this contract, it is a module that should be inherited. The document is not accurate and needs to be updated (will do this).

_After_ all the `FeeRebateHook` PRs are merged, you can move it to `prepo-shared-contracts`.

apps/smart-contracts/core/contracts/FeeRebateHook.sol
```diff
@@ -6,19 +6,26 @@ import "./interfaces/ITokenSender.sol";
...
 
-  function setTreasury(
-    address treasury
-  ) external override onlyRole(SET_TREASURY_ROLE) {}
+  function setTreasury(address newTreasury)
+    external
+    override
+    onlyRole(SET_TREASURY_ROLE)
```
ramenforbreakfast, line 20:
```suggestion
    external
    virtual
    override
```
Remove access control and make it virtual so it can be override

apps/smart-contracts/core/contracts/FeeRebateHook.sol
```diff
@@ -6,19 +6,26 @@ import "./interfaces/ITokenSender.sol";
...
+    emit TreasuryChange(newTreasury);
+  }
 
-  function setTokenSender(
-    ITokenSender tokenSender
-  ) external override onlyRole(SET_TOKEN_SENDER_ROLE) {}
+  function setTokenSender(ITokenSender tokenSender)
+    external
+    override
+    onlyRole(SET_TOKEN_SENDER_ROLE)
```
ramenforbreakfast, line 29:
```suggestion
    external
    virtual
    override
```
same here.

apps/smart-contracts/core/contracts/FeeRebateHook.sol
```diff
@@ -6,19 +6,26 @@ import "./interfaces/ITokenSender.sol";
...
-  function setTreasury(
-    address treasury
-  ) external override onlyRole(SET_TREASURY_ROLE) {}
+  function setTreasury(address newTreasury)
+    external
+    override
+    onlyRole(SET_TREASURY_ROLE)
+  {
+    treasury = newTreasury;
+    emit TreasuryChange(newTreasury);
```
ramenforbreakfast, line 23: Move this to an `internal` method `_setTreasury(address treasury)`, then have this function just call `_setTreasury`.

apps/smart-contracts/core/contracts/FeeRebateHook.sol
```diff
@@ -6,19 +6,26 @@ import "./interfaces/ITokenSender.sol";
...
+  }
 
-  function setTokenSender(
-    ITokenSender tokenSender
-  ) external override onlyRole(SET_TOKEN_SENDER_ROLE) {}
+  function setTokenSender(ITokenSender tokenSender)
+    external
+    override
+    onlyRole(SET_TOKEN_SENDER_ROLE)
+  {}
```
ramenforbreakfast, line 30: Do the same for `setTokenSender` in the following PR. For the future, if you are just submitting simple setters, you can include them in one PR, rather than splitting them.

ramenforbreakfast (approved):
LGTM, but please make the this contract follow the same format as `AllowlistHook` and `AllowedCallers` since this is also an inheritable module.

apps/smart-contracts/core/contracts/FeeRebateHook.sol
ramenforbreakfast: Ok actually, since this is a variable that needs to be accessed by inheriting contracts, make this `internal`, I made this mistake for `AllowlistHook` and will submit a PR correcting this. I have changed this in the doc as well. Still add a getter for it though.

apps/smart-contracts/core/contracts/FeeRebateHook.sol
```diff
@@ -6,19 +6,26 @@ import "./interfaces/ITokenSender.sol";
 import "prepo-shared-contracts/contracts/SafeAccessControlEnumerable.sol";
 
 contract FeeRebateHook is IFeeRebateHook, SafeAccessControlEnumerable {
-  address private _treasury;
+  address public treasury;
   ITokenSender private _feeRebate;
```
ramenforbreakfast, line 10:
```suggestion
  ITokenSender internal _feeRebate;
```
So the inheriting contract can access this.

davidprepo:
All comments are fixed, removed access control (that was the primary reason these setters/getters were split into multiple PRs).

PR #600, opened 2022-11-30
refactor: simplify PrePOMarket params

ramenforbreakfast:
For `DepositTradeHelper` testing, we are going to need to add a special `PrePOMarket` fixture for smocked testing. In preparation, so we do not have to repeat these giant list of parameters everywhere, I created TS types to share common variables between fixtures for deploying `PrePOMarket` by itself (unconventional, but necessary for some tests including DepositTradeHelper) and via `PrePOMarketFactory`.


XavierEkkel (approved):
LGTM

PR #601, opened 2022-11-30
feat: add smock fixtures needed for DepositTradeHelper testing

ramenforbreakfast:
For `DepositTradeHelper` testing, to simplify things, we will need to smock almost every component since we would need to setup a test to deploy a Uniswap pool and actually supply liquidity and simulate swap amounts. This PR adds smock fixtures for those contracts that we did not already have smock fixtures for.


XavierEkkel (approved):
LGTM

PR #602, opened 2022-11-30
feat: add custom smock PrePOMarket creation fixture

ramenforbreakfast:
To deploy markets for `DepositTradeHelper` testing, we need a custom fixture to setup a smock `PrePOMarket` in a similar way to `PrePOMarketFactory`, since in production we would deploy via `PrePOMarketFactory`, but this won't work if we want to smock it and verify it being called externally by `DepositTradeHelper`.


XavierEkkel (approved):
LGTM

PR #606, opened 2022-11-30
feat: adding nft score logic to deposit hook

davidprepo:


apps/smart-contracts/core/test/DepositHook.test.ts
```diff
@@ -138,9 +165,7 @@ describe('=> DepositHook', () => {
       ).to.revertedWith('msg.sender != collateral')
     })
 
-    it('reverts if user not allow list and deposits not allowed', async () => {
-      mockAllowlist.isIncluded.returns(false)
-      expect(await mockAllowlist.isIncluded(user.address)).to.be.false
+    it('reverts if deposits not allowed', async () => {
```
davidprepo, line 168: One option here which might help with test coverage but also readability of code would be to include multiple tests in this one it() block - ie test that the call revers when deposits are not allowed if not on allowlist, on allowlist, passes nft score, fails nft score. Doesn't fit with the best practices, but given that we're sidestepping best practices a little for the sake of brevity, it might be an acceptable compromise?

apps/smart-contracts/core/test/DepositHook.test.ts
```diff
@@ -149,39 +174,52 @@ describe('=> DepositHook', () => {
...
+        .connect(vault)
+        .hook(user.address, TEST_AMOUNT_BEFORE_FEE, TEST_AMOUNT_AFTER_FEE)
     })
 
-    it('reverts if account not on allow list', async () => {
-      mockAllowlist.isIncluded.returns(false)
-      expect(await mockAllowlist.isIncluded(user.address)).to.be.false
-      await depositHook.connect(deployer).setDepositsAllowed(true)
-      expect(await depositHook.depositsAllowed()).to.eq(true)
+    it('succeeds if account not on allowlist, and required score = 0', async () => {
```
XavierEkkel, line 186:
```suggestion
    it('succeeds if required score = 0', async () => {
```
don't need to say 'account not on allowlist' for this test (or any of the tests below), as that is already implied from us already testing the 'succeeds if account on allowlist' case

similarly, don't need to say 'required score > 0' for the tests below

apps/smart-contracts/core/test/DepositHook.test.ts
XavierEkkel: I think the tests we have in this PR are adequate

apps/smart-contracts/core/test/DepositHook.test.ts
```diff
@@ -130,6 +130,33 @@ describe('=> DepositHook', () => {
...
+      }
+
+      // Verify account score and required score
+      if (accountScore > requiredScore) {
+        expect(await depositHook.getAccountScore(user.address)).to.be.gt(requiredScore)
+      } else if (accountScore === requiredScore) {
+        expect(await depositHook.getAccountScore(user.address)).to.eq(requiredScore)
+      } else if (accountScore < requiredScore) {
+        expect(await depositHook.getAccountScore(user.address)).to.be.lt(requiredScore)
+      }
```
XavierEkkel, line 157: i think it will slow things down doing assertions in setup functions that get called multiple times. can remove

XavierEkkel (approved):
LGTM after comments

ramenforbreakfast (approved):
Have addressed some changes, like maintaining consistency of having inherited overriden functions and roles before ones exclusive to the hook.

In addition, there was a bug in tests that needed fixing, the tests did not have an allowlist setup so the tests would fail attempting to call it.

PR #611, opened 2022-11-30
feat: NFTAccess: implementation of AllowlistHook

davidprepo:
https://www.notion.so/prepo/NFTAccessHook-SC-Mini-Spec-5bbf9f36b33f4005933272429b274740#f5f925b0fb7d42e4a67282952fb84809

This is the head PR for the NFT access list feature.

Implemented as one PR, as per previous comments of skeleton implementations being too skinny.


ramenforbreakfast (approved):
So I made quite a few changes here since the submission as it was would not have worked.

- Defined an `internal` `_setAllowlist` that an inheriting contract would call.
- Moved out `SafeAccessControlEnumerable` from `AllowlistHook` since we should allow the inheriting contract to define access control.
- Added the role to `DepositHook` and setup the overriding `external` function, verifying that it reverts if not role holder.

I acknowledge that this could have been moved to `prepo-shared-contracts`, but this will have to happen in a later PR, because it uses `IAccountList` which was regrettably not already moved to `prepo-shared-contracts`, so it would require modifying a bunch of contract imports.

apps/smart-contracts/core/test/AllowlistHook.test.ts
```diff
@@ -0,0 +1,67 @@
...
+  let user: SignerWithAddress
+  let allowlist: SignerWithAddress
+
+  beforeEach(async () => {
+    ;[deployer, user, allowlist] = await ethers.getSigners()
+    allowlistHook = await allowlistHookFixture()
+  })
+
+  describe('# initialize', () => {
+    it('should be initialized with no allowlist set', async () => {
```
XavierEkkel, line 21:
```suggestion
    it('initializes allowlist to zero address', async () => {
```

apps/smart-contracts/core/test/AllowlistHook.test.ts
```diff
@@ -0,0 +1,67 @@
...
+      expect(await allowlistHook.getAllowlist()).to.eq(ethers.constants.AddressZero)
```
XavierEkkel, line 22: are we phasing out the `ZERO_ADDRESS` constant? if using this form, can we make it more concise (i.e. without the `ethers.constants.` each time)?

apps/smart-contracts/core/test/AllowlistHook.test.ts
```diff
@@ -0,0 +1,67 @@
...
+    it('is idempotent for non-zero address', async () => {
+      await allowlistHook.setAllowlist(allowlist.address)
+      expect(await allowlistHook.getAllowlist()).to.eq(allowlist.address)
+
+      await allowlistHook.setAllowlist(allowlist.address)
+
+      expect(await allowlistHook.getAllowlist()).to.eq(allowlist.address)
+    })
+
+    it('is idempotent for zero address', async () => {
```
XavierEkkel, line 53:
```suggestion
    it('is idempotent if zero address', async () => {
```

apps/smart-contracts/core/test/AllowlistHook.test.ts
```diff
@@ -0,0 +1,67 @@
...
+    it('sets to zero address', async () => {
+      await allowlistHook.setAllowlist(allowlist.address)
+      expect(await allowlistHook.getAllowlist()).to.eq(allowlist.address)
+
+      await allowlistHook.setAllowlist(ethers.constants.AddressZero)
+
+      expect(await allowlistHook.getAllowlist()).to.eq(ethers.constants.AddressZero)
+    })
+
+    it('is idempotent for non-zero address', async () => {
```
XavierEkkel, line 44:
```suggestion
    it('is idempotent if non-zero address', async () => {
```

XavierEkkel (approved):
LGTM after comments

apps/smart-contracts/core/test/AllowlistHook.test.ts
ramenforbreakfast: No, we are not, I missed this, I think i commented on some other PRs that we should stick to ZERO_ADDRESS

PR #615, opened 2022-12-01
fix: make _allowlist internal for inheriting contracts to access

ramenforbreakfast:
Need to make this `internal` so inheriting contract can access it. `AllowlistMarketHook` will need to revert inside `checkInclusion`, `require(_allowlist.isIncluded(sender))`.

I have also commented on PR #598 for @davidprepo to make `FeeRebateHook`'s variables `internal`


XavierEkkel (approved):
LGTM

PR #621, opened 2022-12-01
feat:core: remove treasury from PrePOMarket

ramenforbreakfast:
You'll notice that this PR just sets the `_collateralAmountAfterFee` to `_collateralAmount`, even though  the line will be changed. This was done to mimic what will happen once we setup the hook logic, since it will not transfer anything to the hook anyway, in a way that blends in with the next PR.

All of the old testcases have their treasury fees removed, and I will be adding on with new testcases, rather than modifying existing ones to accomodate the approval/hook logic, since trying to refactor these testcases would be too tedious and more content that would need review (our old testcases test many things together, rather than very specific things).


XavierEkkel (approved):
LGTM

PR #623, opened 2022-12-02
feat:core: add hook call to PrePOMarket mint

ramenforbreakfast:


apps/smart-contracts/core/test/PrePOMarket.test.ts
```diff
@@ -413,8 +416,19 @@ describe('=> prePOMarket', () => {
...
+    let mintHook: MockContract<Contract>
+    let allowlist: MockContract<Contract>
+    beforeEach(async () => {
       prePOMarket = await prePOMarketAttachFixture(await createMarket(defaultParams))
+      mintHook = await smockAllowlistMarketHookFixture()
+      allowlist = await smockAccountListFixture()
+      await mintHook.setAllowedCallers([prePOMarket.address], [true])
+      await mintHook.setAllowlist(allowlist.address)
+      await allowlist.set([user.address], [true])
+      await prePOMarket.connect(treasury).setMintHook(mintHook.address)
```
XavierEkkel, line 428: these only need to be in the hook-related tests. for every other test it will slow it down
move to a function instead and only call when needed?

XavierEkkel (approved):
LGTM after comment

PR #628, opened 2022-12-02
feat: FeeRebate: Add fee logic to DepositHook

davidprepo:
- Added getSignerForContract to utils as mentioned in discord
- Renamed "vault" to "collateral" as discussed with Michael to better reflect contract
- Renamed "outputToken" to "testToken" as this is now used for fees in hook tests
- New setup in beforeEach of hook is minimum required setup to allow collateral to work

Docs: https://www.notion.so/prepo/Fee-Rebate-SC-Spec-5723bb91a17c42e2a409ffd9af591085#ffb3f71d61074f4e8ed17a0f864e1a7f

Parent PR: https://github.com/prepo-io/prepo-monorepo/pull/627

PR #629, opened 2022-12-02
feat: FeeRebate: Add fee reimbursement logic to WithdrawHook

davidprepo:
- This follows all the same comments as the sibling implementation for DepositHook (https://github.com/prepo-io/prepo-monorepo/pull/628)
- Treasury and TokenSender setters are added in this PR per previous comments about simplicity
- Needed to add `tokenSender.send.returns()` in the beforeEach for hook() or tokenSender.send() would revert
- Renamed `vault` to `collateral`

Parent PR: https://github.com/prepo-io/prepo-monorepo/pull/628


ramenforbreakfast (approved):
@davidprepo I have corrected this to also put inherited roles/contracts first (should be in the order of inheritance). Also `SET_TOKEN_SENDER` should have been `SET_TOKEN_SENDER_ROLE`.

In addition, I think you forgot to fix the test suites for both `Collateral` and `WithdrawHook` so I went ahead and made those changes.

Besides needing to actually configure `depositHook` and `withdrawHook` with the new variables, our old tests assumed that the fee would not be taken, so we need to modify these to assume otherwise.

PR #630, opened 2022-12-03
feat:core: make AllowedCallers and AllowlistHook utilize public visiblity

ramenforbreakfast:
Not going to change names of `Hook` suffix contracts because there would have to be many changes and don't want to over complicate this PR so we can get `PrePOMarket` stuff merged in.

Also spotted in `TokenSender`, `setAllowedCallers` should be above functions defined within, since its an inherited one.


apps/smart-contracts/core/contracts/TokenSender.sol
```diff
@@ -52,6 +52,14 @@ contract TokenSender is
     _outputToken.transfer(recipient, outputAmount);
   }
 
+  function setAllowedCallers(address[] memory callers, bool[] memory allowed)
```
XavierEkkel, line 55: is there a reason we want to put the inherited functions first? I think it'd be clearer if the new functions that this contract introduces are shown first, followed by any inherited ones that are overridden
(plus, even if we did it your way, this function would belong above send)

XavierEkkel (approved):
LGTM after comment

PR #632, opened 2022-12-03
feat: add hook logic to MintHook

ramenforbreakfast:


XavierEkkel (commented):
.

apps/smart-contracts/core/contracts/MintHook.sol
```diff
@@ -11,7 +11,9 @@ contract MintHook is IMarketHook, AllowlistHook, AllowedCallers, SafeOwnable {
     address sender,
     uint256 amountBeforeFee,
     uint256 amountAfterFee
-  ) external virtual override onlyAllowedCallers {}
+  ) external virtual override onlyAllowedCallers {
+    require(_allowlist.isIncluded(sender), "sender not allowed");
```
XavierEkkel, line 15:
```suggestion
    require(_allowlist.isIncluded(sender), "minter not allowed");
```
to distinguish from allowed callers (which are PrePOMarket contracts) and to be clearer

apps/smart-contracts/core/contracts/MintHook.sol
```diff
@@ -11,7 +11,9 @@ contract MintHook is IMarketHook, AllowlistHook, AllowedCallers, SafeOwnable {
     address sender,
     uint256 amountBeforeFee,
     uint256 amountAfterFee
-  ) external virtual override onlyAllowedCallers {}
+  ) external virtual override onlyAllowedCallers {
```
XavierEkkel, line 14: since the allowed callers list will be the same list as in redeem hook, this would be better as an external allowlist shared between both `MintHook` and `RedeemHook`
is it possible to inherit `AllowlistHook` twice to use 2 allowlists?

XavierEkkel, line 14: as discussed, can change this in a future PR

XavierEkkel (approved):
LGTM after nit

PR #634, opened 2022-12-03
feat:core: add RedeemHook logic

ramenforbreakfast:
So these tests are similar to David's where since `FeeRebates` require actually using `Collateral` and `BaseToken` for transferring tokens, and the caller is restricted to that same `Collateral` contract being used in the logic, we require David's `getSignerForCollateral` workaround.

I chose `FakeContract<Contract>` for `Collateral` to ensure no `Collateral` logic is actually simulated, which would make setup much more complicated. Since there was no existing `FakeContract` fixture for `Collateral`, I have added one.


apps/smart-contracts/core/contracts/RedeemHook.sol
```diff
@@ -18,7 +19,15 @@ contract RedeemHook is
     address sender,
     uint256 amountBeforeFee,
     uint256 amountAfterFee
-  ) external virtual override onlyAllowedCallers {}
+  ) external virtual override onlyAllowedCallers {
+    require(_allowlist.isIncluded(sender), "sender not allowed");
```
XavierEkkel, line 23: nit:
```suggestion
    require(_allowlist.isIncluded(sender), "redeemer not allowed");
```

apps/smart-contracts/core/test/RedeemHook.test.ts
```diff
@@ -102,4 +111,75 @@ describe('=> RedeemHook', () => {
       await redeemHook.connect(deployer).setTokenSender(tokenSender.address)
     })
   })
+
+  describe('hook', () => {
```
XavierEkkel, line 115: nit: (same for others)
```suggestion
  describe('# hook', () => {
```

apps/smart-contracts/core/contracts/RedeemHook.sol
```diff
@@ -18,7 +19,15 @@ contract RedeemHook is
...
+    uint256 fee = amountBeforeFee - amountAfterFee;
+    if (fee > 0) {
+      IERC20 collateral = IPrePOMarket(msg.sender).getCollateral();
+      collateral.transferFrom(msg.sender, _treasury, fee);
```
XavierEkkel, line 27: nit:
```suggestion
      IPrePOMarket(msg.sender).getCollateral().transferFrom(msg.sender, _treasury, fee);
```

apps/smart-contracts/core/test/RedeemHook.test.ts
```diff
@@ -102,4 +111,75 @@ describe('=> RedeemHook', () => {
...
+            ceilingLongPayout: parseEther('0.8'),
+            floorValuation: BigNumber.from(10000000),
+            ceilingValuation: BigNumber.from(20000000),
+            expiryTime: nowPlusMonths(2),
+          },
+          user.address,
+          treasury.address
+        )
+        marketSigner = await getSignerForContract(market)
+        await redeemHook.setAllowedCallers([market.address], [true])
```
XavierEkkel, line 155: for these tests why is this necessary? couldn't you just set `user` as the allowed caller?

XavierEkkel (approved):
LGTM after comments

apps/smart-contracts/core/test/RedeemHook.test.ts
ramenforbreakfast: No, because we wrap `msg.sender` with `IPrePOMarket` to grab the `Collateral` contract with `getCollateral`, this would not be possible with a normal signer. Thus we need to imitate the contract as a signer.

However, I have noticed that we do not even need to instantiate a smock market here, since all it needs is the collateral, so we can just use a `FakeContract` here and directly return `Collateral`.

PR #635, opened 2022-12-03
chore:core: rename FeeRebateHook to TokenSenderCaller

davidprepo:
Per discussion, it was decided to rename this class after the feature was merged.

Additionally, ITokenSenderCaller, TokenSenderCaller, and ITokenSender have been moved to `prepo-shared-contracts`.


apps/smart-contracts/core/test/TokenSenderCaller.test.ts
```diff
@@ -4,22 +4,22 @@ import { SignerWithAddress } from '@nomiclabs/hardhat-ethers/dist/src/signer-wit
...
 import { grantAndAcceptRole } from './utils'
-import { FeeRebateHook } from '../typechain'
+import { TokenSenderCaller } from '../typechain'
 
 chai.use(smock.matchers)
 
-describe('=> FeeRebateHook', () => {
-  let feeRebateHook: FeeRebateHook
+describe('=> TokenSenderCaller', () => {
+  let feeRebateHook: TokenSenderCaller
```
XavierEkkel, line 14: please comprehensively rename all usages of 'feeRebate'

XavierEkkel (approved):
LGTM after comment

ramenforbreakfast (approved):
Can you move it to `prepo-shared-contracts` in this PR as well? If we are renaming it might as well move it over in the same PR.

apps/smart-contracts/core/test/fixtures/HookFixture.ts
```diff
@@ -24,9 +24,9 @@ export async function managerWithdrawHookFixture(): Promise<ManagerWithdrawHook>
   return (await managerWithdrawHook.deploy()) as ManagerWithdrawHook
 }
 
-export async function feeRebateHookFixture(): Promise<FeeRebateHook> {
-  const feeRebateHook = await ethers.getContractFactory('FeeRebateHook')
-  return (await feeRebateHook.deploy()) as FeeRebateHook
+export async function tokenSenderCallerFixture(): Promise<TokenSenderCaller> {
+  const tokenSenderCallerHook = await ethers.getContractFactory('TokenSenderCaller')
+  return (await tokenSenderCallerHook.deploy()) as TokenSenderCaller
 }
```
ramenforbreakfast, line 30:
```suggestion
export async function tokenSenderCallerFixture(): Promise<TokenSenderCaller> {
  const factory = await ethers.getContractFactory('TokenSenderCaller')
  return (await factory.deploy()) as TokenSenderCaller
}
```
As per other PRs, I know this is a departure of most of the others, but trying to slowly change over since no point in giving it a different name when it only exists within this context.

davidprepo:
> Can you move it to `prepo-shared-contracts` in this PR as well? If we are renaming it might as well move it over in the same PR.

TokenSenderCaller references TokenSender, do we want that moved over to `prepo-shared-contracts` too? I assume we don't want things in shared referencing core?

ramenforbreakfast:
> > Can you move it to `prepo-shared-contracts` in this PR as well? If we are renaming it might as well move it over in the same PR.
> 
> TokenSenderCaller references TokenSender, do we want that moved over to `prepo-shared-contracts` too? I assume we don't want things in shared referencing core?

No, we only want to move contracts that we would not instantiate ourselves in a deployment to `prepo-shared-contracts`. This is because we still would need to figure out how to have deployments within a project folder work with contracts defined outside the project.

Since we wouldn't deploy `TokenSenderCaller` on its own, its fine, but for `TokenSender`, we should keep it within `core`.

PR #637, opened 2022-12-03
chore:code: attempt at optimizing role assignment

davidprepo:
This PR introduces a batchGrantAndAcceptRoles function for two reasons:

- grantAndAcceptRoles was used often, resulting in the same copy-pasted code cluttering test files
- These calls were done with sequential awaits despite not depending on the execution of the previous call

This does not provide a large optimisation (saves 6 seconds off 11 minutes locally). If approved, moving other grantAndAcceptRoles to this new batch method can be a separate PR.


ramenforbreakfast:
Rebase this, convert it to a real PR, and get rid of the `OPTIMIZE` keyword, can just directly replace sequential granting of roles.

ramenforbreakfast (approved):
Added batching to WithdrawHook tests as well since these two contracts have the most role setting by far.

PR #639, opened 2022-12-04
refactor:core: Use snapshots for Collateral test

ramenforbreakfast:
Note that `smock` functions are `reset` at the end of the testcase. This is because I think there is an issue where when the snapshot is reset, the `smock` is not a real contract, and so the virtual state within it is persisted across testcases. One of the `deposit` testcases is still failing and I could not figure out why.


ramenforbreakfast:
Have fixed why the testcase was failing.

So when the EVM snapshot happens, this saves the state of the EVM. Since we just perform snapshot reverts after we initialize all global/local contract variables, and not redeploying these contracts every test, these global contract variables are still valid and existing in the network at time of snapshot.

The issue is then when we redeploy contracts to use the new base token. Since now we have overwritten the global contract variables with these new contracts, when we finish those tests, and go back to testing in the default 6 decimal environment, those global contract variables are pointing to an EVM where they never existed.

And you cannot just reinitialize the old 6 decimal environment, because those are technically new contracts, that also don't exist on the snapshot. So what I realized is we just have to update the snapshot after the different decimal testcases are done.

All tests are working for `Colllateral.test.ts`, currently verifying they don't break other tests.

This reduces the time on my machine from 16mins to 3 mins, so should be reduced from ~30 mins to 6 mins on GitHub Actions.

apps/smart-contracts/core/test/Collateral.test.ts
```diff
@@ -1025,13 +1054,18 @@ describe('=> Collateral', () => {
   })
 
   describe('# withdraw', () => {
+    before(async function () {
+      await setupCollateralStackForWithdrawals()
+      snapshotAfterEachTest = await ethers.provider.send('evm_snapshot', [])
+    })
```
XavierEkkel, line 1060: nit:
```suggestion
    })
    
```

apps/smart-contracts/core/test/Collateral.test.ts
```diff
@@ -231,13 +233,15 @@ describe('=> Collateral', () => {
     await setupWithdrawHook()
   }
 
-  before(() => {
+  before(async () => {
     upgrades.silenceWarnings()
+    snapshotBeforeTests = await ethers.provider.send('evm_snapshot', [])
```
XavierEkkel, line 238: is there any point to this one? since no evm action has been performed yet

apps/smart-contracts/core/test/Collateral.test.ts
```diff
@@ -38,6 +38,8 @@ describe('=> Collateral', () => {
   let managerWithdrawHook: MockContract<Contract>
   let tokenSender: FakeContract<Contract>
   let allowlist: MockContract<Contract>
+  let snapshotBeforeTests: string
+  let snapshotAfterEachTest: string
```
XavierEkkel, line 42: nit:
```suggestion
  let snapshotBeforeAllTests: string
  let snapshotBeforeEachTest: string
```

apps/smart-contracts/core/test/Collateral.test.ts
```diff
@@ -735,17 +748,31 @@ describe('=> Collateral', () => {
...
+         */
         await setupCollateralStackForDeposits()
+        sender = user1
+        recipient = user2
+        snapshotAfterEachTest = await ethers.provider.send('evm_snapshot', [])
       }
-      sender = user1
-      recipient = user2
-      expect(sender.address).to.not.eq(recipient.address)
+      // expect(sender.address).to.not.eq(recipient.address)
```
XavierEkkel, line 775: nit:
```suggestion
```

apps/smart-contracts/core/test/Collateral.test.ts
```diff
@@ -1243,6 +1278,8 @@ describe('=> Collateral', () => {
     })
 
     it('ignores hook if hook not set', async () => {
+      // reset smock contract since its callCount persists across all tests
+      withdrawHook.hook.reset()
```
XavierEkkel, line 1282: assuming this function takes negligible time to run, I think it'd be more reliable+readable+maintainable to reset all the relevant smocks for the describe block at the start of the `afterEach` instead

apps/smart-contracts/core/test/Collateral.test.ts
```diff
@@ -735,17 +748,31 @@ describe('=> Collateral', () => {
...
+      } else if (this.currentTest?.title.includes('mints to sender if sender = recipient')) {
+        /**
+         * We have to reset the stack here and take a new snapshot, because now the global
+         * contract variables have been overwritten by the special base token setups above.
+         * If we do not update the snapshot, the contracts we setup to return back to 6 decimals
+         * will be interacting with a network where they never existed.
+         */
         await setupCollateralStackForDeposits()
+        sender = user1
+        recipient = user2
```
XavierEkkel, line 772: why are these reinitialized? aren't `user1/2` still the same as in the `before` block?

apps/smart-contracts/core/test/Collateral.test.ts
```diff
@@ -735,17 +748,31 @@ describe('=> Collateral', () => {
...
+      snapshotAfterEachTest = await ethers.provider.send('evm_snapshot', [])
+    })
+
     beforeEach(async function () {
       if (this.currentTest?.title.includes('= base token decimals')) {
         await setupCollateralStackForDeposits(18)
       } else if (this.currentTest?.title.includes('< base token decimals')) {
         await setupCollateralStackForDeposits(19)
-      } else {
+      } else if (this.currentTest?.title.includes('mints to sender if sender = recipient')) {
```
XavierEkkel, line 763: if this is only for one specific test, I think it'd be clearer to just put these calls within that specific test?

apps/smart-contracts/core/test/Collateral.test.ts
```diff
@@ -735,17 +748,31 @@ describe('=> Collateral', () => {
...
+        /**
+         * We have to reset the stack here and take a new snapshot, because now the global
+         * contract variables have been overwritten by the special base token setups above.
+         * If we do not update the snapshot, the contracts we setup to return back to 6 decimals
+         * will be interacting with a network where they never existed.
+         */
         await setupCollateralStackForDeposits()
+        sender = user1
+        recipient = user2
+        snapshotAfterEachTest = await ethers.provider.send('evm_snapshot', [])
```
XavierEkkel, line 773: why is this resetting only done in the else block but not the `if` or `else if` blocks?

apps/smart-contracts/core/test/Collateral.test.ts
```diff
@@ -735,17 +748,31 @@ describe('=> Collateral', () => {
...
     let sender: SignerWithAddress
     let recipient: SignerWithAddress
+    before(async function () {
+      await setupCollateralStackForDeposits()
+      sender = user1
+      recipient = user2
+      snapshotAfterEachTest = await ethers.provider.send('evm_snapshot', [])
+    })
+
     beforeEach(async function () {
```
XavierEkkel, line 758: could we improve performance by putting these different setups into their own describe blocks with their own snapshots in their own `before` blocks?
that way we don't need to re-setup the stack multiple times for these cases, nor do we have to override the snapshot multiple times.

apps/smart-contracts/core/test/Collateral.test.ts
ramenforbreakfast: This is needed to reset the state of the EVM prior to the test suite, or else this will affect the next test suite that runs.

apps/smart-contracts/core/test/Collateral.test.ts
ramenforbreakfast: This would save maybe 5-15 seconds at best? I don't think this is worth it.

apps/smart-contracts/core/test/Collateral.test.ts
ramenforbreakfast: For the `deposit` testcases, we only need the special basetoken setups for 1 test case per base token decimals, so doesn't make sense to use snapshots since we never need to revert back to it.

The `wtihdraw` ones, we need them for two testcases per base token decimals, so if we saved it in a snapshot, it'd probably save 20-30 seconds?

apps/smart-contracts/core/test/Collateral.test.ts
ramenforbreakfast: Yes I just thought it might be better to put it up here to make it clear with all the other setups, like hey, for these special ones, we are doing these setups, and then right after we need to reset. I think its cleaner keeping them all together here in the `beforeEach` block.

apps/smart-contracts/core/test/Collateral.test.ts
XavierEkkel: have added to backlog for the `withdraw` cases

XavierEkkel (approved):
LGTM after remaining comments

PR #643, opened 2022-12-06
fix: correct collateral smock to run initialize and remove mock where it's not needed

ramenforbreakfast:
So earlier, when we were using the `mock` `Collateral` contract for test suites, I couldn't figure out an easy way to get smock to represent an upgradeable contract like `Collateral`. This is because `smock` does not support simulation of upgradeable contracts.

But because what we needed didn't really require any of the underlying token logic of `Collateral` for any of the test suites that mocked it, I just deployed it uninitialized.

Now that the `DepositTradeHelper` needs it to process approvals, I needed to fix this. While I was going through test suites, I also, instead of replacing any use of `smockCollateralFixture`, I just replace it with a `Fake` since we don't need the underlying logic to be simulated for any of these testcases.


XavierEkkel (approved):
LGTM

PR #644, opened 2022-12-06
feat: add permit fixes to Collateral and test DepositTradeHelper approvals

ramenforbreakfast:
So while working with approvals, I identified some 1-line changes to other contracts that were needed in order for `DepositTradeHelper` to work. I decided to first submit approvals first since this makes sense to submit and test separately.

- `Collateral` doesn't need to import both `ERC20Upgradeable` and `ERC20PermitUpgradeable`, this is because while the interfaces don't inherit each other, `ERC20Upgradeable` inherits `ERC20PermitUpgradeable`. This doesn't break anything, but its just not necessary.
- `ICollateral` needs to expose `permit` functions
- `TestERC20` our testing ERC20, needs to be made into a `ERC20Permit`
- You'll see I've defined two "junk" permits. This is because I mistakenly thought I could just submit empty structs, but `ethers.js` and im sure any provider will check before hand that your data at least follows the specified struct according to the abi, so we need to actually submit data, and just ensure its useless data.
- There is not a simple way to just check if a struct parameter is empty, so I decided to use `deadline` since its a very easy and reliable way to discard a permit.
- Forgot to fix `deadline`, should be a `uint256`, not an `address`


ramenforbreakfast:
Closing since should come after a different PR.

PR #645, opened 2022-12-06
feat:core: add return value to Collateral deposit

ramenforbreakfast:
We already added this to `PrePOMarket`, but forgot we need to add a return value to `deposit` as well so `DepositTradeHelper` does not have to guess how much was minted to the user.


apps/smart-contracts/core/contracts/Collateral.sol
```diff
@@ -78,8 +79,10 @@ contract Collateral is
       baseToken.approve(address(depositHook), 0);
     }
     /// Converts amount after fee from base token units to collateral token units.
-    _mint(_recipient, (_amountAfterFee * 1e18) / baseTokenDenominator);
+    uint256 _amountToMint = (_amountAfterFee * 1e18) / baseTokenDenominator;
```
XavierEkkel, line 82: nit: since all the other amount variables are referring to the base token, let's make it clearer that this is denominated in Collateral
```suggestion
    uint256 _collateralMintAmount = (_amountAfterFee * 1e18) / baseTokenDenominator;
```

apps/smart-contracts/core/test/Collateral.test.ts
```diff
@@ -963,6 +963,20 @@ describe('=> Collateral', () => {
...
     })
 
+    it('returns collateral minted', async () => {
+      const amountToDeposit = await baseToken.balanceOf(sender.address)
+      expect(amountToDeposit).to.be.gt(0)
+      expect(await baseToken.allowance(sender.address, collateral.address)).to.be.eq(
+        amountToDeposit
+      )
+      const fee = amountToDeposit.mul(await collateral.getDepositFee()).div(FEE_DENOMINATOR)
+      const expectedCT = amountToDeposit.sub(fee).mul(parseEther('1')).div(USDC_DENOMINATOR)
```
XavierEkkel, line 973: nit: add expect that this is >0

XavierEkkel (approved):
LGTM after nits

PR #646, opened 2022-12-06
fix:core: minor fixes to contracts caught for permits

ramenforbreakfast:
So while working with approvals, I identified some 1-line changes to other contracts that were needed in order for DepositTradeHelper to work.

- `Collateral` doesn't need to import both `ERC20Upgradeable` and `ERC20PermitUpgradeable`, this is because while the interfaces don't inherit each other, `ERC20Upgradeable` inherits `ERC20PermitUpgradeable`. This doesn't break anything, but its just not necessary.
- `ICollateral` needs to expose permit functions
- `TestERC20` our testing `ERC20`, needs to be made into a `ERC20Permit`
- Fix `deadline` to be a `uint256`
- Make `_baseToken` in `DepositTradeHelper` a `IERC20`


XavierEkkel (approved):
LGTM

PR #647, opened 2022-12-06
feat:core: add permit processing for depositAndTrade

ramenforbreakfast:
- You'll see I've defined two "junk" permits. This is because I mistakenly thought I could just submit empty structs, but ethers.js and im sure any provider will check before hand that your data at least follows the specified struct according to the abi, so we need to actually submit data, and just ensure its useless data.
- There is not a simple way to just check if a struct parameter is empty, so I decided to use deadline since its a very easy and reliable way to discard a permit.


apps/smart-contracts/core/test/DepositTradeHelper.test.ts
```diff
@@ -49,4 +72,116 @@ describe('=> DepositTradeHelper', () => {
...
+          deadline: timestampToSignFor,
+          v: signature.v,
+          r: signature.r,
+          s: signature.s,
+        },
+        junkPermit,
+        junkTradeParams
+      )
+
+      expect(await core.baseToken.allowance(user.address, depositTradeHelper.address)).to.eq(
```
XavierEkkel, line 100: expect allowance to be 0 in the arrange section
same for other tests

XavierEkkel (approved):
LGTM after comments

PR #650, opened 2022-12-07
feat:core: deposit and trade transfer logic

ramenforbreakfast:
The approach I went with is a balance between having actual token logic simulated and simply just verifying function call parameters.

I have included revert cases to ensure at the token level, transfers and approvals still operate as expected. However, to ensure we do not need to do more harness setup work and create large fixtures for stack setup, I have written tests in a way that ensures both call order and function call correctness utilizing `smock`'s handy `atCall` and `calledBefore`/`calledAfter` methods.

You will also notice redefining of `baseToken` and `Collateral` at the top, rather than directly referencing it via `core`. This is because although we can still setup the entire stack just using the harness, because the harness was written in a way that allows it to instantiate either as a smock or real deployment using the same variables, typescript doesn't know what kind of object it is, and so we have to cast it at the beginning of the `beforeEach` block. Doesn't impact performance since its just a casting, but not ideal.

Also you will notice the last case I have omitted the swap call since I figured that would be more appropriate to follow in a different PR.


apps/smart-contracts/core/test/DepositTradeHelper.test.ts
```diff
@@ -203,5 +252,100 @@ describe('=> DepositTradeHelper', () => {
         )
       })
     })
+
+    describe('testing with approvals already given', () => {
```
XavierEkkel, line 256:
```suggestion
    describe('if all permits provided', () => {
```

apps/smart-contracts/core/test/DepositTradeHelper.test.ts
```diff
@@ -203,5 +252,100 @@ describe('=> DepositTradeHelper', () => {
...
+        await setNextTimestamp(ethers.provider, timestampToSignFor)
+
+        await depositTradeHelper
+          .connect(user)
+          .depositAndTrade(baseTokenToDeposit, baseTokenPermit, collateralPermit, junkTradeParams)
+
+        expect(await baseToken.balanceOf(user.address)).to.be.eq(
+          userBTBalanceBefore.sub(baseTokenToDeposit)
+        )
+        expect(baseToken.transferFrom.atCall(0)).to.have.been.calledWith(
```
XavierEkkel, line 305: nit:
```suggestion
        expect(baseToken.transferFrom.atCall(0)).calledWith(
```
same/similarly for others

XavierEkkel (approved):
LGTM after nits

PR #651, opened 2022-12-07
feat: added salt parameters for create2 LS token deployments

ramenforbreakfast:
So this one was a little tricky to test, since we need the initial bytecode generated, but we do not deploy the LongShortToken ourselves, so we don't have access to the deployment transaction.

I use the `getDeployTransaction` method to give me what the bytecode would be as if `PrePOMarketFactory` had been the deployer, then turn it into a `keccak256` hash so it is a 32-byte value digestible by the salt address calculator.


apps/smart-contracts/core/test/PrePOMarket.test.ts
```diff
@@ -68,6 +69,8 @@ describe('=> prePOMarket', () => {
       factory: prePOMarketFactory,
       tokenNameSuffix: TEST_NAME_SUFFIX,
       tokenSymbolSuffix: TEST_SYMBOL_SUFFIX,
+      longTokenSalt: formatBytes32String('DEFAULT_SALT'),
+      shortTokenSalt: formatBytes32String('DEFAULT_SALT'),
```
XavierEkkel, line 73: nit: should use different values to be more realistic, same in the other file
```suggestion
      longTokenSalt: formatBytes32String('LONG_SALT'),
      shortTokenSalt: formatBytes32String('SHORT_SALT'),
```

XavierEkkel (approved):
LGTM after nit

PR #652, opened 2022-12-07
feat: add swap call logic and testing

ramenforbreakfast:
So for this test suite, you will see that there is an abnormal way for checking swap parameters are called correctly. This is because there seems to be a problem with `smock` recognizing `calledWith` parameters, if the parameter is a struct.

Since the swap parameters are passed in as a struct, I had to resort to this method I discovered that requires looping through the positional call arguments themselves within the smock object, so that I could verify them.


apps/smart-contracts/core/contracts/DepositTradeHelper.sol
```diff
@@ -57,6 +57,18 @@ contract DepositTradeHelper is IDepositTradeHelper, SafeOwnable {
       address(this),
       _collateralAmountMinted
     );
+    ISwapRouter.ExactInputSingleParams memory _routerParams = ISwapRouter
```
XavierEkkel, line 60: nit:
```suggestion
    ISwapRouter.ExactInputSingleParams memory exactInputSingleParams = ISwapRouter
```

apps/smart-contracts/core/test/DepositTradeHelper.test.ts
```diff
@@ -336,7 +336,47 @@ describe('=> DepositTradeHelper', () => {
           depositTradeHelper.address,
           baseTokenToDeposit
         )
-        // expect(collateral.transferFrom).to.have.been.calledBefore(swapRouter.exactInputSingle)
+        expect(collateral.transferFrom).to.have.been.calledBefore(swapRouter.exactInputSingle)
```
XavierEkkel, line 339: nit:
```suggestion
        expect(collateral.transferFrom).calledBefore(swapRouter.exactInputSingle)
```

XavierEkkel (approved):
LGTM after nits

PR #657, opened 2022-12-07
refactor: shared: Removing "new" from parameter names 

davidprepo:


ramenforbreakfast (approved):
Same thing as PR 655 and 656, implementation contracts should follow the new prefix convention.

XavierEkkel (approved):

PR #658, opened 2022-12-07
chore: core: use nominatedowner rather than governance

davidprepo:
Spec: https://www.notion.so/prepo/SC-Backlog-Tasks-0c6725bb870f4c22b7ba771347333662#6aca6a522f7a4e519d3faf5e1351062e


ramenforbreakfast (approved):
LGTM, probably going to need to change and rebase this given the changes in 655-657

XavierEkkel (approved):
is this changing to SafeOwnable? if not, the argument name can just be `owner` (and no underscore prefix)

PR #660, opened 2022-12-07
chore: reduce test verbosity

davidprepo:
Spec: https://www.notion.so/prepo/SC-Backlog-Tasks-0c6725bb870f4c22b7ba771347333662#10095c3cc740423fa2b40f713c8ef226

This PR touches many files but is singular in scope and extremely low in complexity, only removing surplus verbocity in test cases. As such, splitting this across multiple PRs did not seem to be wise.

Scope:

- Use `.revertedWith` instead of `.to.be.revertedWith` for conciseness
- Use `.reverted` instead of `.to.be.reverted` for conciseness
- Use `.not.called` instead of `to.not.be.called`
- Use `.calledWith` instead of `to.be.calledWith`


ramenforbreakfast (approved):
LGTM

PR #661, opened 2022-12-07
chore: core: rename isCollateralValid to isValidCollateral

davidprepo:
Spec: https://www.notion.so/prepo/SC-Backlog-Tasks-0c6725bb870f4c22b7ba771347333662#e182cba999a441e5ae77ba2378213857

`fetchExistingCollateral` was required from the deploy contract, but commented out in helpers. I have reinstated the function and just have it throw a not implemented error pending direction from either @XavierEkkel or @ramenforbreakfast .


apps/smart-contracts/core/helpers/index.ts
```diff
@@ -22,15 +22,17 @@ async function fetchExistingDeploymentFromEnvironment(
...
 // TODO: replace collateral with preUSD
-// export async function fetchExistingCollateral(
-//   chainId: string,
-//   ethers: HardhatEthersHelpers
-// ): Promise<Collateral> {
-//   return (await fetchExistingDeploymentFromEnvironment(
-//     `COLLATERAL_${chainId}`,
-//     await ethers.getContractFactory('Collateral')
-//   )) as Collateral
-// }
```
ramenforbreakfast, line 33: Can just delete, no longer relevant (shouldn't break anything since we already comented it out anyway?)

ramenforbreakfast (approved):
LGTM

apps/smart-contracts/core/helpers/index.ts
davidprepo: The function is called elsewhere but had been commented out, and then the pre-commit checks fail.

XavierEkkel (approved):

PR #662, opened 2022-12-07
chore: token: Rename token directories

davidprepo:
Spec: https://www.notion.so/prepo/SC-Backlog-Tasks-0c6725bb870f4c22b7ba771347333662#935bf19fb9004a8387c22d15c2b6d678

Decided to put both renames in a single PR as the directory rename is picked up well by github and there are only a couple of small file changes.

Changes:
- Rename `token-shop` directory to just `shop`
- Rename `ppo-staking` directory to just `staking`


ramenforbreakfast (approved):
LGTM

PR #666, opened 2022-12-07
refactor: core: Tests use FakeContracts where sensible

davidprepo:
Spec: https://www.notion.so/prepo/SC-Backlog-Tasks-0c6725bb870f4c22b7ba771347333662#8531079b88c34f27ba8709a98d8776dc

> Convert all contracts at the lowest level in a test suite (ones that we do not need to verify internal logic for) to `FakeContract`’s. In some cases contract logic does actually need to be represented so this should not happen for all contracts external to the main contract of the test suite. Once you do that, you can remove any existing configuration within the test suite for the `FakeContract` since their logic is not simulated.

While ensuring `FakeContract`s are used wherever sensible, it also seemed sensible to take advantage of smock's typing ability for the fakes. Moving to fakes also allows us to use the signer of the fake directly, unlike with mocks.

`WithdrawHook.test.ts` shows as a large diff due to a large number of lines having `collateralSigner` replaced with `collateral.wallet` to take advantage of the `FakeContract` feature.


apps/smart-contracts/core/test/RedeemHook.test.ts
```diff
@@ -145,7 +138,7 @@ describe('=> RedeemHook', () => {
 
     describe('fee reimbursement', () => {
       beforeEach(async () => {
-        await allowlist.set([user.address], [true])
+        allowlist.isIncluded.whenCalledWith(user.address).returns(true)
```
XavierEkkel, line 141: couldnt this be in a `before` rather than `beforeEach`?

apps/smart-contracts/core/test/utils.ts
```diff
@@ -104,6 +104,13 @@ export async function getSignerForContract(
   return signer
 }
 
+export async function setAccountBalance(address: string, eth = '0.1'): Promise<void> {
```
XavierEkkel, line 107: i dont think having a default for the amount is intuitive - let's make the calling file specify it explicitly instead

apps/smart-contracts/core/test/DepositHook.test.ts
```diff
@@ -47,8 +46,8 @@ describe('=> DepositHook', () => {
     firstERC721 = await testERC721Fixture('NFT Collection 1', 'NFT1')
     secondERC721 = await testERC721Fixture('NFT Collection 2', 'NFT2')
     collateral = await fakeCollateralFixture()
+    await setAccountBalance(collateral.address)
```
XavierEkkel, line 49: what is this replacing? it seems to be a new addition without any corresponding balance setting logic being removed

XavierEkkel (approved):
LGTM after comments

apps/smart-contracts/core/test/DepositHook.test.ts
davidprepo: It's replacing the `getSignerForContract()` call. When using `FakeContract` we can use the wallet signer for that contract address directly, but it still needs some funds.

PR #674, opened 2022-12-08
refactor: core: Convert tests to use snapshots (part 1)

davidprepo:
Spec: https://www.notion.so/prepo/SC-Backlog-Tasks-0c6725bb870f4c22b7ba771347333662#f3232ddf2d484dd58dfa29d459c52898

This PR includes the `Snapshotter` class, and refactors the `DepositHook` test to use it.

Design:
- We use a stack model for snapshots to ensure that we are only working with one snapshot at a given point in time.
  - This is due to the fact that snapshots are not entirely independent - it is only possible to go "back in time", not forwards
  - As such, we must be careful to not reset to an older snapshot if we still want to be using a newer one
- All snapshot logic is abstracted away inside `usesCustomSnapshot` which simply needs to be called inside relevant `describe` blocks, and `saveSnapshot` which can be called (but isn't required) from the `before` block after setup is complete.
  - Handles all setup of needed mocha hooks
  - Doesn't matter where this is called, but putting it directly above the `before` block which does the setup seems sensible
- We must call `reset` on any fake/mock functions if we are doing any called/not called checks


apps/smart-contracts/core/test/LongShortToken.test.ts
```diff
@@ -12,13 +13,19 @@ describe('=> LongShortToken', () => {
   let deployer: SignerWithAddress
   let user: SignerWithAddress
   let user2: SignerWithAddress
+  const setupSnapshot: Snapshot = new Snapshot('Setup')
```
davidprepo, line 16: Unsure of the best naming scheme for this, it will be repeated across all tests. OUTER_SNAPSHOT, TEST_SNAPSHOT, MAIN_SNAPSHOT would be decent candidates if we want to stick with const = all caps.

ramenforbreakfast:
@davidprepo The reason why your nested resets don't work is on the iteration before `hook`, what will happen is the root `beforeEach` `TEST_SNAPSHOT` reset will happen after the nested `before` block. `before` blocks ALWAYS run after `beforeEach`, even if it is nested. So it is resetting to `TEST_SNAPSHOT` _after_ the initial `before` setup in `describe('# hook')` happens, making the setup not have any effect.

XavierEkkel (commented):
will wait until @ramenforbreakfast's comment is addressed

ramenforbreakfast:
I have edited my comments to reflect the new implementation.

While I am mostly satisfied with the solution you have come up with, there is one big concern I think that needs to be addressed. I will hold off on further debugging and allow you to come up with a solution. I think it has to do with us not completely resetting the VM to the state _prior to the test suite running_ as we talked about earlier today.

The following scenario occurred, after I tried to convert the `Collateral.test.ts` to use your snapshot library as well. The `Collateral` tests have no issues, but when the `DepositHook.test.ts` tries to run, it fails for many testcases. However, if you try to run the `DepositHook.test.ts` by itself, these issues do not occur, hence why I think it has to do with resetting the evm.

Note: In the screenshots below, I added debug statements to alert me whenever the snapshot is hooked into a `beforeEach` block. Notice that for every single test, it is correctly hooking on to the `beforeEach` block. I am establishing hook contexts the exact same way as you have done in `DepositHook.test.ts`, outside the `before` block of the `describe` block, and then calling `saveSnapshot` within the actual `before` block. **I am saving snapshots somewhat unconventionally with the different base token tests in `Collateral.test.ts`, but I am pretty sure it has nothing to do with what happening here.**

image
image

However, when `DepositHook.test.ts` runs, it does not seem to be hooking into the `beforeEach` block
image

ramenforbreakfast:
@davidprepo Just a heads up, I have submitted a commit that includes the migration of `Collateral.test.ts` so that you can reproduce the issue on your end.

apps/smart-contracts/core/test/DepositHook.test.ts
```diff
@@ -33,7 +35,8 @@ describe('=> DepositHook', () => {
   const TEST_AMOUNT_BEFORE_FEE = parseEther('1.01')
   const TEST_AMOUNT_AFTER_FEE = parseEther('1')
 
-  beforeEach(async () => {
+  snapshotter.usesCustomSnapshot('outer')
```
ramenforbreakfast, line 38:
```suggestion
  snapshotter.usesCustomSnapshot('DepositHook')
```

apps/smart-contracts/core/test/DepositHook.test.ts
```diff
@@ -145,7 +149,8 @@ describe('=> DepositHook', () => {
      * Tests below use different values for TEST_AMOUNT_BEFORE_FEE and
      * TEST_AMOUNT_AFTER_FEE to ensure TEST_AMOUNT_BEFORE_FEE is ignored.
      */
-    beforeEach(async () => {
+    snapshotter.usesCustomSnapshot('hook')
```
ramenforbreakfast, line 152:
```suggestion
    snapshotter.usesCustomSnapshot('DepositHook-hook')
```

apps/smart-contracts/core/test/DepositHook.test.ts
```diff
@@ -254,6 +262,7 @@ describe('=> DepositHook', () => {
     })
 
     it('calls recordDeposit() if fee = 0', async () => {
+      depositRecord.recordDeposit.reset()
```
ramenforbreakfast, line 265: these smock resets should be bundled in the `afterEach`

apps/smart-contracts/core/test/DepositHook.test.ts
```diff
@@ -309,13 +323,15 @@ describe('=> DepositHook', () => {
   })
 
   describe('# setAccountList', () => {
-    beforeEach(async () => {
+    snapshotter.usesCustomSnapshot('setAccountList')
```
ramenforbreakfast, line 326:
```suggestion
    snapshotter.usesCustomSnapshot('DepositHook-setAccountList')
```

ramenforbreakfast (approved):
LGTM, besides a few nits from what we talked about earlier.

apps/smart-contracts/core/test/snapshots.ts
```diff
@@ -0,0 +1,79 @@
...
+      await this.headSnapshot().reset()
+    } else if (this.snapshotStack.length === 0) {
+      // Otherwise we've popped the last one off the stack, so we're done, and our
+      // outer setup is no longer complete.
+      this.outerSetupComplete = false
+    }
+  }
+
+  // This is the actual function which sets up snapshotting for a given block.
+  public usesCustomSnapshot(name: string | undefined = undefined): void {
```
ramenforbreakfast, line 46:
```suggestion
  public setupSnapshotContext(name: string | undefined = undefined): void {
```
This still hasn't been changed

apps/smart-contracts/core/test/DepositHook.test.ts
```diff
@@ -34,7 +34,8 @@ describe('=> DepositHook', () => {
   const TEST_AMOUNT_BEFORE_FEE = parseEther('1.01')
   const TEST_AMOUNT_AFTER_FEE = parseEther('1')
 
-  beforeEach(async () => {
+  usesCustomSnapshot('outer')
```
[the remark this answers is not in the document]
ramenforbreakfast: This is still not addressed

apps/smart-contracts/core/test/DepositHook.test.ts
```diff
@@ -250,73 +254,81 @@ describe('=> DepositHook', () => {
...
-    beforeEach(async () => {
+    usesCustomSnapshot('setAccountList')
+    before(async () => {
       await grantAndAcceptRole(
         depositHook,
         deployer,
         deployer,
         await depositHook.SET_ACCOUNT_LIST_ROLE()
       )
+      await saveSnapshot()
```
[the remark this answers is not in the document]
ramenforbreakfast: This still hasn't been removed. There should be any `before` block in this testcase

XavierEkkel (approved):
LGTM after questions

apps/smart-contracts/core/test/DepositHook.test.ts
```diff
@@ -206,6 +212,7 @@ describe('=> DepositHook', () => {
 
     it('succeeds if account not on allowlist, and required score = 0', async () => {
       await setupScoresForNFTAccess(0, 0)
+      allowlist.isIncluded.whenCalledWith(user.address).returns(false)
```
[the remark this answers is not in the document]
davidprepo: Before, there would be a new `FakeContract<AllowList>` created for every test. Now we are reusing the same fake, and we need to ensure the proper return value here.

apps/smart-contracts/core/test/DepositHook.test.ts
```diff
@@ -33,7 +35,8 @@ describe('=> DepositHook', () => {
   const TEST_AMOUNT_BEFORE_FEE = parseEther('1.01')
   const TEST_AMOUNT_AFTER_FEE = parseEther('1')
 
-  beforeEach(async () => {
+  snapshotter.setupSnapshotContext('DepositHook')
```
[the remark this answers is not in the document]
davidprepo: This exists at the `describe` level as this function call actually adds other `before`, `beforeEach`, `after`, or `afterEach` hooks to the describe block to enforce proper snapshot behaviour.

PR #680, opened 2022-12-13
refactor: move to shared harnesses and introduce a "Base" core harness with shared methods

ramenforbreakfast:
Thinking in the medium-short term, once we figure out the Arbitrage doc specs, we will need to start working on a script that needs a stack to test against. 

The harness we introduced for `DepositTradeHelper` isn't really workable for this scenario and needs to be available outside of the test suite.

This PR refactors the harness to extend from a `Base` core harness, that will contain all the methods that both "real" and "mock" stacks will use. 

Using this `Base` harness, we can even in the future when we write our deployment scripts, utilize the harness and harness methods to configure our contracts (rather than creating monolithic setups in each deployment script).

Instead, contracts would be deployed, and then setup separately, initializing a harness that connects to the contracts that were just deployed. This way we do not have to repeat configuration code twice.

This PR contains both the `Base` and `Core` class for deploying a "real" stack (not mock), since there'd be little point in just showing the reviewer the `Base` class.


ramenforbreakfast:
closing, will have this follow up change to new type directory structure.

PR #681, opened 2022-12-13
refactor: introduce new mock stack harness, and migrate current tests to use it

ramenforbreakfast:
Continuing off the previous PR, this introduces the mock version of the stack. Currently only one test, `DepositTradeHelper.test.ts` uses it, and needs minimal refactoring to switch.

This also deletes any harness files that were just under `test/`


ramenforbreakfast:
closing, will have this follow up change to new type directory structure.

PR #685, opened 2022-12-13
feat: chart stats

chrisling-dev:
https://www.notion.so/prepo/FE-Project-Management-d3878c29e9534234a50f4f8838525bee?p=05c99818423f461689f080a66078664a&pm=s


XavierEkkel (approved):
LGTM from UI perspective after also changing the loading box color for the stats text to grey (like with the chart)

PR #688, opened 2022-12-14
feat: core: Switch PrePOMarket to SafeAccessControlEnumerable

davidprepo:
Spec: https://www.notion.so/prepo/SC-Backlog-Tasks-0c6725bb870f4c22b7ba771347333662#84f44167cf584145833e79dcf9319f99

- This only switches `PrePOMarket` to `SafeAccessControlEnumerable`, and doesn't change the factory.
- Removed unused variable `user2`

Additionally, I've added a couple of helper functions to utils to reduce amount of copy/pasted test code we're doing when testing roles.
- The goal for these functions was to minimise call verbosity in order to improve readability of the test files
- `testRoleConstants` is a utility to do all our role constant checks without requiring the same copy/pasted test framework around it
  - Takes a flat array (rather than, for example, an array of tuples) in order to minimise call signature when using it
  - Exposes better error message handling (now displays the actual name which failed rather than `0x1234... != 0x3456...`
- `revertIfNotRoleHolder` does exactly as it suggests and implements the "reverts if not role holder" test we have duplicated everywhere
  - Uses `PopulatedTransaction` in order to not have to use `sender` twice in the function call
  - By using `PopulatedTransaction` we can then generate (in the case of the sender account) or find (in the case of the contract) the required information inside the function


apps/smart-contracts/core/contracts/PrePOMarket.sol
```diff
@@ -7,8 +7,13 @@ import "./interfaces/IHook.sol";
 import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
 import "@openzeppelin/contracts/access/Ownable.sol";
 import "@openzeppelin/contracts/security/ReentrancyGuard.sol";
+import "prepo-shared-contracts/contracts/SafeAccessControlEnumerable.sol";
 
-contract PrePOMarket is IPrePOMarket, Ownable, ReentrancyGuard {
+contract PrePOMarket is
+  IPrePOMarket,
+  SafeAccessControlEnumerable,
+  ReentrancyGuard
```
ramenforbreakfast, line 15:
```suggestion
  ReentrancyGuard,
  SafeAccessControlEnumerable,
```

apps/smart-contracts/core/contracts/PrePOMarket.sol
```diff
@@ -58,7 +72,7 @@ contract PrePOMarket is IPrePOMarket, Ownable, ReentrancyGuard {
     require(_expiryTime > block.timestamp, "Invalid expiry");
     require(_ceilingLongPayout <= MAX_PAYOUT, "Ceiling cannot exceed 1");
 
-    transferOwnership(owner);
+    _grantRole(DEFAULT_ADMIN_ROLE, owner);
```
ramenforbreakfast, line 75: What about revoking `DEFAULT_ADMIN_ROLE` from the `PrePOMarketFactory`, (i.e. `msg.sender`)

apps/smart-contracts/core/test/PrePOMarket.test.ts
```diff
@@ -102,10 +113,25 @@ describe('=> prePOMarket', () => {
...
+      ])
+    })
+
+    it('sets governance as role admin', async () => {
+      const createMarketResult = await createMarket(defaultParams)
+      prePOMarket = await prePOMarketAttachFixture(createMarketResult)
 
-      expect(await prePOMarket.owner()).to.eq(treasury.address)
+      expect(await prePOMarket.hasRole(DEFAULT_ADMIN_ROLE, treasury.address)).eq(true)
     })
```
ramenforbreakfast, line 135: Should also add a test to revoke `DEFAULT_ADMIN_ROLE` from the deployer.

apps/smart-contracts/core/test/utils.ts
```diff
@@ -126,3 +128,36 @@ export async function getPermitFromSignature(
     s: signature.s,
   }
 }
+
+export async function revertIfNotRoleHolder(
+  roleGetter: Promise<string>,
+  populatedTransactionPromise: Promise<PopulatedTransaction>
```
ramenforbreakfast, line 134:
```suggestion
  rolePromise: Promise<string>,
  populatedTransactionPromise: Promise<PopulatedTransaction>
```
naming consistency, since `const role` should `= await rolePromise`

apps/smart-contracts/core/test/PrePOMarket.test.ts
```diff
@@ -505,6 +536,7 @@ describe('=> prePOMarket', () => {
 
     const setupMarket = async (): Promise<BigNumber> => {
       prePOMarket = await prePOMarketAttachFixture(await createMarket(defaultParams))
+      await grantAllRoles(treasury)
```
ramenforbreakfast, line 539: We could put both 
```
prePOMarket = await prePOMarketAttachFixture(await createMarket(defaultParams))
await grantAllRoles(treasury)
```
In the `beforeEach`, _if_ we refactored it so `mint` and `redeem` were isolated from the root block. But understand if we want to leave this out for this PR. Probably can do it when we refactor for snapshots.

ramenforbreakfast (approved):
LGTM, but I would move these utilities to `prepo-hardhat` under `smart-contract-utils`. Besides my small nits, I think you need to connect the wallet to a provider since tests are breaking. But thanks for the nice addition to our testing framework!

apps/smart-contracts/core/test/PrePOMarket.test.ts
davidprepo: Yes I think this can go into the snapshot refactor

apps/smart-contracts/core/contracts/PrePOMarket.sol
```diff
@@ -31,6 +36,15 @@ contract PrePOMarket is IPrePOMarket, Ownable, ReentrancyGuard {
   uint256 private constant FEE_DENOMINATOR = 1000000;
   uint256 private constant FEE_LIMIT = 100000;
 
+  bytes32 public constant SET_MINT_HOOK_ROLE =
+    keccak256("PrePOMarket_setMintHook(IHook)");
```
ramenforbreakfast, line 40:
```suggestion
    keccak256("setMintHook");
```
Same for other roles, as per discord discussion.

apps/smart-contracts/core/test/PrePOMarket.test.ts
```diff
@@ -102,10 +113,32 @@ describe('=> prePOMarket', () => {
       expect(await prePOMarket.getFeeLimit()).to.eq(MARKET_FEE_LIMIT)
     })
 
-    it('should set owner to governance', async () => {
+    it('sets role constants to the correct hash', async () => {
```
XavierEkkel, line 116:
```suggestion
    it('sets role constants to correct hash', async () => {
```

XavierEkkel (approved):
LGTM after nits

PR #693, opened 2022-12-15
feat: shared: Remove SafeOwnable from WithdrawERC20 and move access control logic to inheriting contracts

davidprepo:
Spec: https://www.notion.so/prepo/SC-Backlog-Tasks-0c6725bb870f4c22b7ba771347333662#5637e32e2d7b4d0999ea94895f8a7cac

- Added the same role for both withdraw functions as this maintains the previous logic and they both share the same name
- Changes in the `SafeAccessControlEnumerable` tests are only from the linter, not real changes
- Modifier on the functions in `WithdrawERC20` will be still be called when accessing these functions through `super.func()`


apps/smart-contracts/core/contracts/TokenSender.sol
```diff
@@ -31,6 +31,8 @@ contract TokenSender is
     keccak256("TokenSender_setScaledPriceLowerBound(uint256)");
   bytes32 public constant SET_ALLOWED_MSG_SENDERS_ROLE =
     keccak256("TokenSender_setAllowedMsgSenders(IAccountList)");
+  bytes32 public constant WITHDRAW_ERC20_ROLE =
+    keccak256("TokenSender_withdrawERC20(address[])");
```
ramenforbreakfast, line 35:
```suggestion
  bytes32 public constant WITHDRAW_ERC20_ROLE =
    keccak256("withdrawERC20");
```
I am in support of keeping the same role for both `withdrawERC20` functions, can change the other role signature later.

packages/prepo-shared-contracts/test/SafeAccessControlEnumerableUpgradeable.test.ts
```diff
@@ -162,9 +162,7 @@ describe('SafeAccessControlEnumerableUpgradeableTest', () => {
         false
       )
 
-      await expect(
-        safeAccessControlEnumerable.connect(user1).acceptRoleAdmin(roleB)
-      ).revertedWith(
+      await expect(safeAccessControlEnumerable.connect(user1).acceptRoleAdmin(roleB)).revertedWith(
```
ramenforbreakfast, line 165: This is just because we forgot to run the linter when we removed `to.be.revertedWith` calls everywhere or something? Nothing has changed here right?

ramenforbreakfast (dismissed):

packages/prepo-shared-contracts/test/SafeAccessControlEnumerableUpgradeable.test.ts
davidprepo: Yes correct

apps/smart-contracts/core/contracts/TokenSender.sol
```diff
@@ -107,4 +108,19 @@ contract TokenSender is
   {
     return _scaledPriceLowerBound;
   }
+
+  function withdrawERC20(
+    address[] calldata erc20Tokens,
+    uint256[] calldata amounts
+  ) public override onlyRole(WITHDRAW_ERC20_ROLE) {
+    super.withdrawERC20(erc20Tokens, amounts);
```
ramenforbreakfast, line 116: I do not think this will work, `withdrawERC20` sends to `msg.sender`. In this case, this is an internal call, so it would send to the contract itself.

I think the appropriate fix here would be to change `withdrawERC20` to accept a `recipient`, and leave it up to the inheriting contract to decide whether to lock sending to `msg.sender` or not.

apps/smart-contracts/core/test/TokenSender.test.ts
```diff
@@ -75,6 +75,7 @@ describe('=> TokenSender', () => {
       expect(await tokenSender.SET_ALLOWED_MSG_SENDERS_ROLE()).to.eq(
         id('TokenSender_setAllowedMsgSenders(IAccountList)')
       )
+      expect(await tokenSender.WITHDRAW_ERC20_ROLE()).eq(id('withdrawERC20'))
     })
```
ramenforbreakfast, line 79: Where are the tests to ensure it is callable?

ramenforbreakfast (changes requested):

ramenforbreakfast (approved):

XavierEkkel (approved):
LGTM but should also change to `SafePausable` for many of these contracts later

apps/smart-contracts/core/test/TokenSender.test.ts
```diff
@@ -358,4 +365,24 @@ describe('=> TokenSender', () => {
       expect(await outputToken.balanceOf(user.address)).to.be.eq(outputAmount)
     })
   })
+
+  describe.only('# withdrawERC20', () => {
```
XavierEkkel, line 369:
```suggestion
  describe('# withdrawERC20', () => {
```

PR #694, opened 2022-12-16
chore: updated TypeChain packages, made core type directory consistent with others

ramenforbreakfast:
As talked about previously, this PR makes the `core` project follow the same type output structure as the other projects.

This does mean that there needed to be a minor change to the import of every test/fixture in `core`.

Also resolved in this PR are minor edits to the `package.json` files for all projects, since we had different `typescript` and `typechain` versions between `core` and the rest of the repo. Have now moved these to the root since there is no reason why these should be different.


XavierEkkel (approved):
LGTM if all tests still pass

PR #695, opened 2022-12-16
refactor: introduce base class for shared harnesses

ramenforbreakfast:
Thinking in the medium-short term, once we figure out the Arbitrage doc specs, we will need to start working on a script that needs a stack to test against.

The harness we introduced for `DepositTradeHelper` isn't really workable for this scenario and needs to be available outside of the test suite.

This PR refactors the harness to extend from a `Base` core harness, that will contain all the methods that both "real" and "mock" stacks will use.

Using this Base harness, we can even in the future when we write our deployment scripts, utilize the harness and harness methods to configure our contracts (rather than creating monolithic setups in each deployment script).

Instead, contracts would be deployed, and then setup separately, initializing a harness that connects to the contracts that were just deployed. This way we do not have to repeat configuration code twice.


apps/smart-contracts/core/types/harnesses.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export type MarketWithHooks = PrePOMarket & {
+  longToken?: ERC20
+  shortToken?: ERC20
+  hash?: string
+  mintHook?: MintHook
+  redeemHook?: RedeemHook
+}
+
+// No hash because smock PrePOMarkets are not created using factories
```
XavierEkkel, line 42: do you mean salt or hash? not clear what hash is

apps/smart-contracts/core/types/harnesses.ts
```diff
@@ -0,0 +1,48 @@
...
+  DepositHook,
+  ERC20,
+  ManagerWithdrawHook,
+  MintHook,
+  PrePOMarket,
+  RedeemHook,
+  WithdrawHook,
+} from './generated'
+
+export type DepositHookWithList = DepositHook & {
```
XavierEkkel, line 14:
```suggestion
export type DepositHookWithAllowlist = DepositHook & {
```

apps/smart-contracts/core/types/harnesses.ts
```diff
@@ -0,0 +1,48 @@
...
+  allowlist?: AccountList
+}
+
+export type MockDepositHookWithList = MockContract & {
```
XavierEkkel, line 18:
```suggestion
export type MockDepositHookWithAllowlist = MockContract & {
```

XavierEkkel, line 18: are we able to specify the type of MockContract? e.g. `MockContract<DepositHook>`

XavierEkkel (approved):
LGTM after comments/questions

apps/smart-contracts/core/types/harnesses.ts
ramenforbreakfast: There is a hash in the `PrePOMarketFactory` created from the addresses of the Long and Short tokens belonging to that market, this is used to store a record of it for later retrieval.

apps/smart-contracts/core/types/harnesses.ts
ramenforbreakfast: This will be addressed once PR #699 is merged in.

ramenforbreakfast:
Since I don't want to retype all of these fixtures and harnesses, I am going to hold up on #699 else it will be a game of cat and mouse, having to modify recently modified content to conform with new typing in #699.

PR #706, opened 2022-12-21
refactor: move event fixtures out into more generalized ones shared by all

ramenforbreakfast:
So this PR removes some old event finding fixtures that we had used for our old tests, but now we just verify events directly.

However, there are cases, in the harness setup, and for future tasks, where we will want to search for an event. To make it generalizable, I have written these event searcher functions with the ability to _optionally_ provide a start and end block, which we have needed to do before with ACQR.

These now exist within the `helpers` folder since they will be used by tasks and harness code.


apps/smart-contracts/core/helpers/events.ts
```diff
@@ -0,0 +1,24 @@
...
+import { MarketAddedEvent } from '../types/generated/contracts/PrePOMarketFactory'
+
+export async function findMarketAddedEvent(
+  factory: PrePOMarketFactory | MockContract<PrePOMarket>,
+  startBlock?: string,
+  endBlock = 'latest'
+): Promise<MarketAddedEvent[]> {
+  const filter = factory.filters.MarketAdded()
+  const events = await factory.queryFilter(filter, startBlock, endBlock)
+  return events as unknown as MarketAddedEvent[]
```
XavierEkkel, line 13: is there a way to do this without `as unknown`?

XavierEkkel (approved):
LGTM

PR #707, opened 2022-12-21
refactor: move market helper functions to helper folder and reduce code duplication

ramenforbreakfast:
So we have fixtures, that were originally designed to simplify market deployment, since there are alot of parameters, we created fixtures and parameter types to shorten how we'd pass and initialize markets in our test suites. 

A lot of this code is needed within the harness as well, and has no reason to be limited to tests, so I am moving it out and also moving market-specific type declarations out of `fixture` files and into their specific `types/market.ts` file.


XavierEkkel (approved):
LGTM

PR #708, opened 2022-12-21
feat: add function to deploy and record a smock market to a harness

ramenforbreakfast:
Since we cannot deploy a smock market normally (since it's deployed via `PrePOMarketFactory`), we need a special fixture to deploy and add one to our harnesses for forked network testing.


XavierEkkel (approved):
LGTM

PR #713, opened 2022-12-21
feat: partners page

chrisling-dev:
https://www.notion.so/prepo/FE-Project-Management-d3878c29e9534234a50f4f8838525bee?p=72781721630d45809222dfd76ff20022&pm=s


XavierEkkel (approved):
LGTM

apps/frontend/docs/.eslintrc.js
```diff
@@ -1,5 +1,9 @@
 module.exports = {
   ...require('config/eslint-frontend'),
+  rules: {
+    ...require('config/eslint-frontend').rules,
+    'import/no-unresolved': [2, { ignore: ['^@theme', '^@docusaurus', '^@site'] }],
```
XavierEkkel, line 5: what is this line doing?

chrisling-dev: This line tells eslint to ignore import unresolved rule for docusaurus aliases because these are just aliases, not the actual file path.

This is required because I need to import useBaseUrl from @docusaurus/useBaseUrl but there's no such file as "@docusaurus/useBaseUrl", the @docusaurus is just an alias.

PR #722, opened 2022-12-23
fix: instant hover effect on links

chrisling-dev:
Instant hover effect! (e.g. border turning purple, text underline)

Tested that it works for header items, link in docs, bottom navigation (next/previous page), let me know if anywhere else still has slow transition @XavierEkkel


XavierEkkel (approved):
LGTM

PR #727, opened 2022-12-25
feat: mintAndSell without swap

ramenforbreakfast:


apps/smart-contracts/core/contracts/ArbitrageBroker.sol
```diff
@@ -25,7 +25,12 @@ contract ArbitrageBroker is IArbitrageBroker, SafeAccessControlEnumerable {
   function mintAndSell(
     IPrePOMarket market,
     OffChainTradeParams calldata tradeParams
-  ) external override onlyRole(MINT_AND_SELL_ROLE) returns (uint256) {}
+  ) external override onlyRole(MINT_AND_SELL_ROLE) returns (uint256) {
+    if (!_marketToValidity[address(market)])
+      revert InvalidMarket(address(market));
```
XavierEkkel, line 30: nit: prefer using braces unless it fits on one line

apps/smart-contracts/core/contracts/ArbitrageBroker.sol
```diff
@@ -25,7 +25,12 @@ contract ArbitrageBroker is IArbitrageBroker, SafeAccessControlEnumerable {
...
+    _collateral.approve(address(market), tradeParams.longShortAmount);
```
XavierEkkel, line 31: add dev comment in interface that this assumes that at least `longShortAmount` collateral has been sent to the contract beforehand

apps/smart-contracts/core/contracts/interfaces/IArbitrageBroker.sol
```diff
@@ -21,6 +21,8 @@ interface IArbitrageBroker {
 
   event MarketValidityChanged(address market, bool validity);
 
+  error InvalidMarket(address market);
```
XavierEkkel, line 24: 👍 agree on transitioning towards use of errors in interfaces going forward

apps/smart-contracts/core/test/ArbitrageBroker.test.ts
```diff
@@ -5,29 +5,41 @@ import { SignerWithAddress } from '@nomiclabs/hardhat-ethers/dist/src/signer-wit
...
+    collateralLimitForLong: parseEther('2'),
+    collateralLimitForShort: parseEther('3'),
+  }
+
   snapshotter.setupSnapshotContext('ArbitrageBroker')
   before(async () => {
     core = await MockCore.Instance.init(ethers, parseEther('100000'), parseEther('10000'))
     ;[deployer, governance, user] = core.accounts
     swapRouter = await fakeSwapRouterFixture()
+    market = await fakePrePOMarketFixture()
```
XavierEkkel, line 42: nit: move this right before `saveSnapshot`, since other variables should be declared/assigned right before they are used

apps/smart-contracts/core/test/ArbitrageBroker.test.ts
```diff
@@ -96,4 +108,50 @@ describe('=> ArbitrageBroker', () => {
       expect(tx).to.emit(arbitrageBroker, 'MarketValidityChanged').withArgs(user.address, true)
     })
   })
+
+  describe('mintAndSell', () => {
```
XavierEkkel, line 112: nit: 
```suggestion
  describe('# mintAndSell', () => {
```

apps/smart-contracts/core/test/ArbitrageBroker.test.ts
```diff
@@ -96,4 +108,50 @@ describe('=> ArbitrageBroker', () => {
...
+    })
+
+    it('reverts if market not valid', async () => {
+      const invalidMarket = await fakePrePOMarketFixture()
+      expect(await arbitrageBroker.isMarketValid(invalidMarket.address)).to.eq(false)
+
+      await expect(
+        arbitrageBroker.connect(governance).mintAndSell(invalidMarket.address, tradeParams)
+      )
+        .revertedWithCustomError(arbitrageBroker, 'InvalidMarket')
```
XavierEkkel, line 132: is there a way to use the error from the interface itself rather than a manual string?

XavierEkkel (approved):
LGTM after comments

apps/smart-contracts/core/contracts/ArbitrageBroker.sol
ramenforbreakfast: I will put braces since the linter prefers it go into the next line due to char length limit.

apps/smart-contracts/core/test/ArbitrageBroker.test.ts
ramenforbreakfast: For the chai matchers, from their documentation, it does not seem you can do that. Similarly for events, the event must be specified via a string when using it in the context of a chai matcher.

We do extract the event from the interface types for our event filtering helpers using `ethers.js`

PR #731, opened 2022-12-27
feat: add swap and profit logic to mintAndSell

ramenforbreakfast:
So while writing out the logic today, I actually started with the profit checking, since swap logic was something I didn't think was going to be too difficult and we had written tests before in `DepositTradeHelper`, so I didn't expect any surprises.

One change I realized is that we can't use a modifier, because we want the trade functions to return the profit, for the script to use for its estimates (since these are non-state changing simulation calls). If we put profit checking in a modifier, we'd be grabbing balances twice in the function itself (to calculate the profit to return), and in the modifier (to revert if there was a loss). Since modifiers cant pass values to the function, it made no sense to do balance checking twice so I ended up just doing it inside the function itself.

The real problem today was that since we need to know that the function correctly a) reverts upon a loss and b) returns the difference in contract balance, i needed a way to mock a change in the contract balance within the context of the transaction.

I found that Smock has a way to do this via `returnsAtCall` which lets you specify a return value at a specific call, the plan was to have a value representing the balance before the trade `returnsAtCall(0, <value>)` and after the trade `returnsAtCall(1, <value>)`.

This however, did not work as expected, without going too in-depth, the calls simply were not returning the expected values at the times I specified. Many other things made it incredibly difficult to debug, I tried abandoning mocks and just using fakes, I spent hours directly trying to analyze the smock call history, but I ended up not really figuring out why this was happening.

Eventually I came upon a finding that if I ran the transaction using a static call (using `callStatic`), the `returnsAtCall` function returned the correct values as expected.

Because I have already spent the entire day trying to get this to work and debugging, I am going to resort to switching the `mintAndSell` tests to use static calls instead of state changing transactions.

Technically, this should still be an accurate representation of checking the contract logic, but I know it is not ideal.


apps/smart-contracts/core/test/ArbitrageBroker.test.ts
```diff
@@ -133,25 +144,122 @@ describe('=> ArbitrageBroker', () => {
...
+      expect(longToken.approve).to.be.calledWith(swapRouter.address, tradeParams.longShortAmount)
+    })
+
+    it('sells long token for sale after approval', async () => {
+      await arbitrageBroker.connect(governance).callStatic.mintAndSell(market.address, tradeParams)
+
+      expect(swapRouter.exactInputSingle).to.be.calledAfter(longToken.approve)
+      const swapRouterCallArgs = swapRouter.exactInputSingle
+        .atCall(0)
+        .callHistory[0].args[0].slice(0, 8)
```
XavierEkkel, line 202: rather than `8`, use a `const SWAP_ARG_COUNT = 8` variable, and reuse this in the short token test as well

apps/smart-contracts/core/test/ArbitrageBroker.test.ts
```diff
@@ -133,25 +144,122 @@ describe('=> ArbitrageBroker', () => {
...
+    })
+
+    it('approves short token for sale after selling long', async () => {
+      await arbitrageBroker.connect(governance).callStatic.mintAndSell(market.address, tradeParams)
+
+      expect(shortToken.approve).to.be.calledAfter(swapRouter.exactInputSingle)
+      expect(shortToken.approve).to.be.calledWith(swapRouter.address, tradeParams.longShortAmount)
+    })
+
+    it('sells short token for sale after approval', async () => {
```
XavierEkkel, line 225: nit:
```suggestion
    it('sells short token after approval', async () => {
```

apps/smart-contracts/core/test/ArbitrageBroker.test.ts
```diff
@@ -133,25 +144,122 @@ describe('=> ArbitrageBroker', () => {
         .withArgs(invalidMarket.address)
     })
 
+    it('reverts if no profit earned', async () => {
```
XavierEkkel, line 147: nit: (for clarity)
```suggestion
    it('reverts if profit exactly 0', async () => {
```

apps/smart-contracts/core/test/ArbitrageBroker.test.ts
```diff
@@ -133,25 +144,122 @@ describe('=> ArbitrageBroker', () => {
...
+    })
+
+    it('approves long token for sale after minting', async () => {
+      await arbitrageBroker.connect(governance).callStatic.mintAndSell(market.address, tradeParams)
+
+      expect(longToken.approve).to.be.calledAfter(market.mint)
+      expect(longToken.approve).to.be.calledWith(swapRouter.address, tradeParams.longShortAmount)
+    })
+
+    it('sells long token for sale after approval', async () => {
```
XavierEkkel, line 196: nit:
```suggestion
    it('sells long token after approval', async () => {
```

apps/smart-contracts/core/test/ArbitrageBroker.test.ts
```diff
@@ -133,25 +144,122 @@ describe('=> ArbitrageBroker', () => {
         .withArgs(invalidMarket.address)
     })
```
XavierEkkel, line 146: add tests for reverting if either of the swap calls revert

apps/smart-contracts/core/contracts/ArbitrageBroker.sol
```diff
@@ -34,8 +34,42 @@ contract ArbitrageBroker is IArbitrageBroker, SafeAccessControlEnumerable {
...
+        address(this),
+        tradeParams.deadline,
+        tradeParams.longShortAmount,
+        tradeParams.collateralLimitForLong,
+        0
+      );
+    _swapRouter.exactInputSingle(longTradeParams);
+    ILongShortToken shortToken = market.getShortToken();
+    shortToken.approve(address(_swapRouter), tradeParams.longShortAmount);
+    ISwapRouter.ExactOutputSingleParams memory shortTradeParams = ISwapRouter
```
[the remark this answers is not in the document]
ramenforbreakfast: Yeah not sure why I did that, my bad.

ramenforbreakfast:
I have addressed all comments including:
- Addressing mistake of using wrong swap for short token sale, and moving swap logic to `private` function.
- Remove`to` and `to.be` from tests
- Moved swap args in tests to `const` values and swap length
- Added extra tests for reversion when either swap reverts.

apps/smart-contracts/core/contracts/ArbitrageBroker.sol
```diff
@@ -63,4 +81,27 @@ contract ArbitrageBroker is IArbitrageBroker, SafeAccessControlEnumerable {
...
+      : tradeParams.collateralLimitForShort;
+    ISwapRouter.ExactInputSingleParams memory exactInputSingleParams = ISwapRouter
+      .ExactInputSingleParams(
+        address(longShortToken), // tokenIn
+        address(_collateral), // tokenOut
+        POOL_FEE_TIER,
+        address(this), // recipient
+        tradeParams.deadline,
+        tradeParams.longShortAmount, // amountIn
+        amountOutMinimum, // amountOutMinimum
```
XavierEkkel, line 102: nit:
```suggestion
        amountOutMinimum,
```

XavierEkkel (approved):
LGTM after nit

PR #733, opened 2022-12-28
feat: start with swap logic in buyAndRedeem

ramenforbreakfast:
For `buyAndRedeem`, the collateral approval is to the `swapRouter`, which will not change, so we can do an unlimited approval ahead of time in the constructor.


apps/smart-contracts/core/test/ArbitrageBroker.test.ts
```diff
@@ -59,26 +61,32 @@ describe('=> ArbitrageBroker', () => {
...
       tradeParams.longShortAmount,
       tradeParams.collateralLimitForShort,
       0,
     ]
+    correctSellLongArgs = [correctBuyLongArgs[1]]
+      .concat(correctBuyLongArgs[0])
+      .concat(correctBuyLongArgs.slice(2))
+    correctSellShortArgs = [correctBuyShortArgs[1]]
+      .concat(correctBuyShortArgs[0])
+      .concat(correctBuyShortArgs.slice(2))
```
XavierEkkel, line 89: nit: this is overly complex, can just input collateral/shortToken/longToken directly instead of a particular index

XavierEkkel (approved):
LGTM after nits

PR #737, opened 2022-12-29
core:fix: remove init args from DepositRecord

ramenforbreakfast:
This seems to have been missed from the backlog item merge last week, these should be removed since `DepositRecord` does not take parameters anymore.

I realize this was most likely not caught because we have been forgetting to append `core:` in front of our PRs!


XavierEkkel (approved):
LGTM

PR #739, opened 2022-12-29
core:feat: add fixtures for deploying and connecting to a pool on a forked mainnet

ramenforbreakfast:


apps/smart-contracts/core/test/fixtures/UniswapFixtures.ts
```diff
@@ -1,8 +1,21 @@
...
+import { ethers } from 'hardhat'
 
 export async function fakeSwapRouterFixture(): Promise<FakeContract<BaseContract>> {
   const fakeContract = await smock.fake(SWAP_ROUTER_ABI)
   return fakeContract
 }
+
+export async function attachUniV3Factory(address: string): Promise<Contract> {
+  const contract = await ethers.getContractAt(UNIV3_FACTORY_ABI, address)
+  return contract
```
XavierEkkel, line 15:
```suggestion
  return await ethers.getContractAt(UNIV3_FACTORY_ABI, address)
```
?
(maybe need brackets around the await part)

same for other functions

XavierEkkel (approved):
LGTM after comment

apps/smart-contracts/core/test/fixtures/UniswapFixtures.ts
ramenforbreakfast: Ok yeah I figured out what was happening here, we don't need to `await` the expression here and this doesn't need to be labeled `async` since we are just directly returning the promise for fakes.

PR #740, opened 2022-12-29
core:feat: add fixtures to easily mint positions and collateral

ramenforbreakfast:


apps/smart-contracts/core/helpers/index.ts
```diff
@@ -26,3 +34,91 @@ export async function createMarket(marketParams: CreateMarketParams): Promise<Cr
     hash: events[0].args.longShortHash,
   }
 }
+
+export async function baseTokenNeededForMinting(
```
XavierEkkel, line 38: not clear what this function is supposed to be
is it supposed to return the amount of input base token required in order to receive a specified output collateral amount (the argument)? if so I think the calculation is wrong - base token amount needs to be higher by accounting for the fee (i.e. `collateralAmount / (1 - fee)`), it shouldn't be subtracting

clearer naming would also be `getBaseTokenAmount` with argument `collateralAmount`, so that we know it doesn't change state

apps/smart-contracts/core/helpers/index.ts
```diff
@@ -26,3 +34,91 @@ export async function createMarket(marketParams: CreateMarketParams): Promise<Cr
...
+  collateral: Collateral | MockContract<Collateral>,
+  amount: BigNumber
+): Promise<BigNumber> {
+  const fee = amount.mul(await collateral.getDepositFee()).div(FEE_DENOMINATOR)
+  return amount.sub(fee)
+}
+
+export async function depositCapCheck(
+  recipient: string,
+  amount: BigNumber,
```
XavierEkkel, line 48: for all amount arguments in this file, please be specific about the denomination (baseToken or collateral)

apps/smart-contracts/core/harnesses/base.ts
```diff
@@ -27,4 +29,29 @@ export abstract class Base {
...
+
+  public lsToCollateral(
+    funder: SignerWithAddress,
+    amount: BigNumber,
+    marketSuffix: string
+  ): Promise<BigNumber> {
+    return lsToCollateral(funder, amount, this.markets[marketSuffix])
+  }
+
+  public baseTokenToLS(
```
XavierEkkel, line 49: let's rename these functions to be clearer about their functionality (e.g. using words like convert/mint/redeem/sell/buy)
are they actually performing state changes, or just returning a value, or both?

same for the other file

apps/smart-contracts/core/helpers/index.ts
```diff
@@ -26,3 +34,91 @@ export async function createMarket(marketParams: CreateMarketParams): Promise<Cr
...
+
+export async function baseTokenNeededForMinting(
+  collateral: Collateral | MockContract<Collateral>,
+  amount: BigNumber
+): Promise<BigNumber> {
+  const fee = amount.mul(await collateral.getDepositFee()).div(FEE_DENOMINATOR)
+  return amount.sub(fee)
+}
+
+export async function depositCapCheck(
```
XavierEkkel, line 46: nit: name functions as actions rather than statements
```suggestion
export async function checkDepositCap(
```

apps/smart-contracts/core/helpers/index.ts
ramenforbreakfast: You are correct, I have rewritten this function to fix this and renamed it as you've requested.

ramenforbreakfast: You are correct, I have rewritten this function to fix this and renamed it as you've requested.

ramenforbreakfast:
I have addressed all naming requests, changing the names of functions to be more clear and represent actions.

In addition, I have added fee calculation to the base token estimation function, I originally did not test fee inclusion because it would have involved writing role assignment code. 

I wrote some temporary bits (not in this PR, just for testing locally) to test fees and also identified two bugs that I have now fixed:
- The collateral minting function should be using `Transfer` to detect how much was minted, not `Mint`, since `Mint` contains how much base token was spent. So the `Deposit` event finder has been replaced with a generic `ERC20` `Transfer` event searcher.
- There was a bug that took me a while to figure out, realized that I had messed up the way I was searching for events. The way it currently works, grabs all events since block 0, and orders it with the earliest one first. We want the default behavior to grab the latest event, so `startBlock` should be `latest` as well, instead of just `endBlock`.

apps/smart-contracts/core/helpers/index.ts
```diff
@@ -26,3 +34,91 @@ export async function createMarket(marketParams: CreateMarketParams): Promise<Cr
     hash: events[0].args.longShortHash,
   }
 }
+
+export async function getBaseTokenAmount(
+  collateral: Collateral | MockContract<Collateral>,
+  collateralAmount: BigNumber
+): Promise<BigNumber> {
+  const factorAfterFee = BigNumber.from(FEE_DENOMINATOR).sub(await collateral.getDepositFee())
+  return collateralAmount.mul(factorAfterFee).div(FEE_DENOMINATOR)
```
XavierEkkel, line 43: this still looks incorrect
it should be:
```suggestion
  const afterFeeFactor = (BigNumber.from(FEE_DENOMINATOR).sub(await collateral.getDepositFee()).div(FEE_DENOMINATOR)
  return collateralAmount.div(afterFeeFactor)
```

apps/smart-contracts/core/helpers/index.ts
```diff
@@ -26,3 +34,91 @@ export async function createMarket(marketParams: CreateMarketParams): Promise<Cr
...
+}
+
+export async function checkDepositCap(
```
XavierEkkel, line 46: this is only checking user deposit cap, but it should also check global net deposit cap

apps/smart-contracts/core/helpers/index.ts
```diff
@@ -26,3 +34,91 @@ export async function createMarket(marketParams: CreateMarketParams): Promise<Cr
...
+        await depositHook.getDepositRecord()
+      )) as DepositRecord
+      const userDepositCap = await depositRecord.getUserDepositCap()
+      const currentDepositTotal = await depositRecord.getUserDepositAmount(recipient)
+      const newDepositTotal = currentDepositTotal.add(baseTokenAmount)
+      if (newDepositTotal.gt(userDepositCap)) {
+        throw new Error(
+          `${recipient} new deposit total of ${formatEther(
+            newDepositTotal
+          )} exceeds the cap ${formatEther(userDepositCap)}`
```
XavierEkkel, line 70:
```suggestion
          )} exceeds the user cap ${formatEther(userDepositCap)}`
```

apps/smart-contracts/core/helpers/index.ts
```diff
@@ -26,3 +34,91 @@ export async function createMarket(marketParams: CreateMarketParams): Promise<Cr
...
+  funder: SignerWithAddress,
+  recipient: SignerWithAddress,
+  baseTokenAmount: BigNumber,
+  market: PrePOMarket | MockContract<PrePOMarket>
+): Promise<BigNumber> {
+  const collateral = (await ethers.getContractAt(
+    'Collateral',
+    await market.getCollateral()
+  )) as Collateral
+  const baseTokenNeeded = await getBaseTokenAmount(collateral, baseTokenAmount)
```
XavierEkkel, line 115: shouldn't `collateralAmount` be passed in rather than `baseTokenAmount` (which wouldn't make sense)?
```suggestion
  const baseTokenAmount = await getBaseTokenAmount(collateral, collateralAmount)
```

apps/smart-contracts/core/helpers/index.ts
ramenforbreakfast: Ah sorry, I was writing this as if we were writing `getCollateralAmount`

apps/smart-contracts/core/helpers/index.ts
ramenforbreakfast: This would conflict with `baseTokenAmount` that is an existing parameter

apps/smart-contracts/core/helpers/index.ts
ramenforbreakfast: I have fixed this so that it correctly takes in the `collateralAmount` that someone wants to mint, and calculates the amount of base token needed. So the above wouldn't be correct either, but I have modified it. Also keep in mind it seems like you do not like doing a `.mul(X).div(Y)` as the final return value, and prefer a singular `.mul` or `.div` with a factor.

Understand that we can't do that because we are performing `BigNumber` integer math, and are doing so to ensure the value returned is the same as what would be returned within the contract. `afterFeeFactor` in the above code would return 0 every time.

ramenforbreakfast:
I have updated the helpers with multiple fixes:
- Fixed inconsistencies with inputs and outputs of functions (i.e. `collateralAmount` should be the input for `mintCollateralFromBaseToken`, not `baseTokenAmount`).
- Fixed `getBaseTokenAmount` math, was incorrectly written as if a `collateralAmount` was the base token being deposited, most likely due to another mistake in `mintLSFromBaseToken`, where the input of `getBaseTokenAmount` is incorrectly `baseTokenAmount`.
- Added global deposit checking. To improve readability, separated deposit checks into 3 smaller functions.

apps/smart-contracts/core/helpers/index.ts
```diff
@@ -39,47 +39,85 @@ export async function getBaseTokenAmount(
   collateral: Collateral | MockContract<Collateral>,
   collateralAmount: BigNumber
 ): Promise<BigNumber> {
-  const factorAfterFee = BigNumber.from(FEE_DENOMINATOR).sub(await collateral.getDepositFee())
-  return collateralAmount.mul(factorAfterFee).div(FEE_DENOMINATOR)
+  const afterFeeFactor = BigNumber.from(FEE_DENOMINATOR).sub(await collateral.getDepositFee())
+  return collateralAmount.mul(FEE_DENOMINATOR).div(afterFeeFactor)
+}
+
+export async function checkDepositRecordExistence(
```
XavierEkkel, line 46: since this returns true or false, a more appropriate name would be `depositRecordExists()`

apps/smart-contracts/core/helpers/index.ts
```diff
@@ -39,47 +39,85 @@ export async function getBaseTokenAmount(
...
+    )
+  }
+}
+
+export async function checkGlobalNetDepositCap(
+  recipient: string,
+  baseTokenAmount: BigNumber,
+  depositRecord: DepositRecord | MockContract<DepositRecord>
+): Promise<void> {
+  const globalNetDepositCap = await depositRecord.getGlobalNetDepositCap()
```
XavierEkkel, line 83: nit: move below `const newGlobalTotal` so it's closest to where it's used

apps/smart-contracts/core/helpers/index.ts
```diff
@@ -39,47 +39,85 @@ export async function getBaseTokenAmount(
...
+}
+
+export async function checkUserDepositCap(
+  recipient: string,
+  baseTokenAmount: BigNumber,
+  depositRecord: DepositRecord | MockContract<DepositRecord>
+): Promise<void> {
+  const userDepositCap = await depositRecord.getUserDepositCap()
+  const currentUserTotal = await depositRecord.getUserDepositAmount(recipient)
+  const newUserTotal = currentUserTotal.add(baseTokenAmount)
```
XavierEkkel, line 68: cap amounts are recorded based on `amountAfterFee` not `amountBeforeFee`
so `collateralAmount` should be used rather than `baseTokenAmount`

same for `checkGlobalNetDepositCap()`

apps/smart-contracts/core/helpers/index.ts
```diff
@@ -90,29 +128,34 @@ export async function mintCollateralFromBaseToken(
...
-  await market.connect(funder).mint(collateralAmount)
-  return collateralAmount
+  await collateral.connect(funder).approve(market.address, lsAmount)
+  await market.connect(funder).mint(lsAmount)
+  /**
+   * Can just directly return lsAmount since there will never be a fee
+   * for minting L/S positions and the amount returned is guaranteed to
+   * be the same.
+   */
+  return lsAmount
```
XavierEkkel, line 145: given this, there isn't any point to returning anything. can remove the return statement here + in `base.ts`

apps/smart-contracts/core/helpers/index.ts
```diff
@@ -90,29 +128,34 @@ export async function mintCollateralFromBaseToken(
 
 export async function mintLSFromCollateral(
```
XavierEkkel, line 129: add comment above this function re: `lsAmount` and `collateralAmount` being one and the same

apps/smart-contracts/core/helpers/index.ts
```diff
@@ -39,47 +39,85 @@ export async function getBaseTokenAmount(
...
+    const depositRecord = (await ethers.getContractAt(
+      'DepositRecord',
+      await depositHook.getDepositRecord()
+    )) as DepositRecord
+    await checkUserDepositCap(recipient, baseTokenAmount, depositRecord)
+    await checkGlobalNetDepositCap(recipient, baseTokenAmount, depositRecord)
   }
 }
 
 export async function mintCollateralFromBaseToken(
```
XavierEkkel, line 114: add comment above that this assumes the funder has enough base token such that it will cover the specified collateral amount after deposit fees are taken into account

apps/smart-contracts/core/helpers/index.ts
```diff
@@ -39,47 +39,85 @@ export async function getBaseTokenAmount(
...
+      `${recipient} new global deposit total of ${formatEther(
+        newGlobalTotal
+      )} exceeds the global cap ${formatEther(globalNetDepositCap)}`
+    )
+  }
 }
 
 export async function checkDepositCap(
   recipient: string,
   baseTokenAmount: BigNumber,
```
XavierEkkel, line 97: should be `collateralAmount`

XavierEkkel (approved):
LGTM after comments

apps/smart-contracts/core/helpers/index.ts
ramenforbreakfast: ah good catch

apps/smart-contracts/core/helpers/index.ts
ramenforbreakfast: This should actually be `amountAfterFee` of BaseToken, since `Collateral` can differ in decimal precision from `BaseToken`, I have fixed the function to accomodate for this.

PR #741, opened 2022-12-30
core:refactor: use types for Uniswap external contracts

ramenforbreakfast:
So the rationale for this switch is that originally we could just directly reference the ABIs to instantiate contracts, but since we now need to write code that swaps or mints/adds liquidity, we want to have access to Uniswap's struct types. We could redefine them ourselves, but this would be cumbersome.

However, this proved to be much much more challenging than I originally anticipated. Technically, `typechain` supports generating types from external contracts out of the box. However, in addition to problems locally with my `node_modules` that I eventually resolved, I figured out the actual problem I was facing was a bug in `typechain` that has a fix that is not merged in yet https://github.com/dethcrypto/TypeChain/issues/772.

This issue is a problem with their common directory resolution code which causes problems when sourcing contracts outside of the current project directory.

Rolling back `typechain` is not an option, because it seems that this bug was introduced when they revamped the way they organize types to prevent namespace conflicts in `6.0.0`. This is actually a big benefit for why we upgraded, since we no longer need to provide the full path when specifying some types, we used to have _N_ separate imports for every contract we used in a test, instead of 1-2 now.

To rollback before this change would mean un-refactoring all of those import changes which I believe to be unacceptable.

I could not find a solution to this, and have instead opted for directly copying the artifacts from the package over to our `core` directory in the interim, until this issue is fixed in `typechain`.

In addition, I have added correct typing to all Uniswap fixtures we used (not that many), and removed generic `Contract` type usage from the 2 test suites that use Uniswap contracts.


apps/smart-contracts/core/test/fixtures/UniswapFixtures.ts
```diff
@@ -1,18 +1,18 @@
+import { ethers } from 'hardhat'
 import { FakeContract, smock } from '@defi-wonderland/smock'
 import { abi as UNIV3_FACTORY_ABI } from '@uniswap/v3-core/artifacts/contracts/UniswapV3Factory.sol/UniswapV3Factory.json'
 import { abi as UNIV3_POOL_ABI } from '@uniswap/v3-core/artifacts/contracts/UniswapV3Pool.sol/UniswapV3Pool.json'
 import { abi as SWAP_ROUTER_ABI } from '@uniswap/v3-periphery/artifacts/contracts/SwapRouter.sol/SwapRouter.json'
```
XavierEkkel, line 5: should these still be imported?

XavierEkkel (approved):
LGTM after question

PR #742, opened 2022-12-30
core:feat: add position manager entry and fixture

ramenforbreakfast:


apps/smart-contracts/core/test/ArbitrageTrading.test.ts
```diff
@@ -79,6 +84,9 @@ describe('=> Functional Testing with Arbitrage Trades', () => {
         POOL_FEE_TIER
       )
     )
+    positionManager = await attachNonfungiblePositionManager(
+      getPrePOAddressForNetwork('UNIV3_POSITION_MANAGER', 'arbitrumOne')
```
XavierEkkel, line 88: would be nicer if we could use constants, e.g. `ContractAddresses.UNIV3_POSITION_MANAGER` and `Networks.ARBITRUM_ONE`
but can leave this as a backlog item

XavierEkkel (approved):
LGTM

PR #743, opened 2022-12-30
core:feat: add uniswap sdk

ramenforbreakfast:
Originally, I had actually not planned to use the SDK, because the main features they advertise are for directly interacting with pools. IMO, their helpers were not actually that useful and it will be easier for our case to simulate calls directly through `ArbitrageBroker.sol`.

However, while beginning to write out LP test fixtures, I did some digging and some of the internal calculation helpers in the SDK will be quite useful out-of-the-box and will mean not having to re-define or modify these utilities like we did with Trident.


XavierEkkel (approved):
LGTM

PR #744, opened 2022-12-31
core:feat: add create2 helpers for prod and testing

ramenforbreakfast:
While I would have liked to finish working on code for provisioning liquidity, I could not test it properly because the token ordering was not right (we need `L/S` => `token0` and `Collateral` => `token1` ).

This forced me to look into ways we could address this issue for testing, without essentially implementing what we will be using for production for `create2`.

I was hoping I could achieve this through `smock` since I originally had written the test harness to deploy mock L/S tokens when a mock `PrePOMarket`, but this is not possible. Faking addresses is only possible with fakes.

This meant I had to:
- Modify the mock market deployer to deploy actual L/S tokens, not mock ones. This is not a problem since none of our tests need smocked L/S tokens. We need to deploy these tokens separately, and not through `PrePOMarketFactory`, since that would mean using non-smock `PrePOMarket`'s, taking away from flexibility within tests.
- Use a basic `Create2Deployer` contract taken from online examples to deploy L/S tokens with `create2`, this is mandatory since `create2` is a solidity function and must be invoked via a contract.
- Write out basic helper functions that we can use later for generating addresses that are less than the `Collateral` address.

All of this proved far more difficult than anticipated, but we will now be on a much better footing when we design our prod deployment process for deterministic addresses.

Also nice thing is that while it doesn't mean it will be true in production, finding addresses for the L/S tokens is way faster than I thought, the process takes only a few seconds to find addresses for both of them.

Note: `Create2Deployer` has a `transferOwnership` pass through since L/S tokens need to be owned by the market.


apps/smart-contracts/core/contracts/test/Create2Deployer.sol
```diff
@@ -0,0 +1,28 @@
+// SPDX-License-Identifier: AGPL-3.0
+pragma solidity 0.8.7;
```
XavierEkkel, line 2:
```suggestion
pragma solidity =0.8.7;
```

apps/smart-contracts/core/contracts/test/Create2Deployer.sol
```diff
@@ -0,0 +1,28 @@
...
+
+  function deploy(bytes memory bytecode, bytes32 _salt) external {
+    address addr;
+    assembly {
+      addr := create2(0, add(bytecode, 0x20), mload(bytecode), _salt)
+      if iszero(extcodesize(addr)) {
+        revert(0, 0)
+      }
+    }
+
```
XavierEkkel, line 15: nit:
```suggestion
```

apps/smart-contracts/core/contracts/test/Create2Deployer.sol
```diff
@@ -0,0 +1,28 @@
+// SPDX-License-Identifier: AGPL-3.0
+pragma solidity 0.8.7;
+
+contract Create2Deployer {
+  event Deploy(address addr);
+
+  function deploy(bytes memory bytecode, bytes32 _salt) external {
```
XavierEkkel, line 7: nit: (assuming this compiles)
```suggestion
  function deploy(bytes memory bytecode, bytes32 salt) external {
```

apps/smart-contracts/core/contracts/test/Create2Deployer.sol
```diff
@@ -0,0 +1,28 @@
...
+    address addr;
+    assembly {
```
XavierEkkel, line 9: add comment above explaining what this assembly code is doing

apps/smart-contracts/core/test/fixtures/LongShortTokenFixture.ts
```diff
@@ -15,6 +16,20 @@ export async function LongShortTokenAttachFixture(tokenAddress: string): Promise
   return factory.attach(tokenAddress) as LongShortToken
 }
 
+export async function create2LongShortTokenFixture(
+  tokenName: string,
+  tokenSymbol: string,
+  deployerFactory: Create2Deployer,
```
XavierEkkel, line 22: nit:
```suggestion
  create2Deployer: Create2Deployer,
```

apps/smart-contracts/core/harnesses/mock.ts
```diff
@@ -60,26 +60,36 @@ export class MockCore extends Base {
   }
 
   public async createAndAddMockMarket(
-    deployer: SignerWithAddress,
     tokenNameSuffix: string,
     tokenSymbolSuffix: string,
-    marketParams: PrePOMarketParams
+    marketParams: PrePOMarketParams,
+    deployerFactory: Create2Deployer
```
XavierEkkel, line 66: nit:
```suggestion
    create2Deployer: Create2Deployer
```

apps/smart-contracts/core/contracts/test/Create2Deployer.sol
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  function transferOwnership(address ownableContract, address newOwner)
+    external
+    returns (bool)
+  {
+    (bool success, ) = ownableContract.call(
+      abi.encodeWithSignature("transferOwnership(address)", newOwner)
+    );
+    return success;
```
XavierEkkel, line 26: get rid of `return` and instead just revert if not success? so that we don't need to add an assertion within the test files to check it was successful

packages/prepo-hardhat/src/smart-contracts-utils.ts
```diff
@@ -163,6 +170,24 @@ function mineBlock(provider: providers.Web3Provider, timestamp: number): Promise
...
+function generateAddressLessThan(
+  deployer: string,
+  initCode: BytesLike,
+  upperBound: string
+): string {
+  let currAddress: string
+  let randomString: string
+  let currSalt: string
+  const hashedInitCode = keccak256(initCode)
+  const upperBoundBN = BigNumber.from(upperBound)
```
XavierEkkel, line 182: nit: move these above the `let` variables

packages/prepo-hardhat/src/smart-contracts-utils.ts
```diff
@@ -163,6 +170,24 @@ function mineBlock(provider: providers.Web3Provider, timestamp: number): Promise
...
+  do {
+    randomString = (Math.random() + 1).toString(36).substring(7)
```
XavierEkkel, line 184: add comment explaining where these magic numbers are coming from

packages/prepo-hardhat/src/smart-contracts-utils.ts
```diff
@@ -163,6 +170,24 @@ function mineBlock(provider: providers.Web3Provider, timestamp: number): Promise
...
 }
 
+function generateAddressLessThan(
+  deployer: string,
+  initCode: BytesLike,
+  upperBound: string
+): string {
+  let currAddress: string
+  let randomString: string
+  let currSalt: string
```
XavierEkkel, line 180: nit: I think these could be declared within the dowhile?

packages/prepo-hardhat/src/smart-contracts-utils.ts
```diff
@@ -163,6 +170,24 @@ function mineBlock(provider: providers.Web3Provider, timestamp: number): Promise
   return provider.send('evm_mine', [timestamp])
 }
 
+function generateAddressLessThan(
+  deployer: string,
+  initCode: BytesLike,
+  upperBound: string
```
XavierEkkel, line 176: nit: (for clarity)
```suggestion
  upperBoundAddress: string
```

packages/prepo-hardhat/src/smart-contracts-utils.ts
```diff
@@ -163,6 +170,24 @@ function mineBlock(provider: providers.Web3Provider, timestamp: number): Promise
   return provider.send('evm_mine', [timestamp])
 }
 
+function generateAddressLessThan(
+  deployer: string,
+  initCode: BytesLike,
```
XavierEkkel, line 175: add comment explaining what initCode is

apps/smart-contracts/core/contracts/test/Create2Deployer.sol
```diff
@@ -0,0 +1,28 @@
+// SPDX-License-Identifier: AGPL-3.0
+pragma solidity 0.8.7;
+
+contract Create2Deployer {
+  event Deploy(address addr);
```
XavierEkkel, line 5: since this is just for tests, is this event even necessary?

XavierEkkel (approved):
LGTM after comments

apps/smart-contracts/core/contracts/test/Create2Deployer.sol
ramenforbreakfast: This technically can be removed, since we now know it works as expected, it was needed during development since we needed to obtain the address of the contract (since it might have not been what we expected).

packages/prepo-hardhat/src/smart-contracts-utils.ts
ramenforbreakfast: No, because then these variables would only live within the do while loop, which wouldn't make it available for the `while` check, and returning the eventual salt and address generated.

apps/smart-contracts/core/contracts/test/Create2Deployer.sol
ramenforbreakfast: I have removed this and refactored the address generation to also return the address generated so we do not have to retrieve it via an event when the contract is deployed.

PR #746, opened 2023-01-04
feat: initial copy over of existing Thales Staking Reward work

ramenforbreakfast:
So while I did not try to integrate it with our existing monorepo environment, there were multiple difficulties that I encountered that made this take much longer than expected, mostly due to my inexperience with how `npm` and `yarn` work.

Initially, when I tried to copy over just the necessary content from the Thales repo, using Thales's own `package.json`, `yarn` would try to reconcile the packages in the Thales `package.json` and the monorepo's root `package.json`. This caused many issues with package conflicts that I gave up quickly on trying to fix.

I then realized, you can tell `package.json` to ignore a folder, so thats why you will see in our root `package.json`, it now specifies `token` and `core` to ignore the new `stakingV1` folder.

Then I encountered issues trying to just operate with the isolated `node_modules` generated in `stakingV1` directory itself, because I kept trying to use `npx` instead of `yarn` since they seem to not use `yarn` to install their packages. This works fine if you only have one `package.json`, but if you try to run `hardhat` with `npx`, it tries to read the parent `hardhat` in the parent `node_modules`, and errors out.

I eventually figured out installing packages using `yarn` and running `hardhat` using `yarn` will correctly direct it to the local `node_modules` and not the parent one.

**This copied over directory includes tests that were already modified earlier to verify our initial deployment, and contract changes from the last time we deployed, none of the changes spoken about since then have been introduced**

This is a diff containing what was changed for our deployments from the source contract for reference.
```
13c13
< contract LPStakingRewards is Initializable, ProxyOwned, ProxyReentrancyGuard, ProxyPausable {
---
> contract StakingRewards is Initializable, ProxyOwned, ProxyReentrancyGuard, ProxyPausable {
83c83
<     function stake(uint256 amount) external nonReentrant notPaused updateReward(msg.sender) {
---
>     function stake(uint256 amount) public nonReentrant notPaused updateReward(msg.sender) {
105a106,110
>     }
> 
>     function compound() public {
>         getReward();
>         stake(stakingToken.balanceOf(msg.sender));
187c192
< }
\ No newline at end of file
---
> }
```


XavierEkkel:
`[apps/smart-contracts/token/.eslintrc.js → apps/smart-contracts/stakingV1/.eslintrc.js]`
why is this being moved?

ramenforbreakfast:
> `[apps/smart-contracts/token/.eslintrc.js → apps/smart-contracts/stakingV1/.eslintrc.js]` why is this being moved?

This was accidental, was meant to copy, not move. We need to add `eslintrc.js` because it is not in the original Thales repo and if absent, breaks our CI/CD

XavierEkkel (approved):
LGTM if checks pass

PR #748, opened 2023-01-04
feat: modify staking rewards constructor

ramenforbreakfast:


XavierEkkel (approved):
LGTM after also updating constructor in deploy script

PR #749, opened 2023-01-04
feat: add changes to notifyRewardAmount and partial test suite

ramenforbreakfast:
I have added extensive comments throughout the test suite to give context to the changes I've made.

Because there are basically no unit-style testcases in the Thales suite, almost all far more involved complicated functional end-to-end testing suites that would take too much effort to unpack, we are going to be writing our own **for the 2 functions that we are modifying**, this is probably for the best as a modifying existing tests would be not great for a reviewer.

Their existing test case for the test suite we've kept (because it is entirely focused on the single staking contract), is 2 reverts for when there are no existing stakers. Since we want to focus more on cases with existing stakers, I only added 3 extra ones just for sanity and to ensure the math still checks out.

As for when stakers exist, I have written the section for when a period has ended, but for adding rewards during a period, those tests are more complex and will be in a following PR to reduce PR size.

You will notice I have added additional helpers to ensure tests are not too verbose. Have also added a `setNextTimestamp` helper that did not exist, for reasons explained in the relevant testcase comment below.


ramenforbreakfast:
Closing since no one has reviewed this yet, and I figured out the problem with my VSCode preventing automatic linting of files, which prevented me from applying our formatting to Thales files.

This is especially important for reviewing test suites since writing them without linting would have made it difficult for someone to review.

PR #750, opened 2023-01-05
chore: formatting on Thales js files

ramenforbreakfast:
So last night/this morning, I couldn't get linting of files to work inside my environment, I then realized it was a problem with my VSCode and it being confused by the config inside `external/yield`.

I abandoned trying to debug it, wanting to focus on having tests to submit for `notifyForReward`, but I eventually got it to work. So I closed PR #749 which contained initial tests written for `notifyRewardAmount`, and will rebase and resubmit it off this one so staking reward test submissions will be linted from now on.


XavierEkkel (approved):
LGTM

PR #751, opened 2023-01-05
feat: modify notifyRewardAmount and revise existing tests

ramenforbreakfast:
I have added extensive comments throughout the test suite to give context to the changes I've made.

Because there are basically no unit-style testcases in the Thales suite, almost all far more involved complicated functional end-to-end testing suites that would take too much effort to unpack, we are going to be writing our own **for the 2 functions that we are modifying**, this is probably for the best as a modifying existing tests would be not great for a reviewer.

Their existing test case for the test suite we've kept (because it is entirely focused on the single staking contract), is 2 reverts for when there are no existing stakers. Since we want to focus more on cases with existing stakers, I only added 3 extra ones just for sanity and to ensure the math still checks out.

As for tests when there are existing stakers, this will come in a later PR to reduce PR size and complexity.

You will notice I have added additional helpers to ensure tests are not too verbose. Have also added a `setNextTimestamp` helper that did not exist, for reasons explained in a relevant testcase comment below.


apps/smart-contracts/external/yield/test/SingleTokenStakingRewards.js
```diff
@@ -514,31 +519,139 @@ contract('SingleTokenStakingRewards', (accounts) => {
...
+       * testcases simply used magic amounts that were high enough in
+       * precision so that it didn't matter and because they didn't need to verify
+       * amounts were calculated correctly (understandably because they didn't
+       * modify the logic like we are).
+       */
+      it('reverts if reward is > balance', async () => {
+        assert.bnEqual(await StakingRewardsDeployed.totalSupply(), toBN(0))
+        const stakingContractBalance = await rewardsToken.balanceOf(StakingRewardsDeployed.address)
+        const rewardsDuration = await StakingRewardsDeployed.rewardsDuration()
+        // Add rewardsDuration to account for integer division loss
```
XavierEkkel, line 564: `Add rewardsDuration to account for integer division loss (because it divides by rewardsDuration)`

apps/smart-contracts/external/yield/test/SingleTokenStakingRewards.js
```diff
@@ -514,31 +519,139 @@ contract('SingleTokenStakingRewards', (accounts) => {
...
+
+      it('reverts if reward after rollover > balance', async () => {
+        assert.bnEqual(await StakingRewardsDeployed.totalSupply(), toBN(0))
+        await StakingRewardsDeployed.notifyRewardAmount(rewardValue, {
           from: owner,
-        }),
-        'Provided reward too high'
-      )
+        })
+        await rewardsToken.transfer(StakingRewardsDeployed.address, rewardValue, { from: owner })
```
XavierEkkel, line 580: this is already done in beforeEach though?

apps/smart-contracts/external/yield/test/SingleTokenStakingRewards.js
```diff
@@ -514,31 +519,139 @@ contract('SingleTokenStakingRewards', (accounts) => {
...
+        // Subtract rewardsDuration here to account for integer division loss
+        const newRewardValue = stakingContractBalance.sub(rewardsDuration)
+
+        await StakingRewardsDeployed.notifyRewardAmount(newRewardValue, {
+          from: owner,
+        })
+
+        const expectedRewardForDuration = await getExpectedRewardForDuration(
+          newRewardValue,
+          await StakingRewardsDeployed.periodFinish(),
```
XavierEkkel, line 641: shouldnt this (`timeRewardsAdded`) be the current time?

apps/smart-contracts/external/yield/test/SingleTokenStakingRewards.js
```diff
@@ -514,31 +519,139 @@ contract('SingleTokenStakingRewards', (accounts) => {
...
+        assert.bnEqual(
+          await StakingRewardsDeployed.periodFinish(),
+          lastTimestamp.add(await StakingRewardsDeployed.rewardsDuration())
+        )
+      })
+
+      it('starts new reward if reward < balance', async () => {
+        const stakingContractBalance = await rewardsToken.balanceOf(StakingRewardsDeployed.address)
+        const rewardsDuration = await StakingRewardsDeployed.rewardsDuration()
+        // Subtract rewardsDuration here to account for integer division loss
```
XavierEkkel, line 632:
```suggestion
        // Subtract rewardsDuration here to account for integer division loss (because it divides by rewardsDuration)
```

apps/smart-contracts/external/yield/test/SingleTokenStakingRewards.js
ramenforbreakfast: I should add a comment here, but no, since there are previous rewards (no rollover as well), there is no need to manipulate a timestamp to control how much is leftover, `periodFinish` is passed here because the period has already passed for this testcase.

apps/smart-contracts/external/yield/test/SingleTokenStakingRewards.js
```diff
@@ -514,31 +519,139 @@ contract('SingleTokenStakingRewards', (accounts) => {
...
+       * precision so that it didn't matter and because they didn't need to verify
+       * amounts were calculated correctly (understandably because they didn't
+       * modify the logic like we are).
+       */
+      it('reverts if reward is > balance', async () => {
+        assert.bnEqual(await StakingRewardsDeployed.totalSupply(), toBN(0))
+        const stakingContractBalance = await rewardsToken.balanceOf(StakingRewardsDeployed.address)
+        const rewardsDuration = await StakingRewardsDeployed.rewardsDuration()
+        // Add rewardsDuration to account for integer division loss
+        const newRewardValue = stakingContractBalance.add(rewardsDuration)
```
XavierEkkel, line 565: add a test for when reward > balance but rounds down to balance
(should not revert)
`const newRewardValue = stakingContractBalance.add(rewardsDuration).sub(toBN(1))`

apps/smart-contracts/external/yield/test/SingleTokenStakingRewards.js
```diff
@@ -514,31 +519,139 @@ contract('SingleTokenStakingRewards', (accounts) => {
...
+       * same.
+       *
+       * The testcases have been heavily modified to use calculated balances
+       * that account for integer division loss explained above. Thales's
+       * testcases simply used magic amounts that were high enough in
+       * precision so that it didn't matter and because they didn't need to verify
+       * amounts were calculated correctly (understandably because they didn't
+       * modify the logic like we are).
+       */
+      it('reverts if reward is > balance', async () => {
```
XavierEkkel, line 560:
```suggestion
      it('reverts if reward > balance', async () => {
```

apps/smart-contracts/external/yield/test/SingleTokenStakingRewards.js
```diff
@@ -514,31 +519,139 @@ contract('SingleTokenStakingRewards', (accounts) => {
...
+
+        await assert.revert(
+          StakingRewardsDeployed.notifyRewardAmount(newRewardValue, {
+            from: owner,
+          }),
+          'Provided reward too high'
+        )
+      })
+
+      it('reverts if reward after rollover > balance', async () => {
```
XavierEkkel, line 575: add 2 additional tests:
when leftover rounds down to 0, when leftover = 0
(both of these should not revert)

apps/smart-contracts/external/yield/test/SingleTokenStakingRewards.js
ramenforbreakfast: Will add comment clarifying this is to provide tokens for the next period (since this is a rollover testcase)

apps/smart-contracts/external/yield/test/SingleTokenStakingRewards.js
ramenforbreakfast: So this wouldn't work, since even if you sub 1, it might still result in an amount that is a multiple of rewardsDuration and results in a higher rewardRate.

I think what I will do for these testcases is write a fixture to perform modulo division so that these amounts being added result in exact multiples of `rewardsDuration`.

ramenforbreakfast:
Ok, so I have significantly reworked the testcase based on last night's discussion.

One thing I should have clarified with you last night, is I still wasn't exactly sure about your request that we modify the `leftoverAmount` rather than the value being notified in `notifyRewardAmount`. 

But I believe your main concern was that one, `leftoverAmount` might be possibly zero if someone modified the test? And that adding `rewardsDuration` was crude and inaccurate, since the value would most likely end up being more than needed to register a higher reward rate.

The problem with altering `leftoverAmount` is that it is impossible to get the function to revert **just** by altering `leftoverAmount` without also intentionally not transferring reward tokens to contract or altering the reward amount.
We want to center tests around modifying the reward amount, rather than the `leftoverAmount`, because the `rewardAmount` is up to the user's discretion and is not bounded like `leftoverAmount`. I cannot force the `leftoverAmount` to revert unless either the `newReward` or token balance is deliberately also altered.

I think what is a easier to reason with approach, that still achieves the accuracy we need to prevent false positives, is determining the exact reward balance that would result in a higher reward rate than the maximum, based on the current balance.

Then we simply determined the `newReward` to add, by finding the difference between it and the contract balance(if there is no existing period), and the difference between it and the contract balance + leftover amount(if there is an existing period). We also assert that leftover amount is 0, so if someone were to alter the ~WEEK long setup, it would provide a sanity check there.

Since we are **now basing values on the exact amount** that would result in a higher reward rate, we can easily see that when it is exactly that amount, the function reverts, and then the following test should pass in that amount - 1, and see it succeed.

ramenforbreakfast:
I have also made some additional modifications to make it even easier to read, I think this version of the tests should be much easier to parse through and read.

ramenforbreakfast:
Modified name of `getNewRewardRate` since it needs to be different for `addReward`. The rest of the functions are still usable.

XavierEkkel (approved):
LGTM after nits and if move the `if period ongoing` describe block to a separate PR since it needs further discussion

apps/smart-contracts/external/yield/test/SingleTokenStakingRewards.js
```diff
@@ -514,31 +521,301 @@ contract('SingleTokenStakingRewards', (accounts) => {
...
+          const lastTimestamp = toBN(await currentTime())
+          assert.bnEqual(await StakingRewardsDeployed.lastUpdateTime(), lastTimestamp)
+          assert.bnEqual(
+            await StakingRewardsDeployed.periodFinish(),
+            lastTimestamp.add(rewardsDuration)
+          )
+        })
+      })
+
+      describe('if period ongoing', () => {
```
XavierEkkel, line 672: move to separate PR

apps/smart-contracts/external/yield/test/SingleTokenStakingRewards.js
```diff
@@ -514,31 +521,301 @@ contract('SingleTokenStakingRewards', (accounts) => {
...
+     * for rewards as a test input. We need to find the nearest reward amount
+     * that will result in a calculated reward rate that exceeds the max
+     * reward rate.
+     */
+    describe('if no stakers', () => {
+      /**
+       * These existing testcases from Thales have been modified and moved
+       * under `if no stakers`, since there were no test cases accounting
+       * for stakers. We obviously need to since these tokens will be the
+       * same.
```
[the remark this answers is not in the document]
ramenforbreakfast: No I meant as in, there needed to be a specific section `if no stakers` since there was no distinction between the two scenarios given that all the testcases had no stakers.

I will rewrite this to make this clearer.

PR #752, opened 2023-01-05
feat: add tests for when there are existing stakers

ramenforbreakfast:
`getBalanceForRewards` is a simple helper function to reduce the number of lines needed to calculate the difference between the contract balance and total supply from 3 lines to 1, since the tests are already quite verbose.


ramenforbreakfast:
Closing since it was requested that this PR and part of 751 be split into 3 more PRs.

PR #754, opened 2023-01-06
feat: modify addReward and add tests

ramenforbreakfast:
Adding helpers from 751 since it would be a hassle to decouple them now, but I wanted to be able to submit this separately from the train, they will be merged later.

The only helper that is added is `getNewRewardRateIfAddingReward`, since I decided not to verify `getRewardForDuration` for `addReward`. This is because a new period is not being set in `addReward`, and the reward rate calculated is based on the remaining time, so the `rewardForDuration` would be misleading anyway.


ramenforbreakfast:
Closing since we no longer need `addReward`

PR #755, opened 2023-01-06
feat: add addReward tests for existing stakers

ramenforbreakfast:
Essentially the same as PR #754, but the balance being used only includes that set aside for rewards, since there is an existing staker.


ramenforbreakfast:
closing since we are no longer using `addReward`

PR #756, opened 2023-01-07
feat: remove unused totalRewards variable

ramenforbreakfast:
While comparing our contract to both the original by Synthetix and Thales's, I realized there is a variable Thales added in, `totalRewards` that isn't used at all.


XavierEkkel (approved):
LGTM

PR #757, opened 2023-01-07
feat: add notifyRewardAmount tests (no stakers and period not finished)

ramenforbreakfast:


apps/smart-contracts/external/yield/test/SingleTokenStakingRewards.js
```diff
@@ -667,6 +667,154 @@ contract('SingleTokenStakingRewards', (accounts) => {
...
+            rewardsDuration
+          )
+          const leftoverRewards = await getLeftoverRewards(
+            timeToNotifyRewards,
+            StakingRewardsDeployed
+          )
+          assert.bnGt(leftoverRewards, toBN(0))
+          // Subtract 1 so that reward rate calculation rounds down
+          const newRewardValue = balanceForHigherRate.sub(leftoverRewards).sub(toBN(1))
+          const expectedRewardForDuration = await getExpectedRewardForDuration(
```
XavierEkkel, line 729: nit: all these can go in the assert section (right before it gets used)

XavierEkkel (approved):
LGTM after nits

apps/smart-contracts/external/yield/test/SingleTokenStakingRewards.js
ramenforbreakfast: `newRewardValue` is needed to actually pass into the `notifyRewardAmount` call, and `expectedRewardForDuration`, I have explained why this is not being done in 758.

PR #761, opened 2023-01-10
feat: remove addReward from SingleTokenStakingRewards

ramenforbreakfast:


apps/smart-contracts/external/yield/test/SingleTokenStakingRewards.js
```diff
@@ -166,7 +166,7 @@ contract('SingleTokenStakingRewards', (accounts) => {
         from: owner,
       })
 
-      await StakingRewardsDeployed.addReward(toUnit(5), { from: owner })
+      await StakingRewardsDeployed.notifyRewardAmount(toUnit(5), { from: owner })
```
XavierEkkel, line 169: I don't think simply replacing `addReward` with `notifyRewardAmount` will give the same behaviour for all these tests

ramenforbreakfast: So I had to do this to get the tests to compile, should have probably explained this in the description. Actually modifying the tests to check exact amounts is going to be a much bigger effort.

All the tests perform > and < than assertions, so they still pass because they do not verify exact amounts.

ramenforbreakfast: I have updated it according to our discussion, no tests were removed, but `notifyRewardAmount` calls were condensed. I did not see any test that actually demanded behavior that would have been different if `addReward` was called.

apps/smart-contracts/external/yield/test/SingleTokenStakingRewards.js
```diff
@@ -216,7 +216,7 @@ contract('SingleTokenStakingRewards', (accounts) => {
...
```
XavierEkkel, line 219: this needs to be condensed with above

apps/smart-contracts/external/yield/test/SingleTokenStakingRewards.js
XavierEkkel, line 169: this needs to be condensed with above

XavierEkkel (approved):
LGTM after comments

PR #762, opened 2023-01-10
feat: add emergencyWithdraw function and tests

ramenforbreakfast:
So I know in our very brief discussion, it was proposed to name this variable `claimingEnded`, but I think that did not accurately capture what this variable should be doing. I believe in this discussion we talked about only having `emergencyWithdraw` paused rewards claiming, but due to the permanent nature of this function, I think it is more apt to also halt other functions as well, so that they are not accidentally called and for consistency.

Therefore, I have named this instead `stakingEnded` and the modifier `whileStakingNotEnded`, and have applied it to the `stake` and `notifyRewardAmount` functions as well, since there is no point in allowing further staking or starting new reward periods.

I have also added this modifier to `emergencyWithdraw` to further emphasize that this is a permanent operation and the contract should not be used again.


apps/smart-contracts/external/yield/contracts/EscrowAndStaking/SingleTokenStakingRewards.sol
```diff
@@ -207,6 +217,19 @@ contract SingleTokenStakingRewards is
     emit RewardsDurationUpdated(rewardsDuration);
   }
 
+  /**
+   * @dev This function is meant to be called prior to migration to a new
+   * staking contract and is a permanent operation.
+   *
+   * Afterwards, stakers can still withdraw their tokens, but the contract
+   * can no longer be used to issue rewards to stakers.
```
XavierEkkel, line 225:
```suggestion
   * can no longer be used to issue new rewards to stakers, and any existing unclaimed rewards will be forfeited.
```

apps/smart-contracts/external/yield/test/SingleTokenStakingRewards.js
```diff
@@ -1087,4 +1078,155 @@ contract('SingleTokenStakingRewards', (accounts) => {
...
+        }),
+        'Staking has ended'
+      )
+    })
+
+    it('allows users to withdraw', async () => {
+      const userBalanceBefore = await rewardsToken.balanceOf(mockRewardsDistributionAddress)
+      const userStakedBalanceBefore = await StakingRewardsDeployed.balanceOf(
+        mockRewardsDistributionAddress
+      )
```
XavierEkkel, line 1213: add asserts that these are both above 0

apps/smart-contracts/external/yield/test/SingleTokenStakingRewards.js
```diff
@@ -1087,4 +1078,155 @@ contract('SingleTokenStakingRewards', (accounts) => {
...
+      const ownerBalanceBefore = await rewardsToken.balanceOf(owner)
+
+      await StakingRewardsDeployed.emergencyWithdraw({
+        from: owner,
+      })
+
+      assert.bnEqual(await rewardsToken.balanceOf(owner), ownerBalanceBefore.add(balanceForRewards))
+    })
+
+    it('stops users from claiming', async () => {
```
XavierEkkel, line 1143: also add revert tests for exit():
- if unclaimed > 0
- if unclaimed = 0

apps/smart-contracts/external/yield/test/SingleTokenStakingRewards.js
```diff
@@ -1087,4 +1078,155 @@ contract('SingleTokenStakingRewards', (accounts) => {
...
+      assert.bnGt(contractBalance, balanceForRewards)
+      // Verify that the contract has rewards to withdraw
+      assert.bnGt(balanceForRewards, toBN(0))
+      const ownerBalanceBefore = await rewardsToken.balanceOf(owner)
+
+      await StakingRewardsDeployed.emergencyWithdraw({
+        from: owner,
+      })
+
+      assert.bnEqual(await rewardsToken.balanceOf(owner), ownerBalanceBefore.add(balanceForRewards))
```
XavierEkkel, line 1140: add asserts:
- contract balance after should = contractBalanceBefore - balanceForRewards
- contract balance after should = totalSupply

apps/smart-contracts/external/yield/test/SingleTokenStakingRewards.js
ramenforbreakfast: Will add an assertion that the staked is above 0, wont for the unstaked because that would mean transferring more rewards again for no real reason. We are verifying that the balance is the balance before + withdrawn amount, so no point in verifying it is above 0.

PR #763, opened 2023-01-10
feat: add additional tests for compound

ramenforbreakfast:
I have deleted the old makeshift test I made initially and have replaced it with a more robust test covering the four cases outlined in our checklist.

Additional helpers are needed for accurate expected rewards calculation. Both tests and helpers should be sufficiently commented for clarification.


apps/smart-contracts/external/yield/test/utils/helpers.js
```diff
@@ -356,4 +356,46 @@ module.exports = {
...
+     * to keep everything in BN since all on-chain values are returned as BN.
+     *
+     * Unlike ethers BigNumber, web3 BN's arithmetic operations do not
+     * automatically convert to BN, so need to explicitly convert to BN.
+     */
+    const timeRewardsClaimedBN = web3.utils.toBN(timeRewardsClaimed)
+    const periodFinish = await stakingContract.periodFinish()
+    const lastTimeRewardApplicable = timeRewardsClaimedBN.gt(periodFinish)
+      ? periodFinish
+      : timeRewardsClaimedBN
```
XavierEkkel, line 383:
```suggestion
    const lastTimeRewardApplicable = timeRewardsClaimedBN.lt(periodFinish)
      ? timeRewardsClaimedBN
      : periodFinish
```
keep consistent with the contract's logic (and should've been >= when it was flipped around anyway)

apps/smart-contracts/external/yield/test/utils/helpers.js
```diff
@@ -356,4 +356,46 @@ module.exports = {
...
+  },
+
+  async getRewardsEarned(stakingContract, account, timeRewardsClaimed) {
+    const rewardPerToken = await module.exports.getRewardPerToken(
+      stakingContract,
+      timeRewardsClaimed
+    )
+    const balance = await stakingContract.balanceOf(account)
+    const userRewardPerTokenPaid = await stakingContract.userRewardPerTokenPaid(account)
+    const rewards = balance.mul(rewardPerToken.sub(userRewardPerTokenPaid)).div(toUnit(1))
```
XavierEkkel, line 398: is this supposed to be the `earned()` function?
if so, it's missing `.add(rewards[account])`

if not, can rename the function `getNewRewardsEarned`

apps/smart-contracts/external/yield/test/SingleTokenStakingRewards.js
```diff
@@ -289,42 +290,151 @@ contract('SingleTokenStakingRewards', (accounts) => {
...
-        from: owner,
+      assert.bnEqual(await rewardsToken.balanceOf(mockRewardsDistributionAddress), toBN(0))
+      // estimate reward amount and ensure it is > 0
+      timeToClaimRewards = (await currentTime()) + 10
+      const userEarnedRewards = await getRewardsEarned(
+        StakingRewardsDeployed,
+        mockRewardsDistributionAddress,
+        timeToClaimRewards
+      )
+      assert.bnGt(userEarnedRewards, toBN(0))
```
XavierEkkel, line 380: instead of this, just check that `earned()` has the desired value at the end of the arrange section, since that's ultimately what we care about
this would probably make the `getRewardsEarned` helper function redundant

XavierEkkel (commented):
almost LGTM, just some comments

apps/smart-contracts/external/yield/test/SingleTokenStakingRewards.js
ramenforbreakfast: Wait but this is `compound`? `compound` will grab the rewards they've earned and stake them, meaning `earned` will be effectively 0 after `compound` is called.

We are doing this at the beginning to ensure that the rewards being claimed via `compound` are non-zero.

apps/smart-contracts/external/yield/test/SingleTokenStakingRewards.js
```diff
@@ -289,42 +290,151 @@ contract('SingleTokenStakingRewards', (accounts) => {
...
-      await stakingToken.approve(StakingRewardsDeployed.address, toUnit(100), {
+    it('reverts if balance = 0 and unclaimed = 0', async () => {
+      /**
+       * Since we need the user's unstaked balance to be zero, and thales test suite
+       * sends the user tokens before all tests, its easiest to just fast forward
+       * a week and then stake the initial amount. This will make the unstaked balance
+       * and unclaimed balance = 0 while not disrupting the setup for the other 3
+       * cases.
+       */
+      await fastForward(WEEK)
```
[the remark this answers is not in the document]
ramenforbreakfast: This is done because there is already an unstaked balance in the user's wallet, given by the thales test setup. So we can either burn the tokens, or let the period pass + stake. I chose the latter since it is better to test having a user with a staked balance after the active period has passed.

apps/smart-contracts/external/yield/test/SingleTokenStakingRewards.js
XavierEkkel: this is the arrange section, so earned() be zero (we expect it to be non-zero before calling compound, and zero after)

apps/smart-contracts/external/yield/test/SingleTokenStakingRewards.js
```diff
@@ -289,42 +290,151 @@ contract('SingleTokenStakingRewards', (accounts) => {
...
+    })
 
-      await stakingToken.approve(StakingRewardsDeployed.address, toUnit(100), {
+    it('reverts if balance = 0 and unclaimed = 0', async () => {
+      /**
+       * Since we need the user's unstaked balance to be zero, and thales test suite
+       * sends the user tokens before all tests, its easiest to just fast forward
+       * a week and then stake the initial amount. This will make the unstaked balance
+       * and unclaimed balance = 0 while not disrupting the setup for the other 3
+       * cases.
```
XavierEkkel, line 314: clarify that you're staking after period has finished and therefore there will be no accumulated rewards

PR #764, opened 2023-01-10
feat: add notifyRewardAmount tests for when amount=0

ramenforbreakfast:


apps/smart-contracts/external/yield/test/SingleTokenStakingRewards.js
```diff
@@ -832,6 +849,37 @@ contract('SingleTokenStakingRewards', (accounts) => {
...
+
+        it('resets period with leftover rewards if reward = 0', async () => {
+          assert.bnEqual(await StakingRewardsDeployed.totalSupply(), toBN(0))
+          const leftoverRewards = await getLeftoverRewards(
+            timeToNotifyRewards,
+            StakingRewardsDeployed
+          )
+          assert.bnGt(leftoverRewards, toBN(0))
+          const newRewardValue = toBN(0)
+          const expectedRewardForDuration = await getExpectedRewardForDuration(
```
XavierEkkel, line 861: add assertion that this is also >0
same for the duplicate below

apps/smart-contracts/external/yield/test/SingleTokenStakingRewards.js
```diff
@@ -950,6 +998,23 @@ contract('SingleTokenStakingRewards', (accounts) => {
             lastTimestamp.add(rewardsDuration)
           )
         })
+
+        it('starts new reward of 0 if reward = 0', async () => {
+          assert.bnGt(await StakingRewardsDeployed.totalSupply(), toBN(0))
```
XavierEkkel, line 1003: this is in the 'if existing stakers' describe block - shouldn't total supply be > 0?
same for the one below

ramenforbreakfast: But it is? `bnGt`?

apps/smart-contracts/external/yield/test/SingleTokenStakingRewards.js
ramenforbreakfast: Will add this.

apps/smart-contracts/external/yield/test/SingleTokenStakingRewards.js
XavierEkkel: oops I read it as `eq`, ignore this!

XavierEkkel (approved):
LGTM

PR #773, opened 2023-01-17
feat: add helper for initialization of market pools

ramenforbreakfast:
Added helpers to setup pools for a market. Made modifications to test setup to move away from interacting with pools directly and needing to store pools, since swap operations will be done through a router and they can be searched for inside a helper if data needs to be polled from them.


apps/smart-contracts/core/test/ArbitrageTrading.test.ts
```diff
@@ -80,21 +71,15 @@ describe('=> Functional Testing with Arbitrage Trades', () => {
...
-      )
-    )
     positionManager = await attachNonfungiblePositionManager(
       getPrePOAddressForNetwork('UNIV3_POSITION_MANAGER', 'arbitrumOne')
     )
+    await core.deployPoolsForMarket(
+      TEST_NAME_SUFFIX,
+      univ3Factory,
+      getNearestSqrtX96FromWei(parseEther('0.5')),
+      getNearestSqrtX96FromWei(parseEther('0.5'))
```
XavierEkkel, line 81: lets also move the `getNearestSqrtX96FromWei()` call into the `deployPoolsForMarket` function itself for simplicity + to reduce chance of mistakes?

apps/smart-contracts/core/harnesses/base.ts
```diff
@@ -84,4 +87,36 @@ export abstract class Base {
     )
     return [longTokenSalt, shortTokenSalt]
   }
+
+  public async deployPoolsForMarket(
+    tokenNameSuffix: string,
+    univ3Factory: UniswapV3Factory,
+    longPoolPrice: BigNumber,
+    shortPoolPrice: BigNumber
```
XavierEkkel, line 95:
```suggestion
    approxLongPoolWeiPrice: BigNumber,
    approxShortPoolWeiPrice: BigNumber
```

XavierEkkel (approved):
LGTM after comments

apps/smart-contracts/core/test/ArbitrageTrading.test.ts
ramenforbreakfast: I actually got stuck while debugging due to this very problem, but hesitated retroactively submitting a change to that function.

PR #774, opened 2023-01-17
feat: add helper for minting liquidity positions

ramenforbreakfast:
Helper function for supplying liquidity in test scenarios, takes in Collateral wei values for price ranges and then converts those to the nearest tick (makes it easier for a reviewer to understand the tests looking at a ethers value, rather than tick or SqrtPriceX96).


apps/smart-contracts/core/helpers/uniswap.ts
```diff
@@ -77,3 +90,41 @@ export function getNearestSqrtX96FromWei(weiPrice: BigNumber): BigNumber {
...
+  const params = <INonfungiblePositionManager.MintParamsStruct>{
+    token0: longShortToken.address,
+    token1: collateral.address,
+    fee: POOL_FEE_TIER,
+    tickLower: lowerTick,
+    tickUpper: upperTick,
+    amount0Desired: longShortAmount,
+    amount1Desired: collateralAmount,
+    amount0Min: 0,
+    amount1Min: 0,
```
XavierEkkel, line 123: what is the effect of not setting min amounts? any risks to this? could you leave a comment here?

apps/smart-contracts/core/helpers/uniswap.ts
```diff
@@ -77,3 +90,41 @@ export function getNearestSqrtX96FromWei(weiPrice: BigNumber): BigNumber {
...
+  const lowerTick = getNearestTickFromWei(approxLowerWeiBound)
+  const upperTick = getNearestTickFromWei(approxUpperWeiBound)
+  const params = <INonfungiblePositionManager.MintParamsStruct>{
+    token0: longShortToken.address,
+    token1: collateral.address,
+    fee: POOL_FEE_TIER,
+    tickLower: lowerTick,
+    tickUpper: upperTick,
+    amount0Desired: longShortAmount,
+    amount1Desired: collateralAmount,
```
XavierEkkel, line 121: there may be leftover amounts of either token if the market price has moved since calculating these values off-chain, correct? 
please add a commen there

XavierEkkel (approved):
LGTM after comments

apps/smart-contracts/core/helpers/uniswap.ts
ramenforbreakfast: Since we plan to use this same function for our actual helper, I will make these parameters with default values set to 0 and add comments, since in prod we'd want to set protections.

PR #775, opened 2023-01-18
feat: simplified trade button

chrisling-dev:
https://www.notion.so/prepo/FE-Project-Management-d3878c29e9534234a50f4f8838525bee?p=61ed1d66f0bb4b9b9074f3b531299130&pm=s

https://user-images.githubusercontent.com/81092286/214215767-75e170a0-a993-4c7b-b133-cd43e540530e.mov


apps/frontend/core-dapp/src/features/connect/ConnectButton.tsx
```diff
@@ -1,36 +1,38 @@
...
-  const { web3Store } = useRootStore()
-  const { signerState } = web3Store
-  const account = signerState.address
-
-  const onClickLogin = (): void => {
+  const handleSwitchNetwork = (): void => {
+    if (!isNetworkSupported) {
+      web3Store.setNetwork(defaultNetwork)
+      return
+    }
```
liqwiz, line 18: is this supposed to always set the default? I'd have assumed it to handle multiple

apps/frontend/core-dapp/src/stores/UiStore.ts
```diff
@@ -49,6 +49,10 @@ export class UiStore {
   }
 
   errorToast(title: string, error: unknown): void {
+    if (typeof error === 'string') {
+      this.message.error(error)
+      return
+    }
```
liqwiz, line 55: I'd suggest only every throwing Errors rather than adding handling for other types if you can 

e.g. replace `throw "error-123"` with `throw Error("error-123")`

packages/prepo-constants/src/networks.ts
```diff
@@ -149,7 +149,7 @@ export const NETWORKS: Networks = {
     blockExplorer: 'https://goerli.etherscan.io',
     iconName: 'weth',
     rpcUrls: [`https://goerli.infura.io/v3/${METAMASK_INFURA_ID}`],
-    gasPrice: utils.parseUnits('4', 'gwei'),
+    gasPrice: utils.parseUnits('30', 'gwei'),
```
liqwiz, line 152: curious behind why you're upping this? goerli seems to be consistently mining txns with very low gas these days

apps/frontend/core-dapp/src/features/connect/ConnectButton.tsx
chrisling-dev: it'll set to default when user is on wrong network, hence it only calls `setNetwork(defaultNetwork)` if `!isNetworkSupported`

Otherwise it'll just connect to whatever network user is on

So if we support both goerli and arbitrum, and you're on either of them, you'll stay on the same chain, but if you're on mainnet, then it'll show "Switch to {default network}" and calls setNetwork(defaultNetwork) on click

packages/prepo-constants/src/networks.ts
chrisling-dev: All tx wasnt going through at until I changed to 30, I tried increasing it incrementally tho at that time when building this, from 4 - 10 - 20 - 30, 30 was the amount that got tx go through quick enough. So I thought since eth on goerli are "free" anyway, we might as well improve the UX by increasing the gas price

apps/frontend/core-dapp/src/stores/UiStore.ts
chrisling-dev: Yeap agree, I found this errorToast in UiStore is deprecated after making this change and changed to using the toastStore, will undo this change and make another PR to remove these toasting from UiStore

chrisling-dev (commented):
self reviewed

apps/frontend/core-dapp/src/features/trade/open-trade/OpenTradeButton.tsx
```diff
@@ -0,0 +1,65 @@
...
+      tradeStore.approve()
+    } else {
+      tradeStore.openTrade()
+    }
+  }
+
+  return (
+    <Button
+      block
+      disabled={tradeDisabled || !selectedMarket || openTradeAmountBN?.eq(0) || !withinBounds}
```
chrisling-dev, line 56:
```suggestion
      disabled={!selectedMarket || openTradeAmountBN?.eq(0) || !withinBounds}
```

remove tradeDisabled

liqwiz (approved):

XavierEkkel (approved):
LGTM after comments

apps/frontend/core-dapp/src/features/trade/TradeStore.ts
```diff
@@ -57,20 +59,48 @@ export class TradeStore {
         this.openTradeAmountOutBN = undefined // clean up while new amountOut gets loaded
         const openTradeAmountOutBN = await this.quoteExactInput(selectedMarket)
         runInAction(() => {
-          this.openTradeAmountOutBN = openTradeAmountOutBN
+          if (openTradeAmountOutBN !== undefined) this.openTradeAmountOutBN = openTradeAmountOutBN
```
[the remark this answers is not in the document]
chrisling-dev: nope it still matters, to prevent the previous call from overriding the latest one. Here's an example:

user types:
1. 10 -> fetch amountOut for 10
2. 100 -> fetch amountOut for 100
3. Some how amountOut for 10 isn't resolved yet, and amountOut for 100 is resolved -> update openTradeAmountOutBN to result of 100
4. then amountOut for 10 resolves to undefined because input is now 100 -> changes amountOutBN to undefined
5. now UI shows loading when it should be showing result of 100

PR #776, opened 2023-01-19
core:fix: add L/S approvals to market for redemption

ramenforbreakfast:
This was identified while debugging functional testing for the `ArbitrageBroker`. It needs this approvals so it can call `redeem` on the L/S tokens. I think it was assumed that LongShort tokens could be burnt without approval by the market but this is not the case.


XavierEkkel (approved):
LGTM

PR #777, opened 2023-01-19
feat: add helpers for estimating swap amounts for testing

ramenforbreakfast:
These are helper functions for statically estimating a swap for testing purposes. This is so we can verify that the amount of loss or profit reported by the broker matches what would happen if the swaps were executed independently.
Notice that since these are meant to be run without slippage constraints, `amountInMaximum` is the maximum uint256, indicating a lack of slippage bounds, whereas `amountOutMinimum` is 0.

Also adds an entry for the UniswapV3 deployed SwapRouter and attach fixture so it can be accessed to run these simulations.


apps/smart-contracts/core/helpers/uniswap.ts
```diff
@@ -91,6 +94,58 @@ export function getNearestSqrtX96FromWei(weiPrice: BigNumber): BigNumber {
...
+  swapRouter: SwapRouter
+): Promise<BigNumber> {
+  const simulationParams = <ISwapRouter.ExactOutputSingleParamsStruct>{
+    tokenIn: tokenIn.address,
+    tokenOut,
+    fee: POOL_FEE_TIER,
+    recipient,
+    deadline: nowPlusMonths(2),
+    amountOut,
+    amountInMaximum: ethers.constants.MaxUint256,
```
XavierEkkel, line 142: add comment about why we use this value (from PR description)

apps/smart-contracts/core/helpers/uniswap.ts
```diff
@@ -91,6 +94,58 @@ export function getNearestSqrtX96FromWei(weiPrice: BigNumber): BigNumber {
...
+  swapRouter: SwapRouter
+): Promise<BigNumber> {
+  const simulationParams = <ISwapRouter.ExactInputSingleParamsStruct>{
+    tokenIn: tokenIn.address,
+    tokenOut,
+    fee: POOL_FEE_TIER,
+    recipient,
+    deadline: nowPlusMonths(2),
+    amountIn,
+    amountOutMinimum: 0,
```
XavierEkkel, line 116: add comment about why we use this value (from PR description)

apps/smart-contracts/core/helpers/uniswap.ts
```diff
@@ -91,6 +94,58 @@ export function getNearestSqrtX96FromWei(weiPrice: BigNumber): BigNumber {
...
+    sqrtPriceLimitX96: 0,
```
XavierEkkel, line 117: add comment about why we use 0 for this
same for the other function

apps/smart-contracts/core/helpers/uniswap.ts
```diff
@@ -91,6 +94,58 @@ export function getNearestSqrtX96FromWei(weiPrice: BigNumber): BigNumber {
...
+  funder: SignerWithAddress,
+  recipient: string,
+  swapRouter: SwapRouter
+): Promise<BigNumber> {
+  const simulationParams = <ISwapRouter.ExactInputSingleParamsStruct>{
+    tokenIn: tokenIn.address,
+    tokenOut,
+    fee: POOL_FEE_TIER,
+    recipient,
+    deadline: nowPlusMonths(2),
```
XavierEkkel, line 114: let's extract to a variable to use in both functions + add a comment for why we use this particular value

apps/smart-contracts/core/helpers/uniswap.ts
```diff
@@ -91,6 +94,58 @@ export function getNearestSqrtX96FromWei(weiPrice: BigNumber): BigNumber {
...
+    amountIn,
+    amountOutMinimum: 0,
+    sqrtPriceLimitX96: 0,
+  }
+  await tokenIn.connect(funder).approve(swapRouter.address, ethers.constants.MaxUint256)
```
XavierEkkel, line 119: the approval is an actual non-simulated tx, correct?
if so and if this is going to be used by the actual arbitrage script, this should check whether approval already exists first so this doesn't get called again unnecessarily

also add comment above approve+exactInputSingle indicating which ones are actual vs. simulated

same for the other function

XavierEkkel (approved):
LGTM after comments

apps/smart-contracts/core/helpers/uniswap.ts
ramenforbreakfast: Modified this to be a parameter with a default value so it is immediately clear rather than requiring additional documentation.

PR #779, opened 2023-01-19
feat: add major modification to test setup identified while tests were written

ramenforbreakfast:
The original test `beforeEach` block was essentially written as scaffolding to provide an environment to test all the bits and pieces that were necessary to get to this point. Now that the actual functional tests have been written, I have separated out all the changes to the test setup in this PR (**to reduce PR size**) prior to the tests for `buyAndRedeem` and `mintAndSell`.

I have renamed the test suite to reflect that I plan to test the script functions as well within this test suite, since we won't need to write a new test setup when we need to test functions that involve `ArbitrageBroker` static calls.

I have also added documentation to the steps since it is quite a lengthy setup, even with all the helpers trying to make things as concise as possible.

Notice that we are no longer deploying pools for markets at the very beginning, because it was realized that the pools needed to be initialized with different prices depending on the test case to make the liquidity setup simpler.


apps/smart-contracts/core/test/ArbitrageTrading.test.ts
```diff
@@ -61,25 +86,47 @@ describe('=> Functional Testing with Arbitrage Trades', () => {
...
-      parseEther('0.5')
+    swapRouter = await attachSwapRouter(
+      getPrePOAddressForNetwork('UNIV3_SWAP_ROUTER', 'arbitrumOne')
     )
+    // Only need to assign deposit fee since test is only using deposit
+    await assignCollateralRoles(deployer, governance, core.collateral)
+    await core.collateral.connect(governance).setDepositFee(10000)
+    // Supply governance with Collateral and LongShort tokens
+    await core.mintLSFromBaseToken(governance, governance, parseEther('10000'), TEST_NAME_SUFFIX)
+    await core.mintCollateralFromBaseToken(governance, governance.address, parseEther('10000'))
```
XavierEkkel, line 118: extract the `parseEther('10000')` to a const variable

apps/smart-contracts/core/test/ArbitrageTrading.test.ts
```diff
@@ -28,16 +40,29 @@ describe('=> Functional Testing with Arbitrage Trades', () => {
...
   const TEST_FLOOR_VAL = BigNumber.from(2000)
   const TEST_CEILING_VAL = BigNumber.from(10000)
   const TEST_EXPIRY = nowPlusMonths(2)
+  const TEST_POSITION_SIZE = parseEther('1')
+  /**
+   * Since L/S tokens will be priced at < 1 Collateral, so supplying
+   * twice the Long and Short amount we plan to buy in Collateral is
+   * more than sufficient, and is not a magic number.
+   */
+  const TEST_BROKER_CAPITAL = TEST_POSITION_SIZE.mul(2)
```
XavierEkkel, line 58: given above, should only need:
```suggestion
  const TEST_BROKER_CAPITAL = TEST_POSITION_SIZE
```
?

XavierEkkel (approved):
LGTM after comments

apps/smart-contracts/core/test/ArbitrageTrading.test.ts
```diff
@@ -28,16 +40,29 @@ describe('=> Functional Testing with Arbitrage Trades', () => {
...
 
   snapshotter.setupSnapshotContext('ArbitrageTrading')
   before(async () => {
+    /**
+     * Connect to Alchemy provider since forking off a specific block
```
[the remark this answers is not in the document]
ramenforbreakfast: I am going to do this in a following PR since you cannot specifically exclude a file, but you can specify a specific file or folder to run. I think the best way to do this would be to have a `unit` and `functional` folder in `test` and then change the `t` command to only run tests under `unit`.

apps/smart-contracts/core/test/ArbitrageTrading.test.ts
ramenforbreakfast: This would be true if L/S tokens were _exactly_ 0.5 `Collateral`, but these tests will price L/S tokens at slightly above or below 0.5. TEST_POSITION_SIZE wouldn't be enough, because it could potentially cost more than TEST_POSITION_SIZE `Collateral` to buy both Long and Short token.
e.g. in the case of a non-profitable `buyAndRedeem` situation, it would cost more than (`TEST_POSITION_SIZE` / 2) `Collateral` to buy Long since Long will be priced at slightly above 0.5 `Collateral`, same with buying Short which will add up to > `TEST_POSITION_SIZE` in `Collateral` being spent.

PR #781, opened 2023-01-19
feat: add mintAndSell functional tests

ramenforbreakfast:


apps/smart-contracts/core/test/ArbitrageTrading.test.ts
```diff
@@ -324,5 +325,159 @@ describe('=> Arbitrage Trading', () => {
...
+          parseEther('0.5')
+        )
+        await mintLiquidityForLongShort(
+          core.markets[TEST_NAME_SUFFIX].longToken,
+          core.collateral,
+          positionManager,
+          parseEther('0.49'),
+          parseEther('0.5'),
+          BigNumber.from(0),
+          TEST_POSITION_SIZE.mul(2),
```
XavierEkkel, line 350:
```suggestion
          TEST_POSITION_SIZE.add(1)
```
same for others

XavierEkkel (approved):
LGTM after comment

apps/smart-contracts/core/test/ArbitrageTrading.test.ts
ramenforbreakfast: So I actually am going to have to backtrack on supplying just + 1, and have changed all LP positions back to `TEST_POSITION_SIZE.mul(2)`.

If we only provide + 1, it will cause the last testcase to revert. I eventually figured out the reason why is that if you just provide the amount or near the amount we are purchasing, the pricing will cause your average purchase/sell price to be much closer to the upper/lower end of the range.

In the last testcase, we are supplying from 0.51 => 0.5, but because we only supply the position size + 1, the average sale price for `mintAndSell` hugs the lower end (0.5), and since 0.5 is not actually 0.5, but 0.496, the sale still results in a loss, even though linearly, if you take the average of the two range ends (0.506 and 0.496), the average sale price should result in > 0.5.

So since the price curve is not linear, I have just switched them all back to `.mul(2)` rather than changing the ticks which I think would make the tests more confusing. I have also added comments explaining this.

PR #783, opened 2023-01-20
feat: track purchase event with public sale subgraph

chrisling-dev:
Track Purchase event with subgraph


XavierEkkel (approved):
LGTM after nit

PR #789, opened 2023-01-23
fix: remove marketName prop from EstimatedValuation

chrisling-dev:
Remove marketName prop usage from all EstimatedValuation tooltip to fix typescript complaint


XavierEkkel (approved):
LGTM

PR #791, opened 2023-01-24
core:chore: separate unit tests from integration ones

ramenforbreakfast:
If desired, we can later also additional commands for `unit` and `integration` along with GitHub Actions mappings. Right now actions will just run unit tests by default if `core:` is specified.


XavierEkkel (approved):
LGTM

PR #792, opened 2023-01-24
feat: add initial Arbitrage script with token getter

ramenforbreakfast:
So with this script, I am going to begin with the smaller functions, and incrementally build out the full script.

This is just a simple token getter so the script can grab Long/Short tokens from the markets it is iterating through.

For context, there are some things I am rewriting and not simply re-using because we have to write the functions with the knowledge that we will have a limited set of available dependencies to us. Things like typechain generated types, hardhat functionality, all will not be available to us inside OZ Autotasks unless we explicitly add those generated files (and some just not possible/practical due to limited file size).

Therefore, I am writing this only using what is available in the default `ethers` library, and using contract artifacts that we will either later have to bundle together with the script (OZ has a tutorial with how to do this, but I am putting it off for testing purposes), or hardcoding it into the file itself (which is what the example Autotasks do).

I am also writing it in a way where we can still just directly test them in our existing test suite, so we do not have to spend additional time writing a new test suite that has a full stack for testing (this function is simple enough where its not actually necessary but others will need it).

After we are confident in testing, we can copy this content over to a isolated directory when we actually test with real relayers against a deployed testnet.


XavierEkkel (approved):
LGTM

PR #793, opened 2023-01-24
feat: add pool getter for Arbitrage script

ramenforbreakfast:
A simple getter that retrieves pools given a set of tokens. It takes in the token addresses directly rather than the market since this is intended to be called after `getTokensFromMarket` is called.

Note that I am also not using our existing packages, like I could have here with the `UNIV3_FACTORY` definition, since there is no reason we need to add additional bloat by importing all of our constants. I will continue writing this script with defining as much as possible within it unless there is a good reason to do so or it is impractical, like with contract artifacts.


apps/smart-contracts/core/scripts/ArbitragePools.ts
```diff
@@ -1,12 +1,22 @@
 import { Contract, ethers } from 'ethers'
 import IERC20Artifact from './exportedArtifacts/IERC20.json'
 import IPrePOMarketArtifact from './exportedArtifacts/IPrePOMarket.json'
+import IUniswapV3FactoryArtifact from './exportedArtifacts/IUniswapV3Factory.json'
+import IUniswapV3PoolArtifact from './exportedArtifacts/IUniswapV3Pool.json'
+
+const UNIV3_FACTORY = '0x1F98431c8aD98523631AE4a59f267346ea31F984'
+const POOL_FEE_TIER = 10000
```
XavierEkkel, line 8: lets reuse existing constants if possible

apps/smart-contracts/core/test/integration/ArbitrageTrading.test.ts
```diff
@@ -504,5 +504,40 @@ describe('=> Arbitrage Trading', () => {
...
+      beforeEach(async () => {
+        await core.deployPoolsForMarket(
+          TEST_NAME_SUFFIX,
+          univ3Factory,
+          parseEther('0.5'),
+          parseEther('0.5')
+        )
+      })
+
+      it('returns pools deployed for market', async () => {
```
XavierEkkel, line 518: there isn't a need for this test, since it just reuses the implementation

some useful tests might be for behaviour with invalid inputs though

XavierEkkel (approved):
LGTM after comments

apps/smart-contracts/core/scripts/ArbitragePools.ts
ramenforbreakfast: I'm going to do this at the very end, when we actually try to bundle everything together in a Autotask package. Right now I am more concerned with guaranteeing compatibility and ensuring the logic works. We can clean it up later by incrementally bundling in our packages.

apps/smart-contracts/core/test/integration/ArbitrageTrading.test.ts
ramenforbreakfast: When we test with actual RPC requests, we will wrap all requests in a block to ensure failed requests (which has a high likelihood of happening due to it being a remote request). So I will remove this test for now.

PR #794, opened 2023-01-24
feat: add wei price getter for initial comparison

ramenforbreakfast:
This is a getter that retrieves the price in wei from a set of market pools. This is used for the initial price comparison to decide whether we should search for an arbitrage opportunity.

Additional test helpers were added so we can easily verify that the script function does indeed grab the prices correctly.

I initially began writing that comparison function with the retrieval of prices inside, but realized that for unit-testing purposes, we should segment it out so we just need to verify that it retrieves prices correctly, and can easily test a separate function that just does the comparison. If they had been combined, it would have been impossible to test the comparison logic to the full extent since getting the pools to trigger certain spread conditions would be impossible (i.e. L + S exactly = 1 +/- spread)


apps/smart-contracts/core/scripts/ArbitragePools.ts
```diff
@@ -7,6 +7,10 @@ import IUniswapV3PoolArtifact from './exportedArtifacts/IUniswapV3Pool.json'
 const UNIV3_FACTORY = '0x1F98431c8aD98523631AE4a59f267346ea31F984'
 const POOL_FEE_TIER = 10000
 
+// Taken from helpers/uniswap.ts
```
XavierEkkel, line 10: why not just import the one from helpers/uniswap.ts?

apps/smart-contracts/core/helpers/uniswap.ts
```diff
@@ -15,8 +15,10 @@ import {
   ISwapRouter,
   NonfungiblePositionManager,
   SwapRouter,
+  UniswapV3Factory,
 } from '../types/generated'
 import { IncreaseLiquidityEvent } from '../types/generated/externalArtifacts/INonfungiblePositionManager'
+import { attachUniV3Pool } from '../test/fixtures/UniswapFixtures'
```
XavierEkkel, line 21: i think `attachUniV3Pool` makes more sense in this file rather than in a test fixtures file

XavierEkkel (approved):
LGTM after comments

apps/smart-contracts/core/test/integration/ArbitrageTrading.test.ts
```diff
@@ -539,5 +544,40 @@ describe('=> Arbitrage Trading', () => {
...
+      beforeEach(async () => {
+        await core.deployPoolsForMarket(
+          TEST_NAME_SUFFIX,
+          univ3Factory,
+          parseEther('0.49'),
+          parseEther('0.5')
+        )
+      })
+
+      it('returns wei prices for pools', async () => {
```
XavierEkkel, line 558: add some tests for behaviour when inputs are invalid

apps/smart-contracts/core/helpers/uniswap.ts
ramenforbreakfast: I went ahead and moved all of the attach functions over to the `uniswap.ts` file.

apps/smart-contracts/core/test/integration/ArbitrageTrading.test.ts
ramenforbreakfast: I will address this later. Mostly because right now I just want to get the logic finished, and focus on error handling when we actually test with live Relayers. We will need to write code that differentiates between calls that fail due to an unresponsive infra or if the call failed because the contract passed in was not valid.

I will add a comment at the top of the file detailing this TODO.

PR #795, opened 2023-01-24
feat: add logic to decide whether to size an opportunity

ramenforbreakfast:
I decided an enum was the simplest way to write the return value since the function also needs to inform the parent loop whether to skip sizing an opportunity and move to the next market.


apps/smart-contracts/core/scripts/ArbitragePools.ts
```diff
@@ -32,6 +32,12 @@ export type PoolPrices = {
   shortPoolPrice: BigNumber
 }
 
+export enum ArbitrageType {
```
XavierEkkel, line 35: nit:
```suggestion
export enum ArbitrageStrategy {
```
to be consistent with function name

apps/smart-contracts/core/test/integration/ArbitrageTrading.test.ts
```diff
@@ -544,5 +547,69 @@ describe('=> Arbitrage Trading', () => {
         expect(prices.shortPoolPrice).eq(expectedShortPoolPrice)
       })
     })
+
+    describe('# getStrategyToUse', () => {
+      it("selects 'buyAndRedeem()' if L + S < (1 - spread)", () => {
```
XavierEkkel, line 552: nit: `'buy and redeem'` or `BUY_AND_REDEEM`
since it's not calling that function here

similarly for other tests

XavierEkkel (approved):
LGTM after nits

PR #798, opened 2023-01-24
feat: deposit summary without fees

chrisling-dev:
https://www.notion.so/prepo/FE-Project-Management-d3878c29e9534234a50f4f8838525bee?p=852ee0a3c0a74f62941912a07ba2d4fc&pm=s


apps/frontend/core-dapp/src/features/deposit/DepositPage.tsx
```diff
@@ -51,6 +52,7 @@ const DepositPage: React.FC = () => {
         showBalance
       />
       <DepositTransactionSummary />
```
XavierEkkel, line 54: should this be removed?

chrisling-dev: In the next one when I add deposit button, it will be replaced by deposit button

PR #802, opened 2023-01-26
feat: add additional return values to arbitrage calls

ramenforbreakfast:
So we need to calculate slippage values for the swap based on the quote we retrieve. For simplicity, we are just quoting using our `ArbitrageBroker` rather than the Uniswap contracts directly.

However, the `ArbitrageBroker` only returns the net profit, and while I could separately re-grab a quote from Uniswap directly to calculate slippage parameters, it is better if we calculate slippage off the amount returned from the `ArbitrageBroker`.

Hence, I decided the best course of action was to add a tiny overhead and return the values we are already grabbing anyway in the `ArbitrageBroker`, so that we can use these values to calculate slippage parameters.


XavierEkkel (approved):
LGTM

PR #803, opened 2023-01-26
feat: add function that returns arbitrage estimates

ramenforbreakfast:
This is the function that will be called within the while loop (parent function) that increments position sizes. Every new position size will be passed into this function to grab the estimated profit and output values based on that position size.

Also includes a ArbitrageBroker attach fixture, since the broker, unlike other contracts that we "attach" to in this file, will be used across various functions (including the test).

The fake ArbitrageBroker is used to simplify the tests.


apps/smart-contracts/core/test/integration/ArbitrageTrading.test.ts
```diff
@@ -611,5 +616,68 @@ describe('=> Arbitrage Trading', () => {
...
+          attachArbitrageBroker(ethers.provider, fakeArbitrageBroker.address),
+          core.markets[TEST_NAME_SUFFIX].address,
+          expectedTradeSize
+        )
+
+        // Verification must be done value by value because equality between structs is not supported
+        expect(estimateReturnValues.tradeSize).eq(expectedTradeSize)
+        expect(estimateReturnValues.profit).eq(expectedProfit)
+        expect(estimateReturnValues.collateralToBuyLong).eq(expectedCollateralToBuyLong)
+        expect(estimateReturnValues.collateralToBuyShort).eq(expectedCollateralToBuyShort)
```
XavierEkkel, line 647: add expects that collateralForSellingLong/Short are undefined
similarly for other test

XavierEkkel (approved):
LGTM after comments

PR #805, opened 2023-01-26
feat: add slippage factor calculator with error handling

ramenforbreakfast:
Since there needs to be error handling with the slippage factor passed in, I thought it best to split it out into its own function for ease of testing. Comments should provide sufficient context.

Note that `else if` is not used because our linter does not allow `else if`, if the conditional block returns a value anyway.


XavierEkkel (approved):
LGTM

PR #806, opened 2023-01-26
feat: add estimate adjustor to account for gas costs and slippage

ramenforbreakfast:
This adjusts a raw estimate before handing off the to the searching loop to decide if the position size is more profitable than the prior position.

These tests are quite verbose, but are all very similar, mostly just because we need to verify that all the values in the estimate returned are correct. Comments should provide sufficient context.


apps/smart-contracts/core/scripts/ArbitragePools.ts
```diff
@@ -224,3 +224,62 @@ export const getSlippageFactor = (
   }
   throw new Error('Cannot get estimate for INSUFFICIENT_SPREAD')
 }
+
+/**
+ * `gasCostBuffer` is the Collateral equivalent we are willing to account
+ * for gas fees, e.g. 1 ether would be 1 Collateral in gas cost buffer.
```
XavierEkkel, line 230: nit:
```suggestion
 * `gasCostBuffer` is the Collateral-denominated gas fee to be accounted for
 * e.g. 1 ether would be 1 Collateral in gas cost buffer
```

apps/smart-contracts/core/test/integration/ArbitrageTrading.test.ts
```diff
@@ -740,5 +741,300 @@ describe('=> Arbitrage Trading', () => {
         )
       })
     })
+
+    describe('# getAdjustedEstimate', () => {
+      const TEST_MAX_SLIPPAGE = '0.01' // 1%
```
XavierEkkel, line 746: nit:
```suggestion
      const TEST_MAX_SLIPPAGE = '0.01' // 1%
      
```

ramenforbreakfast:
I have resubmitted simplified tests that focus only on path verification.

XavierEkkel (approved):
LGTM

PR #807, opened 2023-01-26
feat: add calculator for increasing trade size

ramenforbreakfast:
This will be called at the end of the while loop that searches for position sizes. Separating this out allows us to easily change this function only if we plan to change how we grow positions.

Did not add tests to verify values since it would just be rewriting the function and there isn't much to this.


apps/smart-contracts/core/scripts/ArbitragePools.ts
```diff
@@ -283,3 +283,15 @@ export const getAdjustedEstimate = (
...
   throw new Error('Cannot get estimate for INSUFFICIENT_SPREAD')
 }
+
+/**
+ * `growthFactor` is a string because BigNumber does not support decimal
+ * inputs for calculations. Instead, we convert a float string,
+ * e.g. '1.05'(1.05x), to a 18 decimal factor.
+ */
+export const getNextTradeSize = (currentTradeSize: BigNumber, growthFactor: string): BigNumber => {
+  if (parseEther(growthFactor).lt(0)) {
```
XavierEkkel, line 293:
```suggestion
  if (parseEther(growthFactor).lte(1)) {
```
if <1, then it will shrink. if =1 it won't grow

add tests for =0, =1, and right below 1

XavierEkkel (approved):
LGTM after comment

PR #811, opened 2023-01-28
feat: add Sinon mocking library

ramenforbreakfast:
The Sinon mocking library is a Mocha-compatible mocking library. Originally I was looking at Jest because smock is a Jest based library, but realized Hardhat is a Mocha based testing framework so using Sinon is an easier option.


XavierEkkel (approved):
LGTM

PR #813, opened 2023-01-28
feat: add getIdealTradeSize function for arbitrage script

ramenforbreakfast:
This is the last major function (besides executing the trade), before we put everything together. Utilized mocking libraries to ensure tests are concise and only test what is important to this function.


ramenforbreakfast:
I have revised the testcases with the following changes from what was discussed in Discord:
- Use mocked values for `getNextTradeSize` rather than relying on internal logic to determine iterations.
- Modified testcase for when trade size > max, to also implicitly test when trade size = max. This is done by having the first trade increase set to MAX, then second increase set to over the MAX.
- Expanded assertions to verify the entire estimate, rather than just profit (except the first two since those are just the default zero value estimates).

XavierEkkel (approved):
LGTM after adjusting tests to test all estimate fields and number of times certain mocks/fakes were called

PR #815, opened 2023-01-30
feat: add executeTradeIfProfitable to further isolate trade logic

ramenforbreakfast:
Originally had put this logic in the parent `arbitragePools`, but decided to separate it out into a different function for both ease of testing and simplicity, so that the parent function entirely consists of just internal function calls and no logic to test.


XavierEkkel (approved):
LGTM

PR #819, opened 2023-01-30
feat: withdraw button

chrisling-dev:


apps/frontend/core-dapp/src/stores/CollateralStore.ts
```diff
@@ -31,7 +31,6 @@ export class CollateralStore extends Erc20Store {
   depositHash?: string
   depositing = false
   withdrawHash?: string
-  withdrawing = false
```
chrisling-dev, line 34: I'm removing this because I think it's better to handle feature states (like loading state, disabled, and etc) in feature store, and leave contract stores do only contract interaction stuffs (like getting a balanceOf, making an SC call)

chrisling-dev (commented):
self reviewed

apps/frontend/core-dapp/src/features/withdraw/WithdrawButton.tsx
```diff
@@ -0,0 +1,37 @@
...
+    withdrawButtonInitialLoading,
+    withdrawUILoading,
+    withdrawalDisabled,
+  } = withdrawStore
+
+  const buttonText = useMemo(() => {
+    if (withdrawButtonInitialLoading) return ''
+    if (insufficientBalance) return 'Insufficient Balance'
+    return 'Withdraw'
+  }, [insufficientBalance, withdrawButtonInitialLoading])
```
MeLlamoPablo, line 21: Shouldn't this be localized?

apps/frontend/core-dapp/src/features/withdraw/WithdrawStore.ts
```diff
@@ -15,14 +16,36 @@ export class WithdrawStore {
   }
 
   // eslint-disable-next-line require-await
```
MeLlamoPablo, line 18:
```suggestion
```

apps/frontend/core-dapp/src/features/withdraw/WithdrawStore.ts
```diff
@@ -15,14 +16,36 @@ export class WithdrawStore {
...
+    )
+      return
+
+    this.withdrawing = true
+    const { error } = await this.root.preCTTokenStore.withdraw(this.withdrawalAmountBN)
+
+    if (error) {
+      this.root.toastStore.errorToast('Withdrawal failed', error)
+    } else {
+      this.root.toastStore.successToast('Withdrawal was successful 🎉')
```
MeLlamoPablo, line 33:
```suggestion
      this.root.toastStore.errorToast('The withdrawal failed', error)
    } else {
      this.root.toastStore.successToast('The withdrawal was successful 🎉')
```

Some minor nit on the copies, it feels more natural to me, but feel free to disregard.

apps/frontend/core-dapp/src/features/withdraw/WithdrawStore.ts
```diff
@@ -15,14 +16,36 @@ export class WithdrawStore {
...
-      : { success: false }
+  async withdraw(): Promise<void> {
+    if (
+      this.insufficientBalance ||
+      this.withdrawalAmountBN === undefined ||
+      this.withdrawalAmountBN.eq(0)
+    )
+      return
+
+    this.withdrawing = true
```
MeLlamoPablo, line 27: Why do we not use `runInAction` here but use it below?

apps/frontend/core-dapp/src/features/withdraw/WithdrawStore.ts
```diff
@@ -35,12 +58,13 @@ export class WithdrawStore {
...
   get withdrawalDisabled(): boolean {
-    const { tokenBalanceRaw } = this.root.preCTTokenStore
     return (
-      tokenBalanceRaw === undefined ||
+      this.receivedAmountBN === undefined ||
       !this.withdrawalAmountBN ||
       this.withdrawalAmountBN.lte(0) ||
-      this.withdrawalAmountBN.gt(tokenBalanceRaw)
+      this.insufficientBalance === undefined ||
+      this.insufficientBalance ||
```
MeLlamoPablo, line 66: This kind of logic feels confusing to me:

* `insufficientBalance = undefined` means `withdrawalDisabled = true`
* `insufficientBalance = false` means `withdrawalDisabled = false`
* `insufficientBalance = true` means `withdrawalDisabled = true`

Two falsy cases evaluate each to different values, and I think this can lead to confusion and bugs. I don't like nullable booleans in general, wdyt about making `insufficientBalance` an union of literals?

```ts
get sufficientBalance(): 'is-sufficient' | 'is-insufficient' | 'no-input'
```

Or maybe just assume that no user input or data not present means `insufficientBalance = true`, I'm not sure if that has implications on the current logic

apps/frontend/core-dapp/src/features/withdraw/WithdrawStore.ts
chrisling-dev: So with mobx, most operations don't need runInAction until we use asynchronous functions. For example you could refer to the doc here: https://mobx.js.org/actions.html#asynchronous-actions (switch to `async/await + runInAction` tab)

The author wrapped any updates with runInAction *after* an await operation.

apps/frontend/core-dapp/src/features/withdraw/WithdrawStore.ts
chrisling-dev: so in most cases, we use undefined to indicate that something is loading (because they're not loaded yet so it's undefined, e.g. an erc20 balance would go from undefined -> 123.45678)

In this case, we're trying to load the user's preCT Token balance, and this logic is trying to do:

if insufficientBalance is loading -> disable button
if insufficientBalance is true -> disable button

We can refactor this to insufficientBalance !== false, but I feel like the 2 lines approach is more readable

apps/frontend/core-dapp/src/features/withdraw/WithdrawButton.tsx
chrisling-dev: nope, we'll add later but not priority now

apps/frontend/core-dapp/src/features/withdraw/WithdrawStore.ts
MeLlamoPablo: How would you indicate the absence of a value after loading, `null = no value` and `undefined = loading`?

chrisling-dev: we haven't encounter any use case where the value would be absence (e.g. it's all BigNumber, boolean, timestamp and etc), but yes in the case of a real empty value, we will use null

apps/frontend/core-dapp/src/features/withdraw/WithdrawStore.ts
```diff
@@ -14,15 +15,36 @@ export class WithdrawStore {
...
+    )
+      return
+
+    this.withdrawing = true
+    const { error } = await this.root.preCTTokenStore.withdraw(this.withdrawalAmountBN)
+
+    if (error) {
+      this.root.toastStore.errorToast('The withdrawal failed', error)
+    } else {
+      this.root.toastStore.successToast('The withdrawal was successful 🎉')
```
XavierEkkel, line 32:
```suggestion
      this.root.toastStore.successToast('Withdrawal successful 🎉')
```

XavierEkkel (approved):
LGTM from UI perspective

PR #821, opened 2023-01-31
feat: Autotask starting directory with copied over ArbitragePools script

ramenforbreakfast:
This is just a bare Autotasks directory with copied over files and the default packages. I threw it in `external` since it needs to exist in its own environment, similar to `yield`.

The packages included are the same ones available by default to OZ Autotasks.

Lint rules were carried over so our format carries over for dev purposes.

Currently the `exportedArtifacts` containing our contract JSONs is just copied over as-is, currently I am figuring out how to bundle them into a build with `Webpack` so that they can be output into a single build file for Autotasks.


ramenforbreakfast:
Closing due to decision to rearrange submission.

PR #824, opened 2023-02-01
feat: add skeleton for production harness

ramenforbreakfast:
Our liquidity helper tasks will use the production harness so I am providing a basic skeleton here in the mean time that will later have additional functions for our deployment scripts + other tasks.


XavierEkkel (approved):
LGTM

PR #825, opened 2023-02-01
core:fix: make helper functions independent of ethers

ramenforbreakfast:
So something when writing deployment scripts and tasks, is that if it sources code elsewhere in the directory, that external code cannot directly import `hardhat`. This is because when hardhat runs scripts + tasks, it initializes itself along with the scripts + tasks, thus creating a chicken and egg problem. I already accounted for this in most places, but forgot to do so in some. This PR fixes the last few helpers that directly relied on a `ethers` imported from `hardhat` rather than being agnostic to the `ethers` used.


XavierEkkel (approved):
LGTM

PR #828, opened 2023-02-01
fix: remove helper dependencies within test subfolder, move to shared utilities and constants

ramenforbreakfast:
This should be the last of the decoupling technical debt (including the last PR) regarding the problematic dependency our shared helpers and harnesses have with `hardhat`'s `ether`'s. This dependency prevents us from compiling and running scripts we develop that run outside tests.

The prior PR was focused on specifically removing the dependency on `hardhat`'s injected `ethers` from helpers and harness functions.

This PR is focused on usage of constants and utility functions that were declared within the `test` subfolder, which is problematic because almost everything inside the `test` subfolder uses the `hardhat` injected `ethers`.

These constants and functions should not reside only within the `test` subfolder and should be more freely accessible.
- core related constants, such as `FEE_DENOMINATOR` and `MARKET_FEE_LIMIT` have been moved to a `core.ts` file in `prepo-constants`.
- `grantAndAcceptRole` and `batchGrantAndAcceptRole`, highly generalized role acceptance functions, were moved to our existing `prepo-hardhat` utils package, where are shared smart contract utilities reside.


XavierEkkel (approved):
LGTM

PR #831, opened 2023-02-02
feat: uniswap SDK based liquidity estimation helpers

ramenforbreakfast:
So after some research, it seems Uniswap's own SDK also estimates liquidity amounts simply using their own math, rather than any sort of static call functionality. This is very similar to the sorts of math helpers we were using for calculating liquidity positions on Trident.

However, we can rely on much of Uniswap's existing SDK so that we avoid having to define most of these functions ourselves. Using as many parts of the existing SDK as I could, this PR contains a function that will allow us to know ahead of time amounts to provide based on current pool parameters and approximate bounds.

In a production scenario, we would most likely just pass the market lower and upper bounds directly for simplicity(since they are wei values) to this function, and however much of each token we want to provide.

As per comments, `getAmountsNeededForLiquidity` is based on an existing Uniswap SDK function that needed to be modified, and I have posted a link to the source in the comments.

I think it is appropriate that these be submitted in the same PR since they are tightly correlated and one is an adaptation of existing Uniswap code.


ramenforbreakfast:
Closing due to preference on using Gnosis's Safe App

PR #833, opened 2023-02-02
feat: add deployment name tree for core contracts

ramenforbreakfast:
You will notice that the existing tree for token contracts is called `DEPLOYMENT_NAMES` already. To save both of us time, I'm just going to leave it untouched, with the initial core ones labeled `CORE_DEPLOYMENT_NAMES`.

The `TokenSender` we deploy will be labeled `PPOTokenSender`, since it is a generic contract, but we will be sending `PPO`, and this contract will be used by both markets and collateral for rebates.


packages/prepo-constants/src/deployments.ts
```diff
@@ -35,3 +35,59 @@ export const DEPLOYMENT_NAMES = {
 } as const
 
 export type DeploymentNames = typeof DEPLOYMENT_NAMES
+
+export const CORE_DEPLOYMENT_NAMES = {
```
XavierEkkel, line 39: why not just use a single `DEPLOYMENT_NAMES`, since there are no name conflicts?

XavierEkkel (approved):
LGTM after question

packages/prepo-constants/src/deployments.ts
ramenforbreakfast: Mostly for organizational purposes, but can combine them.

PR #834, opened 2023-02-02
feat: add Collateral deployment script

ramenforbreakfast:
So in hindsight from what we've learned from prior deployments, especially with a set of contracts this large, it's going to be much better if we write our deployment scripts only around deploying, no configuration.

Since none of our contracts rely on passing dependent contracts into a constructor, this is doable, and means we do not have to write our deployment scripts that form a web of dependencies. Configuration can be done altogether in a different script that runs after all contracts are deployed.


apps/smart-contracts/core/deploy/Collateral.ts
```diff
@@ -0,0 +1,121 @@
...
+  ethers,
+  upgrades,
+}: HardhatRuntimeEnvironment): Promise<void> {
+  const { save, getOrNull, getArtifact } = deployments
+  const deployer = (await ethers.getSigners())[0]
+  const currentChain = Number(await getChainId()) as ChainId
+  const currentNetwork = getNetworkByChainId(currentChain)
+  console.log(
+    `Running Collateral deployment script against ${currentNetwork} with ${deployer.address} as the deployer`
+  )
```
XavierEkkel, line 30: this log is missing from the other non-upgradable deploy scripts

apps/smart-contracts/core/deploy/Collateral.ts
```diff
@@ -0,0 +1,121 @@
...
+  /**
+   * Make sure this script is not accidentally targeted towards a production environment.
+   * This can be temporarily removed if deploying to prod.
+   */
+  assertIsTestnetChain(currentChain)
+  const usdcAddress = getPrePOAddressForNetwork('USDC', currentNetwork.name, process.env.USDC)
+  const collateralFactory = await ethers.getContractFactory('Collateral')
+  const deploymentName = CORE_DEPLOYMENT_NAMES.preUSDC.name
+  const existingDeployment = await getOrNull(deploymentName)
+  if (!existingDeployment) {
```
XavierEkkel, line 40: could the boilerplate structure of this part of upgradable deploy scripts be shared to reduce duplication?

XavierEkkel (approved):
LGTM after comments (can save for future PRs)

apps/smart-contracts/core/deploy/Collateral.ts
ramenforbreakfast: yes I can try to separate out this piece

apps/smart-contracts/core/deploy/Collateral.ts
ramenforbreakfast: It's in there, just instead of Collateral, it has a contractName variable since its a generic function.

apps/smart-contracts/core/deploy/Collateral.ts
ramenforbreakfast: I think its better to just create a externalized generic function, but I think we can add it as a backlog item since theres only 2 upgradeable scripts.

PR #835, opened 2023-02-03
feat: add DepositHook deployment script

ramenforbreakfast:
Other non-upgradeable contract deployment scripts will follow this format.


ramenforbreakfast:
Closing due to refactor to use a new reusable deployment function for non-upgradeable contracts.

PR #836, opened 2023-02-03
feat: add generic non-upgradeable contract deployer

ramenforbreakfast:
Since we are shifting to a model where deployment happens before configuration, this makes it much easier to write deployment scripts and much of the logic for deploying can be externalized into a helper function.

Upgradeable contracts, due to complexities around proposals and upgrade mechanisms, would require a lot more effort to externalize and we will not be doing so for now.

I tried to move this to a shared package, but this relies on types that hardhat injects based on the project's environment, thus making it not shareable friendly.


XavierEkkel (approved):
LGTM

PR #837, opened 2023-02-03
feat: add deployment scripts that do not have args

ramenforbreakfast:
I have bundled all no-argument, non-upgradeable, deployment scripts since they only differ by their function names, tags, and name of the contract itself, so really makes no sense to split out into different PRs.


XavierEkkel (approved):
LGTM

PR #838, opened 2023-02-04
feat: add core deployment scripts that take args

ramenforbreakfast:
I have bundled these remaining non-upgradeable deployment scripts. Both ArbitrageBroker and DepositTradeHelper take in the same arguments, and TokenSender takes in PPO, so nothing too complicated here.


apps/smart-contracts/core/deploy/TokenSender.ts
```diff
@@ -0,0 +1,38 @@
...
+  hre: HardhatRuntimeEnvironment
+): Promise<void> {
+  const { getChainId } = hre
+  const currentChain = Number(await getChainId()) as ChainId
+  const currentNetwork = getNetworkByChainId(currentChain)
+  /**
+   * Make sure this script is not accidentally targeted towards a production environment.
+   * This can be temporarily removed if deploying to prod.
+   */
+  assertIsTestnetChain(currentChain)
```
XavierEkkel, line 26: could these boilerplate lines be shared across all deploy functions?

XavierEkkel (approved):
LGTM after comment (can address in a future PR)

apps/smart-contracts/core/deploy/TokenSender.ts
ramenforbreakfast: I can add a boolean argument to disable the testnet guard, moving it to the external function

PR #839, opened 2023-02-04
feat: prePOMarketFactory deployment script

ramenforbreakfast:
Very similar to the Collateral upgrade script, but does not include the more complex overrides we need to apply for `Collateral`, due to it having immutable args. `PrePOMarketFactory` is much simpler and does not need these overrides.

Comments detailing specifics about using the upgrades library and hardhat-deploy are omitted because they already exist in `Collateral.ts`. If really necessary, we can spend additonal time trying to figure out how to externalize upgradeable deployments as well into a generic function.


XavierEkkel (approved):
LGTM

PR #842, opened 2023-02-06
refactor:core: move roles to constants rather than relying on promises

ramenforbreakfast:
So before I write the configuration task, I was looking into how we could reduce replicating code since we have to create configuration tools for both the deployer (executing as a EOA signer) and a OZ proposal to a multisig.

I was hoping to redesign some things like our role assigners, to return transaction data, rather than executing a transaction, so that it could be passed to either a proposal or signer.

Unfortunately, this is not doable due to how the proposal API is structured, **but what we can do**, is refactor our role management code to rely on constants rather than asynchronous calls to get roles.

By relying on constants, we can just pass in the roles directly to either function, instead of asynchronously grabbing them, which will make the code much less verbose (and less fragile due to not relying on contract calls). The way David had previously written the function, it asynchronously retrieve the roles from the contract, rather than pre-defining them. I have changed this so that it accepts strings rather than promises.

I have also, since we need to retroactively change some test suites to use the modified role assigner, defined other role assigners that were missing.


apps/smart-contracts/core/test/unit/PrePOMarket.test.ts
```diff
@@ -46,24 +46,12 @@ describe('=> prePOMarket', () => {
...
   beforeEach(async () => {
     ;[deployer, user, treasury] = await ethers.getSigners()
     collateralToken = await testERC20Fixture('prePO USDC Collateral', 'preUSD', 18)
     await collateralToken.mint(deployer.address, MOCK_COLLATERAL_SUPPLY)
     prePOMarketFactory = await prePOMarketFactoryFixture()
-    await batchGrantAndAcceptRoles(prePOMarketFactory, deployer, deployer, [
-      prePOMarketFactory.CREATE_MARKET_ROLE(),
-      prePOMarketFactory.SET_COLLATERAL_VALIDITY_ROLE(),
-    ])
+    await roleAssigners.assignPrePOMarketFactoryRoles(deployer, deployer, prePOMarketFactory)
```
XavierEkkel, line 54: have added a ticket to modify `grantAndAcceptRole()` such that it skips it if admin+nominee are identical

apps/smart-contracts/core/test/unit/PrePOMarket.test.ts
```diff
@@ -202,7 +190,7 @@ describe('=> prePOMarket', () => {
   describe('# setFinalLongPayout', () => {
     beforeEach(async () => {
```
XavierEkkel, line 191: each of these `beforeEach` cases are the same - just move these two lines into the parent `beforeEach`?

packages/prepo-constants/src/index.ts
```diff
@@ -6,3 +6,4 @@ export * from './links'
 export * from './addresses'
 export * from './deployments'
 export * from './core'
+export * from './roles'
```
XavierEkkel, line 9: nit: these should be alphabetically sorted - doesn't the linter correct this?

XavierEkkel (approved):
LGTM after nits

apps/smart-contracts/core/test/unit/PrePOMarket.test.ts
ramenforbreakfast: Ah no, this won't work. The `roleAdmin` of a `role` is not necessarily the `role` itself.

This is true for the base case, which is that the `roleAdmin` for ALL roles is the "0" `role` (initialized to the deployer), but that does not mean that the "0" role holds those roles, it can just assign people to them.

packages/prepo-constants/src/index.ts
ramenforbreakfast: No, let me add a ticket to investigate why that is

apps/smart-contracts/core/test/unit/PrePOMarket.test.ts
ramenforbreakfast: I did not modify it because this should be fixed to only assign the role that pertains to the testcase, like we have done else where.

PR #843, opened 2023-02-06
fix: add missing harness entries for contracts

ramenforbreakfast:
This PR includes a bunch of small and medium size changes to the harnesses that need to be done prior to configuration scripts. I decided to keep these in one PR since the Type changes are directly aligned with the addition of missing entries and it would not make sense to separate them.

Added a `allowedMsgSenders` property for `TokenSender` and `DepositRecord` to store the `AccountList` associated with them, and adding entries for `arbitrageBroker` and `depositTradeHelper`. Also added a `FixedUintValue` entry for the `TokenSender` price oracle.

Types were also changed because it was getting very confusing with the old naming of `<Contract>With<TheseThings>`. Instead, I have switched to just using `Extended<Contract>` across the board for simplicity. Harness types have been normalized so that `Base` harness functions can expect predictable types.

Changes to the `init` portion of the `ProdCore` harness: 
- Deployment names now use the constants defined earlier, rather than placeholder strings.
- `marketFactory` was moved to the prod harness out of `Base` since `marketFactory` is not used in mock stacks.
- `rewardToken` was modified to grab the PPO address associated with a given network, since there is technically no `PPO` deployment in the `core` repo.
- `getContractOrNull` is now used for contracts that will be deployed after the initial deployment (namely `AccountList` contracts), since one might not exist prior to configuration.


XavierEkkel (approved):
LGTM

PR #844, opened 2023-02-07
feat: add FixedUintValue to core repo

ramenforbreakfast:
So I forgot we needed a fixed price oracle for the `TokenSender`, so we need to add it to the core repo since we can't deploy it unless its defined within the contract. I've added it in there along with references to it since it's not that involved to add a deployment script and entry for it.


XavierEkkel (approved):
LGTM

PR #845, opened 2023-02-07
feat:core: add role assignment for harness

ramenforbreakfast:
These functions will be called by the configuration task. I separated them out into ones that will be guaranteed to be shared by both mock and production and ones that will only be in production to reduce code repetition in the future.


ramenforbreakfast:
Added an improvement to first identify which roles an account does not have, and only assign those ones.

XavierEkkel (approved):
LGTM

PR #848, opened 2023-02-07
refactor: simplify date utils

MeLlamoPablo:
I found the `date-utils.ts` file quite confusing, so I took the liberty to make a small refactor.

My biggest problem with it was that some helpers were designed to accept `number | string`. This may seem convenient, but it used `new Date` to transforms the strings to dates. In my opinion this is a no-no. [As explained in MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date#date_string): 

> Date-only strings (e.g. `"1970-01-01"`) are treated as UTC, while date-time strings (e.g. `"1970-01-01T12:00"`) are treated as local. You are therefore also advised to make sure the input format is consistent between the two types.

The above quirk has resulted in hard to debug issues in projects that I've worked on in the past. If we want to parse strings as dates, we should use `date-fns/parseISO` for standard strings or `date-fns/parse` for strings with custom format.

The good thing is that the project didn't actually make use of parsing strings! Those helpers were only called with `number`, so I've removed the `string` case and the tests corresponding to it, since we don't need to test a case we don't use in practice.

This has two benefits:

* The size and complexity of `date-utils.ts` is greatly reduced.
* We don't need to worry about `date-fns/format` throwing due to invalid dates, because all numbers are valid dates.

-----

I have another issue: we handle two kinds of timestamps within the project - seconds and milliseconds. I think this is confusing and we should only handle milliseconds (which is what `date-fns` accepts, and there is little reason to perform operations on dates without `date-fns`). And if we really, really need to support both, we should use branded types.

But I think this can wait for a better time. What do you think?


chrisling-dev (approved):
LGTM! nice work cleaning this! 
- Agree that we should only handle `number` timestamps and not string, also don't remember anywhere we're using string with these functions
- Yes we have functions for ms or seconds, the reason for that is that smart contracts usually return timestamp in seconds instead of ms, so it was more straightforward to do `getDateFromSeconds(contract.dateTimeValue)` than `getDateFromMs(contract.dateTimeValue * 1000)`. But if it's causing more confusion than convenient in formatting timestamp from SC, I'm happy to change these to just handle MS in a separate PR, low priority tho

MeLlamoPablo:
> But if it's causing more confusion than convenient in formatting timestamp from SC, I'm happy to change these to just handle MS in a separate PR, low priority tho

Yeah maybe there's a way to normalize the data from SC from seconds to milliseconds, but I agree it's low prio

PR #849, opened 2023-02-07
feat:core: add support for recipient collateral

nmimran99:
feat!: allow redeem for recipient != withdrawer

This PR adds a address recipient parameter to the withdraw function in the Collateral.sol contract. It allows a user to specify a recipient other then himself.

This change is a breaking change since it requires a different call parameters to the function and also changes the event emitted from the withdraw process (Withdraw event)


ramenforbreakfast (changes requested):
@nmimran99 You forgot to remove the PrePOMarket changes from this branch, since this is a PR only for the changes to Collateral, these need to be removed.

PR #851, opened 2023-02-08
feat: add configuration helpers for the base stack

ramenforbreakfast:
Added all the configuration helpers for the `Base` stack, these are configurators for all contracts that would be present in either harness (this essentially is just every contract except `PrePOMarketFactory`, `ArbitrageBroker`, and `DepositTradeHelper`).

These configuration helpers follow essentially the same format, first assign contracts that are connected to them, then assign values if one was provided.

A `setContractIfNotSet` utility function was provided to reduce code duplication and to ensure txs are not submitted for contracts already connected.


apps/smart-contracts/core/harnesses/base.ts
```diff
@@ -150,4 +150,216 @@ export abstract class Base {
...
+      await sendTxAndWait(await this.tokenSender.price.connect(signer).set(price))
+    if (priceMultiplier !== undefined)
+      await sendTxAndWait(
+        await this.tokenSender.connect(signer).setPriceMultiplier(priceMultiplier)
+      )
+    if (scaledPriceLowerBound !== undefined)
+      await sendTxAndWait(
+        await this.tokenSender.connect(signer).setScaledPriceLowerBound(scaledPriceLowerBound)
+      )
+    if (allowedMsgSenders) {
```
XavierEkkel, line 357:
```suggestion
    if (allowedMsgSenders !== undefined && allowedMsgSenders.length > 0) {
```

XavierEkkel: same for `allowedDepositors`

XavierEkkel (approved):
LGTM after comments

PR #852, opened 2023-02-08
feat: implement the new History design on `/portfolio-next`

MeLlamoPablo:
This PR adds the History tab component to match the new design:

Screenshot 2023-02-08 at 19 22 11

For ease of review, you can skip the first commit which just duplicates the `History` component.

# FE PR Checklist

    - [x] iPhone 12 Pro
    - [x] iPad Air
- [x] (if browser specific logic) Tested on Brave, Firefox, Safari


apps/frontend/core-dapp/src/features/history/HistoryTransaction.tsx
```diff
@@ -0,0 +1,103 @@
...
+        <Flex gap={4}>
+          {nameRedirectUrl ? (
+            <Link href={nameRedirectUrl} passHref>
+              <PositionName as="a">{name}</PositionName>
+            </Link>
+          ) : (
+            <PositionName>{name}</PositionName>
+          )}
+
+          {eventType && <PositionLabel positionType={eventType} />}
```
chrisling-dev, line 82: Can update this component's font-size to 14px to match design, the `PositionLabel` is only used once and that use case will be removed after your changes anyways

apps/frontend/core-dapp/src/features/history/HistoryTransaction.tsx
```diff
@@ -0,0 +1,103 @@
...
+      </Flex>
+      <Flex flexDirection="column" alignItems="end" ml="auto" gap={6}>
+        <Link href={`${NETWORKS[network.name].blockExplorer}/tx/${transactionHash}`} passHref>
+          <EventLink $event={event} target="_blank" rel="noreferrer noopener nofollow">
+            {event}{' '}
+            <Icon
+              name="arrow-up-right"
+              width={spacingIncrement(14)}
+              height={spacingIncrement(14)}
+            />
```
chrisling-dev, line 94: I think can just use the html nearr entity to make it simpler and more readable, wdyt?

Ref: https://www.w3schools.com/charsets/ref_html_entities_n.asp

```suggestion
            {event} &#x02197;
```

chrisling-dev (changes requested):

apps/frontend/core-dapp/src/features/history/HistoryTransaction.tsx
```diff
@@ -0,0 +1,103 @@
...
+    nameRedirectUrl = {
+      pathname: '/trade',
+      query: {
+        action: event === 'Opened' ? 'open' : 'close',
+        direction: eventType,
+        marketId,
+      },
+    }
+  if (event === 'Deposited') nameRedirectUrl = '/portfolio/deposit'
+  if (event === 'Withdrawn') nameRedirectUrl = '/portfolio/withdraw'
```
chrisling-dev, line 67: I guess there isn't a case where nameRedirectUrl would be undefined + since event won't change for each transaction, we can do this in a memo so the logic doesn't get computed again:

```suggestion
  const nameRedirectUrl = useMemo(() => {
    switch (event) {
      case 'Deposited':
        return '/portfolio/deposit'
      case 'Withdrawn':
        return '/portfolio/withdraw'
      default:
        return {
          pathname: '/trade',
          query: { action: event === 'Opened' ? 'open' : 'close', direction: eventType, marketId },
        }
    }
  }, [event, eventType, marketId])
```

apps/frontend/core-dapp/src/features/history/HistoryTransaction.tsx
```diff
@@ -0,0 +1,103 @@
...
+      <Flex flexDirection="column" alignItems="start">
+        <Flex gap={4}>
+          {nameRedirectUrl ? (
+            <Link href={nameRedirectUrl} passHref>
+              <PositionName as="a">{name}</PositionName>
+            </Link>
+          ) : (
+            <PositionName>{name}</PositionName>
+          )}
+
```
chrisling-dev, line 81: Then this can be simplified

```suggestion
          <Link href={nameRedirectUrl} passHref>
            <PositionName as="a">{name}</PositionName>
          </Link>
```

chrisling-dev (changes requested):
Can simplify redirect url, marketId won't ever be undefined for Opened/Closed event, even if it's undefined I think we can still redirect for action and direction

apps/frontend/core-dapp/src/utils/date-utils.ts
```diff
@@ -33,7 +38,10 @@ export const getUTCStartOfDay = (ms: number): number => {
 export const getFullDateFromMs = (ms: number): string => format(ms, DATE_FORMAT)
 
 export const getFullDateShortenMonthFromMs = (ms: number): string =>
-  format(ms, DATE_FORMAT_SHORTEN_MONTH)
+  format(ms, isCurrentYear(ms) ? DATE_TIME_FORMAT : DATE_FORMAT_SHORTEN_MONTH)
```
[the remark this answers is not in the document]
MeLlamoPablo: Reverted to the old format (it doesn't match the design tho)

IMO the tooltip approach is overkill. I think we should aim for simplicity in this case. Also, we would need to add a ℹ️ icon, which adds visual clutter, or the feature wouldn't be discoverable on mobile.

I think that if the user wants to know the time, they can just go to Etherscan.

apps/frontend/core-dapp/src/features/history/HistoryTransaction.tsx
MeLlamoPablo: Nice!

apps/frontend/core-dapp/src/features/history/HistoryTransaction.tsx
MeLlamoPablo: I agree with `useMemo` because it looks cleaner (it's its own function, so we can use return to alter the control flow), but not for the performance optimization. I think that would be premature optimization.

This is one of the examples why Kotlin is my favorite language, control statements are expressions so you can assign them to a variable 😆 

```kt
val nameRedirectUrl = when (event) {
  "Deposited" -> "/portfolio/deposit"
  "Withdrawn" -> "/portfolio/withdrawn"
  else -> // the location object
}
```

apps/frontend/core-dapp/src/features/position/PositionLabel.tsx
```diff
@@ -9,8 +9,8 @@ type Props = {
 
 const Wrapper = styled.div<Props>`
   color: ${({ theme, positionType }): string => theme.positionType[positionType]};
-  font-size: ${({ theme }): string => theme.fontSize.xs};
-  font-weight: ${({ theme }): number => theme.fontWeight.medium};
+  font-size: ${({ theme }): string => theme.fontSize.md};
```
chrisling-dev, line 12: Should this be sm (14px) instead of md?
```suggestion
  font-size: ${({ theme }): string => theme.fontSize.sm};
```

apps/frontend/core-dapp/src/features/history/HistoryTransaction.tsx
```diff
@@ -0,0 +1,80 @@
...
+    }
+  }, [event, eventType, marketId])
+
+  return (
+    <Flex justifyContent="start" gap={8} width="100%">
+      <Icon name={iconName} height={spacingIncrement(48)} width={spacingIncrement(48)} />
+      <Flex flexDirection="column" alignItems="start">
+        <Flex gap={4}>
+          <Link href={nameRedirectUrl} passHref>
+            <PositionName as="a">{name}</PositionName>
```
chrisling-dev, line 61: Could we make it such that when hovering on it, the name turns `primary` purple instead of blue? Same for EventLink

chrisling-dev (changes requested):
LGTM after nits

apps/frontend/core-dapp/src/features/history/HistoryTransaction.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+const EventLink = styled.a<{ $event: string }>`
+  align-items: center;
+  background: ${({ $event, theme }): string =>
+    theme.color[transactionTypeColors[$event].backgroundColor]};
+  border-radius: ${spacingIncrement(8)};
+  color: ${({ $event, theme }): string => theme.color[transactionTypeColors[$event].color]};
+  display: flex;
+  padding: ${spacingIncrement(2)} ${spacingIncrement(8)};
+`
```
chrisling-dev, line 30: ohh also add white-space: nowrap to this

```suggestion
const EventLink = styled.a<{ $event: string }>`
  align-items: center;
  background: ${({ $event, theme }): string =>
    theme.color[transactionTypeColors[$event].backgroundColor]};
  border-radius: ${spacingIncrement(8)};
  color: ${({ $event, theme }): string => theme.color[transactionTypeColors[$event].color]};
  display: flex;
  padding: ${spacingIncrement(2)} ${spacingIncrement(8)};
  white-space: nowrap;
`
```

chrisling-dev (changes requested):

apps/frontend/core-dapp/src/features/history/HistoryTransaction.tsx
```diff
@@ -0,0 +1,95 @@
...
+          <Link href={nameRedirectUrl} passHref>
+            <TransactionName>{name}</TransactionName>
+          </Link>
+
+          {eventType && <PositionLabel positionType={eventType} />}
+        </Flex>
+        <Label>{toUsd(usdValue)}</Label>
+      </Flex>
+      <Flex flexDirection="column" alignItems="end" ml="auto" gap={6}>
+        <Link href={`${NETWORKS[network.name].blockExplorer}/tx/${transactionHash}`} passHref>
```
chrisling-dev, line 84: There's a function from Web3Store that lets you do this:
```suggestion
        <Link href={web3Store.getBlockExplorerUrl(transactionHash)} passHref>
```

apps/frontend/core-dapp/src/features/history/History.tsx
```diff
@@ -11,7 +11,7 @@ const DEFAULT_COLORS: RecordButtonColors = {
   color: 'primaryWhite',
 }
 
-const buttonColors: { [key: string]: RecordButtonColors } = {
+export const buttonColors: { [key: string]: RecordButtonColors } = {
```
chrisling-dev, line 14: Should we duplicate this over instead of exporting since we will likely remove this file later when `/portfolio-next` is done?

apps/frontend/core-dapp/src/features/position/Position.tsx
```diff
@@ -11,7 +11,7 @@ import Skeleton from '../../components/Skeleton'
 
 const { toUsd } = numberFormatter
 
-const PositionName = styled.p`
+export const PositionName = styled.p`
```
chrisling-dev, line 14: ditto about duplicating

apps/frontend/core-dapp/src/features/history/HistoryTransaction.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const { network } = web3Store
+  const { event, eventType, iconName, marketId, name, timestamp, transactionHash, usdValue } =
+    historyEvent
+
+  const nameRedirectUrl = useMemo(() => {
+    switch (event) {
+      case 'Deposited':
+        return '/portfolio/deposit'
+      case 'Withdrawn':
+        return '/portfolio/withdraw'
```
chrisling-dev, line 61: We have a `Routes` enum that we can use here, you can import it from `/src/lib/routes`

```suggestion
      case 'Deposited':
        return Routes.Deposit
      case 'Withdrawn':
        return Routes.Withdraw
```

chrisling-dev (changes requested):
LGTM after comments

apps/frontend/core-dapp/src/features/history/History.tsx
MeLlamoPablo: Yeah I'm aware of that. I simply exported it to not introduce unnecessary lines of code in this PR

apps/frontend/core-dapp/src/features/position/Position.tsx
MeLlamoPablo: Actually this is from the new Position component. We are just using the same styles on the new Position component and the new History component.

chrisling-dev: ahh gotcha

apps/frontend/core-dapp/src/utils/date-utils.ts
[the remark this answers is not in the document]
chrisling-dev: could we have another formatting function then to make it match the design? it should just show [date] [month] [hour]:[sec]

chrisling-dev:
LGTM after the date is resolved, or feel free to create another ticket to do it separately

chrisling-dev (approved):
LGTM

MeLlamoPablo:
@XavierEkkel you commented but didn't approve or request changes, is this good to merge?

PR #853, opened 2023-02-08
feat:core: add deposit record user cap bypass for specific account

nmimran99:
feat!: add deposit record user cap bypass for specific account

Adds a bypass to the `recordDeposit()` function in `DepositRecord.sol` contract.  The bypass uses `AccountList.sol` implementation to determine if a certain account has rights to bypass the user cap for deposit.
Changes that were made:
1. `DepositRecord.sol` now inherits `AccountListCaller.sol` to allow access to an account list
2. a conditional was added to the user cap `require`, making sure that the `require` will only execute if the user is not on the account list. If the user is in the account list, the require never executes, which bypasses the user cap.

This is a breaking change since it requires to have a deployed `AccountList.sol` contract specifically for the usage of `DepositRecord`.


apps/smart-contracts/core/contracts/DepositRecord.sol
```diff
@@ -31,10 +37,12 @@ contract DepositRecord is IDepositRecord, SafeAccessControlEnumerable {
...
-      "User deposit cap exceeded"
-    );
+    if (!_accountList.isIncluded(user)) {
+      require(
+        amount + _userToDeposits[user] <= _userDepositCap,
+        "User deposit cap exceeded"
+      );
+    }
     _globalNetDepositAmount += amount;
     _userToDeposits[user] += amount;
```
ramenforbreakfast, line 47:
```suggestion
    _userToDeposits[user] += amount;
```
Since there's no need to keep track of an uncapped user's deposits, let's put this in the conditional so we don't needlessly sum up their deposits.

apps/smart-contracts/core/test/unit/DepositRecord.test.ts
```diff
@@ -113,6 +126,40 @@ describe('=> DepositRecord', () => {
       )
     })
 
+    it('correctly add to deposit totals and user totals for an uncapped user', async () => {
```
ramenforbreakfast, line 129: As per prior comment, should now be that `correctly adds to global, but not user deposit total if funder uncapped`

apps/smart-contracts/core/test/unit/DepositRecord.test.ts
```diff
@@ -113,6 +126,40 @@ describe('=> DepositRecord', () => {
...
+
+      expect(await depositRecord.getGlobalNetDepositAmount()).to.eq(
+        globalDepositsBefore.add(TEST_AMOUNT_TWO)
+      )
+      expect(await depositRecord.getUserDepositAmount(uncappedUser.address)).to.eq(
+        accountDepositsBefore.add(TEST_AMOUNT_TWO)
+      )
+    })
+
+    it('correctly add to deposit totals and user totals for an uncapped user if starting from a non-zero value', async () => {
```
ramenforbreakfast, line 145:
```suggestion
    it('correctly adds to global, but not user deposit total if funder uncapped and current deposits > 0', async () => {
```

ramenforbreakfast (approved):
LGTM besides one minor modification and some nits.

ramenforbreakfast:
I've removed my comments because I think there is value in still tallying up an uncapped user's deposits so we can at least keep track of them.

apps/smart-contracts/core/test/unit/DepositRecord.test.ts
XavierEkkel, line 129:
```suggestion
    it('adds to user and global deposit totals if uncapped user and cap exceeded', async () => {
```

apps/smart-contracts/core/test/unit/DepositRecord.test.ts
```diff
@@ -113,6 +126,40 @@ describe('=> DepositRecord', () => {
       )
     })
 
+    it('correctly add to deposit totals and user totals for an uncapped user', async () => {
+      await depositRecord.setUserDepositCap(TEST_AMOUNT_ONE)
```
XavierEkkel, line 130: add expect that TEST_AMOUNT_TWO is gt TEST_AMOUNT_ONE

apps/smart-contracts/core/test/unit/DepositRecord.test.ts
```diff
@@ -113,6 +126,40 @@ describe('=> DepositRecord', () => {
...
+      bypassList.isIncluded.whenCalledWith(uncappedUser.address).returns(true)
+      const globalDepositsBefore = await depositRecord.getGlobalNetDepositAmount()
+      const accountDepositsBefore = await depositRecord.getUserDepositAmount(uncappedUser.address)
```
XavierEkkel, line 133: nit: match the existing variable naming
```suggestion
      const userDepositAmountBefore = await depositRecord.getUserDepositAmount(uncappedUser.address)
```

same for other test

apps/smart-contracts/core/test/unit/DepositRecord.test.ts
```diff
@@ -113,6 +126,40 @@ describe('=> DepositRecord', () => {
       )
     })
 
+    it('correctly add to deposit totals and user totals for an uncapped user', async () => {
+      await depositRecord.setUserDepositCap(TEST_AMOUNT_ONE)
+      bypassList.isIncluded.whenCalledWith(uncappedUser.address).returns(true)
+      const globalDepositsBefore = await depositRecord.getGlobalNetDepositAmount()
```
XavierEkkel, line 132:
```suggestion
      const globalDepositAmountBefore = await depositRecord.getGlobalNetDepositAmount()
```

same for other test

apps/smart-contracts/core/test/unit/DepositRecord.test.ts
```diff
@@ -113,6 +126,40 @@ describe('=> DepositRecord', () => {
...
+        globalDepositsBefore.add(TEST_AMOUNT_TWO)
+      )
+      expect(await depositRecord.getUserDepositAmount(uncappedUser.address)).to.eq(
+        accountDepositsBefore.add(TEST_AMOUNT_TWO)
+      )
+    })
+
+    it('correctly add to deposit totals and user totals for an uncapped user if starting from a non-zero value', async () => {
+      await depositRecord.connect(user).recordDeposit(uncappedUser.address, TEST_AMOUNT_TWO)
+
```
XavierEkkel, line 147:
```suggestion
```

apps/smart-contracts/core/test/unit/DepositRecord.test.ts
XavierEkkel, line 129: also add test for `adds to user and global totals if uncapped user and cap not exceeded`

XavierEkkel, line 129: add test for `reverts if uncapped user and global cap exceeded`

apps/smart-contracts/core/test/unit/DepositRecord.test.ts
XavierEkkel, line 145:
```suggestion
    it('adds to global and user totals if uncapped user below cap and starting from non-zero value', async () => {
```

XavierEkkel (approved):
LGTM after comments

PR #854, opened 2023-02-08
feat: add configuration functions for prod-only elements

ramenforbreakfast:
The only elements not shared by both are `PrePOMarketFactory`, `ArbitrageBroker`, and `DepositTradeHelper`. Since there is no initial configuration needed for the latter two (they have constructor initializations), the only thing we need to do is make `Collateral` valid on the `PrePOMarketFactory`.


XavierEkkel (approved):
LGTM

PR #855, opened 2023-02-08
feat: add configuration script to run at the end of all deploy scripts

ramenforbreakfast:
This script will be called at the end of all deployments automatically. Calls all of the helper functions defined in earlier PRs.

Also, while testing this script, I decided to make a minor change to the deployment helper we have. It originally only logged using the name of the contract, not the deployment name. This made debugging confusing because I could see `AccountList` contracts being deployed, but wasn't sure the context behind each one. So the second log message now uses the `deploymentName` rather than `contractName`.


apps/smart-contracts/core/deploy/ConfigureStackViaSigner.ts
```diff
@@ -0,0 +1,155 @@
...
+    deploymentParameters.withdrawHook.userWithdrawLimitPerPeriod
+  )
+  console.log('Configuring ManagerWithdrawHook via Signer...')
+  await core.configureManagerWithdrawHookViaSigner(
+    signer,
+    deploymentParameters.managerWithdrawHook.minReservePercentage
+  )
+  console.log('Configuring DepositRecord via Signer...')
+  await core.configureDepositRecordViaSigner(
+    // TODO update this to add allowedMsgSenders
```
XavierEkkel, line 123: and also the bypass allowlist

apps/smart-contracts/core/deploy/ConfigureStackViaSigner.ts
```diff
@@ -0,0 +1,155 @@
...
+  'ArbitrageBroker',
+  'Collateral',
+  'DepositHook',
+  'DepositRecord',
+  'DepositTradeHelper',
+  'FixedUintValue',
+  'ManagerWithdrawHook',
+  'PrePOMarketFactory',
+  'TokenSender',
+  'WithdrawHook',
```
XavierEkkel, line 152: do we need the ones that aren't configured above as dependencies?

XavierEkkel (approved):
LGTM after question

apps/smart-contracts/core/deploy/ConfigureStackViaSigner.ts
ramenforbreakfast: That is true, mostly did this out of laziness so I only would have to specify on tag when deploying, but I'll separate it out for consistency.

PR #856, opened 2023-02-08
feat:core: allow market to burn tokens without approval

nmimran99:
feat:core: allow market to burn tokens without approval

Allow the market (token contract owner) to burn token without token owner's approval.
The process checks whether the `msg.sender` is the token contract owner and calls `_burn`, which doesn't require a check for allowance.


apps/smart-contracts/core/test/unit/LongShortToken.test.ts
```diff
@@ -45,4 +48,25 @@ describe('=> LongShortToken', () => {
...
+    })
+
+    it('allows token contract owner to burn user tokens without approval', async () => {
+      await longShort.connect(user).approve(deployer.address, 0)
+      expect(await longShort.connect(deployer).burnFrom(user.address, TEST_AMOUNT))
+        .emit(longShort.address, 'Transfer')
+        .withArgs(user.address, ZERO_ADDRESS, TEST_AMOUNT)
+      expect(await longShort.balanceOf(user.address)).eq(0)
+      expect(await longShort.totalSupply()).eq(0)
+    })
```
ramenforbreakfast, line 64: Remember, we should try to keep tests in an Arrange/Act/Assert format unless its a one liner like a revert case.

Also, we should aim to not perform fragile checks like .eq(0) here because the checks here assume that the beginning supply would just be TEST_AMOUNT.

```suggestion
    it('allows token contract owner to burn user tokens without approval', async () => {
      const userBalanceBefore = await longShort.balanceOf(user.address)
      const totalSupplyBefore = await longShort.totalSupply()
      await longShort.connect(user).approve(deployer.address, 0)
      
      const tx = await longShort.connect(deployer).burnFrom(user.address, TEST_AMOUNT)
      
      expect(tx)
        .emit(longShort.address, 'Transfer')
        .withArgs(user.address, ZERO_ADDRESS, TEST_AMOUNT)
      expect(await longShort.balanceOf(user.address)).eq(userBalanceBefore.sub(TEST_AMOUNT))
      expect(await longShort.totalSupply()).eq(totalSupplyBefore.sub(TEST_AMOUNT))
    })
```

apps/smart-contracts/core/test/unit/LongShortToken.test.ts
```diff
@@ -45,4 +48,25 @@ describe('=> LongShortToken', () => {
...
+
+    it('reverts if burner != token owner and burner != token contract owner', async () => {
```
ramenforbreakfast, line 66:
```suggestion
    it('reverts if insufficient allowance and burner != owner', async () => {
```
This I believe is a more concise and accurate way to write this, since its not that `burner != token owner` would mean a burn reverts, its specifically when a burner doesn't have sufficient allowance.

Also need to add assertions at the beginning that verify these assumptions.

ramenforbreakfast (changes requested):
The contents of this PR are mostly fine, just have some test case modifications.

However, you probably were not aware, but this also impacts `ArbitrageBroker`. The `ArbitrageBroker` in `setMarketValidity` needed to provide approvals to a `PrePOMarket` contract to burn their tokens.

Since this is no longer necessary for `redeem` interactions, we can remove the `LongShortToken` approvals to `market`.

Please add these changes as well to this PR since they are directly related.

apps/smart-contracts/core/test/unit/LongShortToken.test.ts
```diff
@@ -45,4 +48,32 @@ describe('=> LongShortToken', () => {
...
+        .emit(longShort.address, 'Transfer')
+        .withArgs(user.address, ZERO_ADDRESS, TEST_AMOUNT)
+      expect(await longShort.balanceOf(user.address)).eq(userBalanceBefore.sub(TEST_AMOUNT))
+      expect(await longShort.totalSupply()).eq(totalSupplyBefore.sub(TEST_AMOUNT))
+    })
+
+    it('reverts if insufficient allowance and burner != owner', async () => {
+      expect(user2.address).not.eq(await longShort.owner())
+      expect(await longShort.allowance(user2.address, user.address)).lt(TEST_AMOUNT)
+      await expect(longShort.connect(user2).burnFrom(user.address, TEST_AMOUNT)).to.revertedWith(
```
ramenforbreakfast, line 74:
```suggestion
      
      await expect(longShort.connect(user2).burnFrom(user.address, TEST_AMOUNT)).to.revertedWith(
```

ramenforbreakfast (approved):
LGTM, can you also remove any use of the `approveTokensForRedemption` in the `PrePOMarket.test.ts`, since we no longer need to rely on approvals for redemption?

**Please rebase or merge changes from main as the PrePOMarket test has been updated recently.**

XavierEkkel (approved):
LGTM after @ramenforbreakfast's comments

apps/smart-contracts/core/test/unit/PrePOMarket.test.ts
```diff
@@ -515,22 +515,20 @@ describe('=> prePOMarket', () => {
     }
 
     // TODO: need to implement a way to remove the need for approval calls, perhaps using permit signatures?
-    const approveTokensForRedemption = async (
+    const attachTokensForRedemption = async (
```
ramenforbreakfast, line 518: What I meant by removing the token approvals in the `PrePOMarket.test.ts` was that you can remove the entire function. There isn't any point to leaving this function since it just attaches to contracts and returns nothing.

ramenforbreakfast (approved):
LGTM except 1 comment, you can merge afterwards!

PR #858, opened 2023-02-09
fix: position name overflowing the screen

MeLlamoPablo:
This PR fixes the positions list overflowing with long texts on small screens. For some reason I can't figure out, `text-overflow: ellipsis` wasn't working.

My solution is to use multi-line clamp limited to 1 line, which achieves the same effects. `-webkit-line-clamp` is non-standard, but it is supported on all browsers we target (the only browser which doesn't support it is IE). If anyone has a better solution I'm open to other ideas.

The UI doesn't break above 390px wide, which is the lowest we aim to support.

https://user-images.githubusercontent.com/11708035/217795063-707b0a4d-bd04-41ea-8835-1300ee1d60ee.mp4

# FE PR Checklist

- [X] Self-reviewed each line of code, preempted questions about logic or decisions
- [X] Noted any intentional deviations from design
- [X] Tried hard to break it and failed, think like a malicious user trying to find edge cases to break the app
- [X] Tested on supported device sizes
    - [X] iPhone 12 Pro
    - [X] iPad Air
    - [X] 1920x1080 (how to add custom size: [https://umaar.com/dev-tips/51-add-new-device/](https://umaar.com/dev-tips/51-add-new-device/))
- [X] Tested all widths between 1920px wide to 390px wide and confirmed no weird breakages at widths in between
- [X] (if browser specific logic) Tested on Brave, Firefox, Safari


XavierEkkel (approved):
LGTM from UI perspective

chrisling-dev (approved):
LGTM! Great approach

PR #859, opened 2023-02-09
chore: remove unnecessary `font-size` prop

MeLlamoPablo:
`font` overrides `font-size`, so it has no effect and it's confusing.

Screenshot 2023-02-09 at 12 50 01


chrisling-dev (approved):

chrisling-dev:
LGTM

PR #860, opened 2023-02-09
feat: simulator

chrisling-dev:
You can review the UI of this PR by:
- git pull this branch and run yarn dev in 'prepo-ui', then go to `localhost:1234` to check out the storybook
- check the next PR where I use this component (https://github.com/prepo-io/prepo-monorepo/pull/861)

https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=6b4403e1ecc748a48a63e32a1f35e012&p=b336d2aa129a4c6bb7ce9325f681e19e&pm=s


packages/prepo-ui/src/components/Simulator/Simulator.tsx
```diff
@@ -0,0 +1,211 @@
...
+  const [entryX, setEntryX] = useState<number>()
+  const [exitX, setExitX] = useState<number>()
+
+  const [trackWidth, setTrackWidth] = useState<number>()
+
+  // map posiiton/offset to payout range to get token price
+  const mapPositionToPrice = (position: number | undefined): number | undefined => {
+    if (!data || position === undefined) return undefined
+    const { payoutRange } = data
+    return (position / 1) * (payoutRange[1] - payoutRange[0]) + payoutRange[0]
```
MeLlamoPablo, line 73:
```suggestion
    return position * (payoutRange[1] - payoutRange[0]) + payoutRange[0]
```

Is this intended?

packages/prepo-ui/src/components/Simulator/Simulator.tsx
```diff
@@ -0,0 +1,211 @@
...
+  const trackRef = useRef<HTMLDivElement>(null)
+  const { color } = useTheme()
+
+  // value between 0 - 1 to represent position/offset on slider.
+  const [entryX, setEntryX] = useState<number>()
+  const [exitX, setExitX] = useState<number>()
+
+  const [trackWidth, setTrackWidth] = useState<number>()
+
+  // map posiiton/offset to payout range to get token price
```
MeLlamoPablo, line 69:
```suggestion
  // map position/offset to payout range to get token price
```

packages/prepo-ui/src/components/Simulator/Simulator.tsx
```diff
@@ -0,0 +1,211 @@
...
+  onChangeEntry,
+  onChangeExit,
+  shouldSyncPrice,
+}) => {
+  const trackRef = useRef<HTMLDivElement>(null)
+  const { color } = useTheme()
+
+  // value between 0 - 1 to represent position/offset on slider.
+  const [entryX, setEntryX] = useState<number>()
+  const [exitX, setExitX] = useState<number>()
```
MeLlamoPablo, line 65: Having the entry and exit coordinates here seems like an anti-pattern because then we use `onChangeEntry` and `onChangeExit` to propagate the state to the parent, violating the parent to child flow. I think it would be simpler if we had:

```typescript
const Parent = () => {
  const [entryPrice, setEntryPrice] = useState<number>()
  const [exitPrice, setExitPrice] = useState<number>()

  return <Simulator
    data={data}
    entryPrice={entryPrice}
    exitPrice={exitPrice}
    onChangeEntry={setEntryPrice}
    onChangeExit={setExitPrice}
  />
}
```

The simulator simply takes data from the parent and notifies the parent of changes. We would have `mapPriceToPosition` instead of `mapPositionToPrice`.

This way, we don't need `shouldSyncPrice` - instead, the parent can hold two states in case it wants to "decuple" the Simulator price from the user input and it decides when to sync/reset.

I may be misunderstanding something about the component, so let me know if this isn't possible or I overlooked something.

packages/prepo-ui/src/components/Simulator/utils.ts
```diff
@@ -0,0 +1,10 @@
+export const getPositionFromPayoutRange = (
+  payoutRange: [number, number],
+  price: number
+): number => {
+  const [floor, ceiling] = payoutRange
+  const position = (price - floor) / (ceiling - floor)
+  if (position < 0) return 0
+  if (position > 1) return 1
+  return position
```
MeLlamoPablo, line 9:
```suggestion
  clamp(position, 0, 1)
```

From `lodash/clamp`. I think it's easier to read, but feel free to disregard - I can see the argument for not using lodash.

packages/prepo-ui/src/components/Simulator/Simulator.stories.tsx
```diff
@@ -0,0 +1,39 @@
...
+    },
+    exitLongTokenPrice: {
+      name: 'Exit long token price',
+      defaultValue: 0.75,
+    },
+  },
+}
+
+// eslint-disable-next-line react/jsx-props-no-spreading
+const Template: Story<SimulatorProps> = (args) => <Simulator {...args} />
```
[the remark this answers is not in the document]
chrisling-dev: oh good idea, wdyt if we turp off jsx-props-no-spreading completely or do you prefer no spreading? I personally prefer spreading cus then we don't have to dereference objects in many cases

packages/prepo-ui/src/components/Simulator/Simulator.tsx
chrisling-dev: good catch haha, I refactored this formula from a source I found and the 1 was previously a variable

packages/prepo-ui/src/components/Simulator/utils.ts
chrisling-dev: yea I don't like lodash haha it's too heavy, will disregard this, but thanks for the feedback!

packages/prepo-ui/src/components/Simulator/Simulator.tsx
chrisling-dev: yea good point, I was thinking to make this component in a way that even if the parent component doesn't manage entry/exit states, the simulator would still work, but I guess I was optimizing for a use case that isn't valid since we always need the coordinates in parent component to get p/l. I'll update this, thanks for the suggestion!

chrisling-dev: actually it might make more sense to do the logic in simulator component cause otherwise we'll have to handle both handle's x in the parent component everytime we use this component

In the PR I'm working on I will combine onChangeEntry and onChangeExit to just onChange which will give `pnlPerToken`, `maxProfitPerToken`, `maxLossPerToken`, `entryPrice` and `exitPrice` in the callback, making this component easier to use

```
<Simulator
  data={{...data}}
  ...
  onChange={({maxProfitPerToken, maxLossPerToken, pnlPerToken}) => { 
    // can then set pnl states in Parent component's local states to display pnl
  }}
/>
```

chrisling-dev: Don't want to include this logic in this PR because this PR already consist a bunch of complex react states logic, so I'll have the pnl calculation logic in the next PR where I work on the summary

packages/prepo-ui/src/components/Simulator/Simulator.stories.tsx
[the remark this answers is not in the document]
MeLlamoPablo: I prefer no spreading too because it is explicit about which props are being passed exactly and prevents passing unwanted props.

But, sometimes I found the rule too inconvenient so we used to turn it off entirely and then leave the decision of using spreading to the dev's discretion and catch slip ups in code reviews.

packages/prepo-ui/src/components/Simulator/Simulator.tsx
MeLlamoPablo: > I was thinking to make this component in a way that even if the parent component doesn't manage entry/exit states

When I came across this need I usually followed this pattern:

```tsx
export const CounterControlled = ({ onIncrement, value }: {
  onIncrement: () => void;
  value: number;
}) => {
  return <button onClick={onIncrement}>Counter: {value}</button>
}

export const Counter = () => {
  const [value, setValue] = useState(0)
  return <CounterControlled onIncrement={() => setValue(it => it + 1)} value={value} />
}
```

Then the parent can decide which component to use depending on its need. Another approach for more flexibility would be to create an extensible `useCounterLogic` hook that provides the necessary props for the `CounterControlled` to work on its own. Then, parents that don't wanna manage state can just use it, and parents that do can extend the hook or not use it entirely.

MeLlamoPablo (approved):
LGTM!

packages/prepo-ui/src/components/Simulator/Simulator.tsx
chrisling-dev: Thanks for suggestions! I like the hook approach, but this component will only be used twice and I think the current implementation meets both requirements, will consider extracting the logic to a hook if we need more flexibility, I think it's good for now tho

packages/prepo-ui/src/components/Simulator/Simulator.stories.tsx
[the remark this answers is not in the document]
chrisling-dev: ahh gotcha, let's keep the rule for now then, but will disable for stories separately

packages/prepo-ui/src/components/Simulator/Simulator.tsx
MeLlamoPablo: Yea yea, just sharing _an approach_, but we should only use when we need it

chrisling-dev: @MeLlamoPablo so I decided to take reference from your approach (partially) and make this component controllable or not

Approach:
- Simulator will have local states entryX and exitX, interacting with the slider will mutate local states + make a callback to parent component
- If the values passed from parent component are changed, the useEffects will handle syncing to parent component's values which override 
- both entry and exit are synced separately so there's 2 separate useEffect hooks
- any advanced logic would be controlled by parent component

The expected output should work somewhat similar to any interactive component like an `<input />`:
- If we pass a value and changes the value from parent component, the component should reflect that
- If we don't pass a value, the component should still work (well, the difference here is that we'd at least need to pass a default first time value so we cannot "not pass" anything value)

Both of these behaviour will serve both of our use cases well

packages/prepo-ui/src/components/Simulator/Simulator.tsx
```diff
@@ -166,10 +171,16 @@ const Simulator: React.FC<SimulatorProps> = ({
...
     )
       return 'success'
-    const profitableExit = data?.direction === 'short' ? entryX >= exitX : exitX >= entryX
+    const profitableExit = direction === 'short' ? entryX >= exitX : exitX >= entryX
 
     return profitableExit ? 'success' : 'error'
-  }, [data?.direction, entryValuation, entryX, exitValuation, exitX])
+  }, [direction, entryValuation, entryX, exitValuation, exitX])
+
+  const loading = entryX === undefined || exitX === undefined
```
MeLlamoPablo, line 179: I find it weird that this is called `loading` but there's not an asynchronous request associated to it. Maybe `hasData` could be the negation of this?

MeLlamoPablo (approved):

packages/prepo-ui/src/components/Simulator/Simulator.tsx
MeLlamoPablo: Cool! I think it's an improvement over the `shouldSyncPrice` logic. Thanks for iterating on this :)

I still think that separate controlled/uncontrolled components are easier to maintain. In my experience, state synchronization is a very frequent source of bugs, and by separating both versions you don't need to synchronize anything.

Also, [React warns if you use an `<input />` uncontrolledly and then switch to control it](https://stackoverflow.com/q/47012169), which indicates that there's no reason to switch back and forth between both state models. Separating both versions would make it impossible to switch, making the code more robust.

But anyway, feel free to merge and iterate later or just disregard. I don't think it's very productive to spend a lot of time designing the API of this component while we're not using it in many places.

chrisling-dev: yea have to agree with you on being a frequent source of bugs, but also don't want to spend too much time on a component we only use twice, thanks for feedback!

packages/prepo-ui/src/components/Simulator/Simulator.tsx
chrisling-dev: yeap sounds good! thanks!

PR #861, opened 2023-02-09
feat: use profit simulator in trade page

chrisling-dev:
Currently hidden in production and will show when summary part is merged

https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=6b4403e1ecc748a48a63e32a1f35e012&p=f15c9b5a8c7c474dac75e07aacd1884d&pm=s


apps/frontend/core-dapp/src/features/simulator/TradeSimulator.tsx
```diff
@@ -0,0 +1,167 @@
...
+`
+
+const TradeSimulator: React.FC = () => {
+  const { tradeStore, web3Store } = useRootStore()
+  const { action, openTradePrice, selectedPosition } = tradeStore
+  const { connected } = web3Store
+  const [entryPrice, setEntryPrice] = useState<number>()
+  const [exitPrice, setExitPrice] = useState<number>()
+  const [isEntryPriceDirty, setIsEntryPriceDirty] = useState(false)
+  const [isExitPriceDirty, setIsExitPriceDirty] = useState(false)
```
MeLlamoPablo, line 68: I think these two can be computed state, and by getting rid of these `useState` we can simplify some logic.

I recorded a [Loom (5m39s)](https://www.loom.com/share/7c483e7022ce47ee8a140481ec64c281) because I think it's easier to show it this way.

MeLlamoPablo (changes requested):

MeLlamoPablo:
I merged main here but that seemed to bring in some unrelated changes, so I force-pushed to undo the merge and leave the branch on a5fc9ef as it was before.

apps/frontend/core-dapp/src/features/simulator/TradeSimulator.tsx
chrisling-dev: hey thanks for the effort, great suggestion! I did consider this, but didn't go with the computed state approach because:
- the numbers are converted from BigNumber and tend to be pretty huge, so the changed price may not always be same as the original price even on the same offset position (e.g. marketPrice may be 0.566666667 and the changed price from simulator may be 0.566666668 after JS math calculation, I think it also happened once in your video when you dragged the entry handle)

MeLlamoPablo (approved):
Cool, thanks for taking a second look :)

PR #862, opened 2023-02-09
feat: add PortfolioValue component to `/portfolio-next`

MeLlamoPablo:
This PR implements the new PortfolioValue component in the `/portfolio-next` page

Screenshot 2023-02-09 at 14 40 01

### Open questions

With the given design, the text could overflow if > 99M (hopefully someone reaches that! 😆). Do we want to add some logic to simplify amounts? i.e. 99.999.999,99 -> 99,99M

# FE PR Checklist

    - [x] iPhone 12 Pro
    - [x] iPad Air
- [x] (if related to balances) Tested with empty wallet, modest wallet, whale wallet
- [x] (if browser specific logic) Tested on Brave, Firefox, Safari


apps/frontend/core-dapp/src/features/portfolio/PortfolioValue.tsx
```diff
@@ -0,0 +1,98 @@
...
+  const { tokenBalanceFormat } = preCTTokenStore
+
+  return (
+    <Container p={16} flexDirection="column" alignItems="start" justifyContent="start" gap={24}>
+      <Flex flexDirection="column" alignItems="start" gap={4}>
+        <SmallLabel>Portfolio Value</SmallLabel>
+        {portfolioValue === undefined ? (
+          <Skeleton width={120} height={34} />
+        ) : (
+          <BigValue>{toUsd(portfolioValue)}</BigValue>
```
chrisling-dev, line 63: good point on > 99M exceeding the box, can use `compactNumber` util function here to make it show `123.45M`

the `compactNumber` has maximumFractionDigits option set as 2, could you help set minimumFractionDigits to 2 as well so it's always 2 decimals

```suggestion
          <BigValue>${compactNumber(+portfolioValue)}</BigValue>
```

apps/frontend/core-dapp/src/features/portfolio/PortfolioValue.tsx
```diff
@@ -0,0 +1,98 @@
...
+        )}
+      </Flex>
+      <Flex justifyContent="space-between" width="100%">
+        <Flex flexDirection="column" alignItems="start" gap={4}>
+          <Label>Open Positions</Label>
+          {tradingPositionsValue === undefined ? (
+            <Skeleton width={54} height={16} />
+          ) : (
+            <Value>{toUsd(tradingPositionsValue)}</Value>
```
chrisling-dev, line 72: ditto can use compactNumber
```suggestion
            <Value>${compactNumber(tradingPositionsValue)}</Value>
```

apps/frontend/core-dapp/src/features/portfolio/PortfolioValue.tsx
```diff
@@ -0,0 +1,98 @@
...
+          )}
+        </Flex>
+        <Flex flexDirection="column" alignItems="start" gap={4}>
+          <Label>Cash Balance</Label>
+          {tokenBalanceFormat === undefined ? (
+            <Skeleton width={54} height={16} />
+          ) : (
+            <Value>{toUsd(tokenBalanceFormat)}</Value>
```
chrisling-dev, line 80: ditto

```suggestion
            <Value>${compactNumber(tokenBalanceFormat)}</Value>
```

apps/frontend/core-dapp/src/features/portfolio/PortfolioValue.tsx
```diff
@@ -0,0 +1,98 @@
...
+    <Container p={16} flexDirection="column" alignItems="start" justifyContent="start" gap={24}>
+      <Flex flexDirection="column" alignItems="start" gap={4}>
+        <SmallLabel>Portfolio Value</SmallLabel>
+        {portfolioValue === undefined ? (
+          <Skeleton width={120} height={34} />
+        ) : (
+          <BigValue>{toUsd(portfolioValue)}</BigValue>
+        )}
+      </Flex>
+      <Flex justifyContent="space-between" width="100%">
```
chrisling-dev, line 66: Let's use Grid here otherwise it looks weird on table that the cash balance is all the way at the end
```suggestion
      <Grid gridTemplateColumns="repeat(2, 1fr)" width="100%">
```

chrisling-dev (changes requested):
- When wallet isn't connected, let's make all values show `-`
- Let's use primaryLight for button color so it's readable in dark mode

Screenshot 2023-02-10 at 2 30 55 PM

apps/frontend/core-dapp/src/features/portfolio/PortfolioValue.tsx
MeLlamoPablo: Good catch, I also set it to currency = USD since every place where we used it we prepended a dollar sign. Removed that hardcoded dollar sign and renamed the function to `compactUsd`.

I don't like that now it shows $1.1K instead of $1100,00. I think thousands are compact enough to display them fully. If that's fine by you, I'll create a ticket and add more logic and tests in another PR to solve this.

MeLlamoPablo:
> When wallet isn't connected, let's make all values show `-`

Good catch

apps/frontend/core-dapp/src/features/portfolio/PortfolioValue.tsx
chrisling-dev: ohh we should keep $ out of it because there will definitely be use case where we need to show compacted value without dollar sign (e.g. when we migrate sales page later, we will need to show `12.34K PPO staked`

As for $1.1k vs $1100, I think we should let it compact when above 999.99 because some number can be fatter and wont fit on smaller screen (e.g. 2888)

MeLlamoPablo: > ohh we should keep $ out of it because there will definitely be use case where we need to show compacted value without dollar sign (e.g. when we migrate sales page later, we will need to show 12.34K PPO staked

I did this to avoid the `$-10K` issue, maybe we can have `compactNumber(value, { showUsdSign: true })`?

chrisling-dev: good point, yeap `showUsdSign` solution sounds good to me, it should default to false or undefined tho, and only show dollar sign when explicitly specified, since not every use case uses the dollar sign.

apps/frontend/core-dapp/src/utils/number-utils.ts
```diff
@@ -119,9 +121,13 @@ export function formatUsd(amount: number | string | undefined, decimals = true):
   return usd.split('.')[0]
 }
 
-export function compactNumber(value: number): string {
+export function compactUsd(value: number): string {
```
chrisling-dev, line 124: as discussed, prefer to keep compactNumber but either with option `showUsdSign` or take an optional NumberFormatOptions and do this:

```
   function compactNumber(value: number, options: Intl.NumberFormatOptions = {}) {

     return Intl.NumberFormat(undefined, {
       notation: 'compact',
       maximumFractionDigits: 2,
       minimumFractionDigits: 2,
       ...options
      }).format(value)
    }
```

chrisling-dev (changes requested):
LGTM after comments

apps/frontend/core-dapp/src/features/portfolio/PortfolioValue.tsx
```diff
@@ -0,0 +1,117 @@
...
+            <>
+              {tokenBalanceFormat === undefined ? (
+                <>{connected ? <Skeleton width={54} height={16} /> : <Value>-</Value>}</>
+              ) : (
+                <Value>{compactUsd(+tokenBalanceFormat)}</Value>
+              )}
+            </>
+          ) : (
+            <Value>-</Value>
+          )}
```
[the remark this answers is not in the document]
MeLlamoPablo: Good catch, I missed this when refactoring

apps/frontend/core-dapp/src/utils/number-utils.ts
MeLlamoPablo: Gone with the first option. I think it's a good habit to wrap libraries with abstractions that we own (so, not exposing `Intl.NumberFormatOptions`). I guess it's fine since this is a browser API, but this is the approach I'd follow with libraries.

chrisling-dev (approved):
LGTM

PR #865, opened 2023-02-10
feat: add market periphery contract deploy helpers

ramenforbreakfast:
So for some context, market deployment will happen via tasks, rather than a traditional deployment script, since we don't want it to automatically run by default with `hardhat-deploy`, and they are unique in that they are similar to `AccountList`'s where they are meant to be deployed multiple times, but also not actually deployed until later. 

This PR contains code that will, prior to the actual proposal for a new market, deploys the periphery contracts and makes the basic contract alignments. The actual allowlists are configured to allow governance to mint by default.

The market address we can compute ahead of time in `getPredictedMarketAddress` since it is also created using `create2`, this value will be passed in to the periphery configuration functions.

I felt it was appropriate to bundle these for contextual reasons.


XavierEkkel (approved):
LGTM after nits

PR #867, opened 2023-02-10
feat: hide History tab if wallet not connected

MeLlamoPablo:
Screenshot 2023-02-10 at 12 07 10


chrisling-dev (changes requested):
I think the requirements of this should be to hide the "FilterButton" when not connected rather than the entire history tab
https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=6b4403e1ecc748a48a63e32a1f35e012&p=37003864b3d4412c963830afa7c2a638&pm=s

You can do this by changing line 90 - `tabBarExtraContent` to

```
tabBarExtraContent={activeTab === 1 && connected && <FilterButton />}
```

MeLlamoPablo:
Whoops! Fixed it, thanks

chrisling-dev (approved):

PR #869, opened 2023-02-10
feat:core: replace allowed hooks with allowed senders

nmimran99:
feat!: replace allowed hooks with allowed senders

Replace the "allowedHooks" mapping in `depositRecord` with `AllowedMsgSenders` mechanism. 

This is a breaking change to the usage of `depositRecord` because it requires deployment of a separate `AccountList` contract and connecting it to the `AllowedMsgSenders` contract inherited by `DepositRecord`


apps/smart-contracts/core/test/unit/WithdrawHook.test.ts
```diff
@@ -46,14 +48,31 @@ describe('=> WithdrawHook', () => {
...
+      withdrawHook.SET_GLOBAL_WITHDRAW_LIMIT_PER_PERIOD_ROLE(),
+      withdrawHook.SET_USER_WITHDRAW_LIMIT_PER_PERIOD_ROLE(),
+      withdrawHook.SET_TREASURY_ROLE(),
+      withdrawHook.SET_TOKEN_SENDER_ROLE(),
+    ])
+    await batchGrantAndAcceptRoles(mockDepositRecord, deployer, deployer, [
+      mockDepositRecord.SET_ALLOWED_MSG_SENDERS_ROLE(),
+      mockDepositRecord.SET_ACCOUNT_LIST_ROLE()
+
+    ])
```
ramenforbreakfast, line 71: When you rebase off main, be careful you don't overwrite the change to using the new `roleAssigners` helpers.

apps/smart-contracts/core/test/unit/DepositRecord.test.ts
```diff
@@ -149,16 +162,17 @@ describe('=> DepositRecord', () => {
...
       await setupDepositRecord()
+      allowlist.isIncluded.whenCalledWith(user.address).returns(true)
       await depositRecord
         .connect(user)
         .recordDeposit(user.address, TEST_AMOUNT_ONE.add(TEST_AMOUNT_TWO))
     })
 
     it('reverts if caller not allowed', async () => {
-      expect(await depositRecord.isHookAllowed(user2.address)).to.eq(false)
+      expect(await allowlist.isIncluded(user2.address)).to.eq(false)
```
ramenforbreakfast, line 172:
```suggestion
      expect(await allowlist.isIncluded(user2.address)).eq(false)
```

apps/smart-contracts/core/test/unit/DepositRecord.test.ts
```diff
@@ -71,21 +82,23 @@ describe('=> DepositRecord', () => {
         id('setGlobalNetDepositCap')
       )
       expect(await depositRecord.SET_USER_DEPOSIT_CAP_ROLE()).to.eq(id('setUserDepositCap'))
-      expect(await depositRecord.SET_ALLOWED_HOOK_ROLE()).to.eq(id('setAllowedHook'))
+      expect(await depositRecord.SET_ALLOWED_MSG_SENDERS_ROLE()).to.eq(id('setAllowedMsgSenders'))
+      expect(await depositRecord.SET_ACCOUNT_LIST_ROLE()).to.eq(id('setAccountList'))
```
ramenforbreakfast, line 86:
```suggestion
      expect(await depositRecord.SET_ALLOWED_MSG_SENDERS_ROLE()).eq(id('setAllowedMsgSenders'))
      expect(await depositRecord.SET_ACCOUNT_LIST_ROLE()).eq(id('setAccountList'))
```

ramenforbreakfast: Since we're modifying these lines anyway, let's change them to use our current convention which is to not use `.to`

apps/smart-contracts/core/test/unit/DepositRecord.test.ts
```diff
@@ -71,21 +82,23 @@ describe('=> DepositRecord', () => {
...
   describe('# recordDeposit', () => {
     beforeEach(async () => {
       await setupDepositRecord()
+      allowlist.isIncluded.whenCalledWith(user.address).returns(true)
     })
 
-    it('should only be callable by allowed contracts', async () => {
-      expect(await depositRecord.isHookAllowed(user2.address)).to.eq(false)
+    it('reverts if caller not allowed', async () => {
+      expect(await allowlist.isIncluded(user2.address)).to.eq(false)
```
ramenforbreakfast, line 97:
```suggestion
      expect(await allowlist.isIncluded(user2.address)).eq(false)
```

apps/smart-contracts/core/test/unit/DepositRecord.test.ts
ramenforbreakfast: Please do this for the rest of this PR as well.

apps/smart-contracts/core/test/unit/WithdrawHook.test.ts
nmimran99: got it. is that a request for change? or the code that I submitted was ok?

apps/smart-contracts/core/test/unit/DepositRecord.test.ts
nmimran99: Fixed throughout the test file and removed all `to` uses

nmimran99:
Feature is currently on hold for further discussion about the usage of AllowMsgSenders in this contract due to the fact that the bypass feature is using an account list, and the allowed msg senders is also using an account list and those 2 features clash as the `AccountListCaller` contract only allows interaction with 1 account list

apps/smart-contracts/core/test/unit/WithdrawHook.test.ts
ramenforbreakfast: No the code as it is is not ok. The current code overwrites the switch over to using the new `roleAssigners` fixtures for `withdrawHook` and `mockDepositRecord` here. See how your code contains the old way shown above, where each role is passed into `batchGrandAndAcceptRole`? The new code now externalizes this to a `roleAssigners` helper which already knows which roles to assign.

So what you will need to do, is modify the `roleAssigner` for `DepositRecord`, to conform to the changes you've made here, and then use the `roleAssigner` to assign roles, rather than explicitly passing them into `batchGrantAndAcceptRoles`

PR #870, opened 2023-02-13
feat: show chart by default and hide settings

chrisling-dev:
https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=6b4403e1ecc748a48a63e32a1f35e012&p=8b71fbd4d9984794964289e326777633&pm=s


chrisling-dev (commented):
self reviewed

MeLlamoPablo (changes requested):
Let's also hide the close button from the chart. If you click it, there's no way to restore it.

Screenshot 2023-02-13 at 17 25 30

I don't understand why would we offer that option to the user. Why not always show the chart? What does the user have to gain by hiding it?

chrisling-dev:
> Let's also hide the close button from the chart. If you click it, there's no way to restore it.
> 
> Screenshot 2023-02-13 at 17 25 30
> 
> I don't understand why would we offer that option to the user. Why not always show the chart? What does the user have to gain by hiding it?

ohh ya should've removed the close button, thanks!

MeLlamoPablo (approved):

PR #871, opened 2023-02-13
feat: hide expired market from markets list

chrisling-dev:
No UI change expected except there's now a loading skeleton for when market's expiry is loading because we don't have any expired market

https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=6b4403e1ecc748a48a63e32a1f35e012&p=fe754a331d334e7ea2180dee1527c9be&pm=s


MeLlamoPablo (approved):
LGTM

PR #872, opened 2023-02-13
feat: add mock global deposit limits

MeLlamoPablo:
Screenshot 2023-02-13 at 21 40 21

This PR adds a `DepositRecordStore` that can interact with the `DepositRecord` contract.

For now, the data is pulled from a [dummy contract that I deployed on Goerli](https://goerli.etherscan.io/address/0x1412039d105c41d19f8c77aebdc118a5945aba6a#code). This contract only holds state and anyone can modify it. We should replace it with the actual `DepositRecord` contract when it is available.

This PR simply pulls that data and displays it on the frontend. The data is only displayed locally and not on production.

Even though this PR is very simple and doesn't add much value, I'm submitting it to ensure that I'm understanding everything correctly and I'm following the Mobx best practices.


chrisling-dev (approved):
LGTM as a PR for mock implementation

apps/frontend/core-dapp/src/features/deposit/DepositPage.tsx
```diff
@@ -72,6 +88,22 @@ const DepositPage: React.FC = () => {
...
+              : 'Loading...'}
+          </p>
+          <p>
+            Global net deposit cap:{' '}
+            {globalNetDepositCapInUsd
+              ? compactNumber(+globalNetDepositCapInUsd, { showUsdSign: true })
+              : 'Loading...'}
+          </p>
+        </DepositCapInfo>
+      )}
```
chrisling-dev, line 106: tho I think we could remove these changes in this file since they're not necessary to validate implementation of mock values in DepositRecordStore, so we can probably remove all changes in this file, and do proper UI in the next PR

chrisling-dev (changes requested):

chrisling-dev (approved):

PR #873, opened 2023-02-13
feat: helpers for nominating governance to all core roles

ramenforbreakfast:


apps/smart-contracts/core/helpers/roles.ts
```diff
@@ -124,3 +229,14 @@ export const roleAssigners = {
...
   assignArbitrageBrokerRoles,
 }
+
+export const roleGranters = {
+  grantCollateralRoles,
+  grantDepositRecordRoles,
+  grantDepositHookRoles,
+  grantWithdrawHookRoles,
+  grantManagerWithdrawHookRoles,
+  grantTokenSenderRoles,
```
XavierEkkel, line 239: need to add `grantTreasuryCallerRoles` too

XavierEkkel (approved):
LGTM after comment

apps/smart-contracts/core/helpers/roles.ts
ramenforbreakfast: Ah so role granting is not done on a module level, also modules don't assume any sort of access control so im not sure this would make sense to do.

PR #874, opened 2023-02-13
feat: acceptRole batch step generators

ramenforbreakfast:
These are helper functions to make generation of the batch transaction steps cleaner, since we will need to handover ownership of 49 different roles, including the `DEFAULT_ADMIN_ROLE` for each contract.


XavierEkkel (approved):
LGTM

PR #875, opened 2023-02-13
feat: SafeOwnable transfer helpers

ramenforbreakfast:
In addition to the 49 roles, there are 6 contracts that use `SafeOwnable` that will need to have their ownership transferred. There are only 4 now, with a TODO to add the last 2 pertaining to `DepositRecord`.

You will notice that these helpers just take in an array of contracts. While I considered applying the same approach to the more complex role assigners, I prefer writing individually defined role assignment helpers for each contract, rather than doing them all in one function call, because each contract has different roles and the parent calling contracts shouldn't need to know what roles to assign for a given contract.

In this case, the function call is `acceptOwnership` no matter the contract, so it makes sense to allow the helper function to take in an array of contracts.


apps/smart-contracts/core/harnesses/prod.ts
```diff
@@ -135,4 +137,33 @@ export class ProdCore extends Base {
...
   }
+
+  public async transferOwnershipForProdStack(
+    owner: SignerWithAddress,
+    nomineeAddress: string
+  ): Promise<void> {
+    // TODO add DepositRecord allowedMsgSenders and bypass list transfer
+    await sendTxAndWait(
+      await this.collateral.depositHook.allowlist.connect(owner).transferOwnership(nomineeAddress)
+    )
```
XavierEkkel, line 148: what allowlist?

XavierEkkel (approved):
LGTM after question

PR #876, opened 2023-02-13
feat: task for handling the initial ownership handover to governance

ramenforbreakfast:
The handover of roles and ownership to governance will be a 2-step process. This task is the most involved one, handing over all roles and ownership to governance, and then submitting a proposal for governance to accept all 49 role transfers + 4 ownership transfers.

The second-step will come up in a following PR, which will involve the deployer renouncing itself from all roles. This should be done afterwards to ensure the role is in the hands of the multi-sig before the deployer revokes themselves from these roles (or else there could be a situation where there is no default admin assigned, locking the contract from future role assignment).


apps/smart-contracts/core/tasks/roles.ts
```diff
@@ -0,0 +1,125 @@
...
+        abi: hre.artifacts.readArtifactSync('ArbitrageBroker').abi.toString(),
+      },
+      {
+        address: core.depositTradeHelper.address,
+        name: DEPLOYMENT_NAMES.depositTradeHelper.name,
+        network: defenderNetworkName,
+        abi: hre.artifacts.readArtifactSync('DepositTradeHelper').abi.toString(),
+      },
+    ],
+    title: `Accept Core Role and Ownership Transfers to Governance`,
```
XavierEkkel, line 114:
```suggestion
    title: `Accept Core Granted Roles & Ownership Nominations as Governance`,
```

XavierEkkel (approved):
LGTM after comments

PR #877, opened 2023-02-14
feat: helpers for revoking role from an account

ramenforbreakfast:
These are helpers to generate batch transaction steps for when governance revokes roles away from the deployer. This is the 2nd and final part of the access control handoff procedure, and ensures that the revocation happens from our multi-sig as well.


XavierEkkel (approved):
LGTM

PR #878, opened 2023-02-14
feat: helper function to check if an account is the default admin of a contract

ramenforbreakfast:
As an extra precautionary measure, this function checks to make sure that an account (the governance address will be passed in to the script) is the `DEFAULT_ADMIN` of a contract, to ensure we never perform a revocation without the governance already being the root admin.


XavierEkkel (approved):
LGTM

PR #879, opened 2023-02-14
feat: add task for role revocation

ramenforbreakfast:


apps/smart-contracts/core/tasks/roles.ts
```diff
@@ -129,3 +129,99 @@ task(
     steps: acceptRoleSteps.concat(acceptOwnershipSteps),
   })
 })
+
```
XavierEkkel, line 132: nit: need to resolve lint errors e.g. around console.log

XavierEkkel (approved):
LGTM after nit

apps/smart-contracts/core/tasks/roles.ts
ramenforbreakfast: I have rebased this branch and they have been resolved by Niv, the disable message was missing a `*/`

PR #882, opened 2023-02-14
feat:core: add recipient to all hooks and modify tests

nmimran99:
feat:core: add recipient to all hooks and modify tests

In all of them, account list is being checked against the funder

In the RedeemHook contract:
1. `require(_accountList.isIncluded(funder), "Redeemer not allowed");`
1. `_tokenSender.send(recipient, fee);`

In the DepositHook contract:
1. `_satisfiesScoreRequirement` require - checked against recipient
2. `recordDeposit` - recording deposit for recipient
3. `_tokenSender.send(recipient, _fee);`

In the WithdrawHook contract:
1. `_userToAmountWithdrawnThisPeriod[recipient] += amountBeforeFee;`  and other similar user references- changed to `recipient` instead of `funder`
2. `_tokenSender.send(recipient, fee); `- changed to `recipient`

`MintHook` and `ManagerWithdrawalHook` contracts only have names changing, no functionality implications

`IHook` interface was added `funder` and `recipient` to the parameters which drove changes across all hooks.


apps/smart-contracts/core/contracts/DepositHook.sol
```diff
@@ -37,23 +37,24 @@ contract DepositHook is
...
+    address funder,
+    address recipient,
     uint256 amountBeforeFee,
     uint256 amountAfterFee
   ) external override onlyCollateral {
     require(_depositsAllowed, "Deposits not allowed");
-    if (!_accountList.isIncluded(user)) {
-      require(_satisfiesScoreRequirement(user), "Depositor not allowed");
+    if (!_accountList.isIncluded(recipient)) {
+      require(_satisfiesScoreRequirement(recipient), "Depositor not allowed");
```
ramenforbreakfast, line 47: Just for context, I have already discussed with Niv that score requirement checks will be removed in a later PR.

ramenforbreakfast (approved):
LGTM except one thing. @nmimran99 How invasive would it be if we defined another account `recipient` for all tests that we pass in instead of `user.address` for? I am thinking it would only be one additional line per test since we just need to define the account?

Some of these test suites might not be as simple since they most likely verify balance changes and if the recipient is not the same as was previously assumed, the tests would break. But for most, I could see this possibly working with only adding the additional account and using `recipient.address` rather than `user.address`?

I know we already discussed tabling more invasive changes for later due to the complexity of this PR, but we should consider a less-invasive alternative that still does a better job of testing for the moment than having `funder` and `recipient` both assigned to `user.address`.

apps/smart-contracts/core/contracts/DepositHook.sol
nmimran99: that can be done, no problem. the question is if we want to put it into this PR since it will substantially increase the size of it. obviously its only gonna be a simple change but a lot of it in this file.

nmimran99:
> LGTM except one thing. @nmimran99 How invasive would it be if we defined another account `recipient` for all tests that we pass in instead of `user.address` for? I am thinking it would only be one additional line per test since we just need to define the account?
> 
> Some of these test suites might not be as simple since they most likely verify balance changes and if the recipient is not the same as was previously assumed, the tests would break. But for most, I could see this possibly working with only adding the additional account and using `recipient.address` rather than `user.address`?
> 
> I know we already discussed tabling more invasive changes for later due to the complexity of this PR, but we should consider a less-invasive alternative that still does a better job of testing for the moment than having `funder` and `recipient` both assigned to `user.address`.

Adding a recipient would be invasive since adding another account and directing token into it require a change in the tests `expect` clause, which will result in even more changes than we have right now.|
I think this type of change should come with its own PR to not add more load into this already very large PR.

ramenforbreakfast:
> > LGTM except one thing. @nmimran99 How invasive would it be if we defined another account `recipient` for all tests that we pass in instead of `user.address` for? I am thinking it would only be one additional line per test since we just need to define the account?
> > Some of these test suites might not be as simple since they most likely verify balance changes and if the recipient is not the same as was previously assumed, the tests would break. But for most, I could see this possibly working with only adding the additional account and using `recipient.address` rather than `user.address`?
> > I know we already discussed tabling more invasive changes for later due to the complexity of this PR, but we should consider a less-invasive alternative that still does a better job of testing for the moment than having `funder` and `recipient` both assigned to `user.address`.
> 
> Adding a recipient would be invasive since adding another account and directing tokens into it is requiring a change in the tests' `expect` clause, which will result in even more changes than we have right now.| I think this type of change should come with its own PR to not add more load into this already very large PR.

Thats what I figured, thanks for confirming!

XavierEkkel (approved):
LGTM

PR #883, opened 2023-02-15
Feat/simulator summary

chrisling-dev:
https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=6b4403e1ecc748a48a63e32a1f35e012&p=f15c9b5a8c7c474dac75e07aacd1884d&pm=s


apps/frontend/core-dapp/src/features/simulator/SimulatorSummary.tsx
```diff
@@ -0,0 +1,85 @@
...
+    <Wrapper>
+      <SummaryRecord label={`Estimated ${(estimatedPnl ?? 0) >= 0 ? 'Profit' : 'Loss'}`}>
+        {renderPnlUI(estimatedPnl)}
+      </SummaryRecord>
+      <SummaryRecord label="Max Profit">
+        {renderPnlUI(maxProfit, { color: 'success', prefix: '+' })}
+      </SummaryRecord>
+      <SummaryRecord label="Max Loss">
+        {renderPnlUI(maxLoss, { color: 'error', prefix: '-' })}
+      </SummaryRecord>
```
MeLlamoPablo, line 81: We may want to add `loading` if:

* A market is selected BUT
* The data for that market hasn't loaded yet

In this video, you can see a super small window (around second 3) where the valuation from the market hasn't loaded, but the summary switches from `???` -> `$$$` instead of `Skeleton` -> `$$$`



https://user-images.githubusercontent.com/11708035/219000673-ba12cc22-71ff-462b-a5f2-3f173e72688d.mp4

Screenshot 2023-02-15 at 11 22 33

Granted, the chances of this happening are very small, so feel free to disregard.

apps/frontend/core-dapp/src/features/simulator/SimulatorSummary.tsx
```diff
@@ -0,0 +1,85 @@
...
+    const pnlAmount =
+      tradeSize === undefined || tradeSize === 0
+        ? ''
+        : `(${compactNumber(Math.abs(tradeSize * pnlPercentage), { showUsdSign: true })})`
+
+    return (
+      <p>
+        <ColoredLabel color={color ?? (pnlPercentage >= 0 ? 'success' : 'error')}>
+          {prefix ?? (pnlPercentage >= 0 ? '+' : '-')}
+          {(Math.abs(pnlPercentage) * 100).toFixed(2)}%
```
MeLlamoPablo, line 64:
```suggestion
          {Intl.NumberFormat(undefined, {
            maximumFractionDigits: 2,
            minimumFractionDigits: 2,
            signDisplay: 'always',
            style: 'percent',
          }).format(pnlPercentage)}
```

Probably better to delegate this logic to the broswer?

apps/frontend/core-dapp/src/features/simulator/SimulatorSummary.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+  const maxProfit = getPnlPercentage(entryPrice, payoutRange?.[1])
+  const maxLoss = getPnlPercentage(entryPrice, payoutRange?.[0])
+
+  const renderPnlUI = (
+    pnlPercentage?: number,
+    options: {
+      color?: keyof Color
+      prefix?: string
+    } = {}
```
MeLlamoPablo, line 48: I'm not sure why do we need the `prefix` override for if we assign it dynamically depending on whether the number is positive (regardless of whether we use the `Intl` API or not)

I tried to remove this locally and I didn't notice any behavior changes.

If we want to be defensive about not showing negative numbers on "max profit" we can do `Math.max(0, maxProfit)`, but I don't see a scenario where we would want to put `+` on a negative number.

MeLlamoPablo (changes requested):

apps/frontend/core-dapp/src/features/simulator/SimulatorSummary.tsx
chrisling-dev: I use a prefix override solution here to make sure max profit will always have a `+` and max loss will always have a `-`, the Intl.NumberFormat will show + prefix when value is 0 which is not ideal for max loss case, but thanks for pointing out!

chrisling-dev: But feel free to suggest another solution if you think the Intl.NumberFormat could solve the reasoning I stated, agree it's better to delegate to browser if possible

MeLlamoPablo: Then we can do `signDisplay: 'exceptZero'`

MeLlamoPablo: Okay, I get it now. Yeah, I don't think it's possible to ever have `-0` with NumberFormat, so let's keep the current approach. 

Screenshot 2023-02-15 at 12 45 04

MeLlamoPablo (approved):
Approving since the other comment is not a blocker

apps/frontend/core-dapp/src/features/simulator/SimulatorSummary.tsx
chrisling-dev: oh sry somehow my previous comment wasn't submitted, the behaviour is intended because the values in simulator may often become undefined when trade size changes, so we decided to just show ??? for any empty state here and use the loading skeleton in the trade component to indicate loading, otherwise the ux might be annoying if everything keeps changing to skeleton, but thanks for pointing out!

PR #884, opened 2023-02-15
chore: move new portfolio from `/portfolio-next` into `/portfolio`

MeLlamoPablo:
This PR removes the current portfolio page and components and replaces it with the new page, which is currently located on `/portfolio-next`.

This PR may seem overwhelming with a lot of changes, but it is actually pretty simple. I advise to review it commit by commit instead of viewing all file changes at once. Then, you'll find that most big changes are just deletions and file renames.


apps/frontend/core-dapp/src/features/portfolio/Portfolio.tsx
```diff
@@ -1,142 +1,30 @@
-import { Col } from 'antd'
-import { runInAction } from 'mobx'
-import { observer } from 'mobx-react-lite'
-import { useMemo } from 'react'
-import Skeleton from 'react-loading-skeleton'
 import styled from 'styled-components'
-import { media, spacingIncrement, Button, Icon, Flex, Typography, Grid } from 'prepo-ui'
-import { t, Trans } from '@lingui/macro'
-import PortfolioBreakdownItem from './PortfolioBreakdownItem'
```
chrisling-dev, line 9: could you also help delete these files that are no longer used please just so we do keep these unused files in our codebase, thanks!

apps/frontend/core-dapp/src/features/position/Positions.tsx
```diff
@@ -1,14 +1,14 @@
-import { Box, Button, Flex, Typography } from 'prepo-ui'
+import { Button, Flex, Typography } from 'prepo-ui'
 import { observer } from 'mobx-react-lite'
-import { t, Trans } from '@lingui/macro'
-import ClosePositionSummary from './ClosePositionSummary'
```
chrisling-dev, line 4: ditto can delete this file (note: there's 2 ClosePositionSummary file, make sure don't delete the one in `features/trade/close-trade`

chrisling-dev (approved):
LGTM, but could you help clean up the files that are no longer used? Also delete `closeTrade` function in TradeStore (which I added a TODO comment to delete when delete ClosePositionSummary) if you don't mind, just so we don't forget after this PR is merged

PR #885, opened 2023-02-15
feat: implement global deposit cap

MeLlamoPablo:
This PR implements the global deposit cap. To keep things small, the user deposit cap will be implemented in a separate PR.

### Scenario 1: global cap already reached

Screenshot 2023-02-15 at 12 21 03

### Scenario 2: global cap would be exceeded by deposit

https://user-images.githubusercontent.com/11708035/219014649-1af6321e-1d28-4869-943b-2bd13c9c9480.mp4

### How to update the mock contract for testing

Go to the [code on Etherscan](https://goerli.etherscan.io/address/0x1412039d105c41d19f8c77aebdc118a5945aba6a#code) and follow these steps:

https://user-images.githubusercontent.com/11708035/219015015-17993350-d5f4-4b51-9392-f5e4647cedf0.mp4

# FE PR Checklist

    - [x] iPhone 12 Pro
    - [x] iPad Air
- [x] (if related to balances) Tested with empty wallet, modest wallet, whale wallet


apps/frontend/core-dapp/src/features/deposit/DepositWarning.tsx
```diff
@@ -0,0 +1,109 @@
...
+import { Alert as BaseAlert, Icon } from 'prepo-ui/lib/src'
+import styled from 'styled-components'
+import { spacingIncrement } from 'prepo-ui'
+import { useRootStore } from '../../context/RootStoreProvider'
+import { compactNumber, formatUsd } from '../../utils/number-utils'
+
+const Alert = styled(BaseAlert).attrs({
+  color: 'neutral1',
+  icon: (
+    <Icon name="info" color="warning" width={spacingIncrement(20)} height={spacingIncrement(20)} />
```
MeLlamoPablo, line 11: @chrisling-dev intentional deviation from design:

This button here is an ℹ️ while on the design it's an !

I can't find an icon like on the design. Should I submit a separate PR to `prepo-ui`?

apps/frontend/core-dapp/src/features/deposit/DepositWarning.tsx
```diff
@@ -0,0 +1,109 @@
...
+  if (globalDepositCapExceeded === 'already-exceeded' && globalNetDepositCapInUsd !== undefined) {
+    return (
+      <Alert
+        message={
+          <p>
+            Global deposit limit reached (
+            {compactNumber(+globalNetDepositCapInUsd, { showUsdSign: true })})
+          </p>
+        }
+      />
```
MeLlamoPablo, line 75: @chrisling-dev intentional deviation from design:

Screenshot 2023-02-15 at 12 53 22

The tooltip is missing, what should I put?

MeLlamoPablo:
I'm not sure [why the deploy is failing](https://github.com/prepo-io/prepo-monorepo/actions/runs/4183504866/jobs/7248112772), @chrisling-dev could you confirm? I don't have access to Vercel.

apps/frontend/core-dapp/src/features/deposit/DepositWarning.tsx
```diff
@@ -0,0 +1,109 @@
+import { observer } from 'mobx-react-lite'
+import { Alert as BaseAlert, Icon } from 'prepo-ui/lib/src'
+import styled from 'styled-components'
+import { spacingIncrement } from 'prepo-ui'
```
chrisling-dev, line 4: I'm not sure why the build is failing (not recognizing the `prepo-ui/lib/src` path on vercel but works locally, could you try importing from prepo-ui instead of `lib/src`

```suggestion
import styled from 'styled-components'
import { Alert as BaseAlert, Icon, spacingIncrement } from 'prepo-ui'
```

chrisling-dev (changes requested):

apps/frontend/core-dapp/src/features/deposit/DepositWarning.tsx
chrisling-dev: Can include in the same PR the `!` icon

apps/frontend/core-dapp/src/features/deposit/DepositWarning.tsx
MeLlamoPablo: Good catch! That seems to be it. It was an auto-import from the IDE, maybe I need to restrict imports from `/lib/src`

apps/frontend/core-dapp/src/features/deposit/DepositStore.ts
```diff
@@ -119,4 +138,12 @@ export class DepositStore {
...
     return this.root.baseTokenStore.formatUnits(depositAmount)
   }
+
+  get globalRemainingDepositAmountInUsd(): string | undefined {
+    const amount = this.root.depositRecordStore.globalNetDepositAmount
+    const cap = this.root.depositRecordStore.globalNetDepositCap
+    if (!amount || !cap) return undefined
+
+    return this.root.baseTokenStore.formatUnits(cap.sub(amount))
+  }
```
chrisling-dev, line 148: not strongly opinionated here but I think I saw somewhere in mobx's docs (forgot where to find it) that it's better to do

const { globalNetDepositAmount } = this.root.depositRecordStore

rather than copying it to a new variable.

Not strongly opinionated tho, feel free to disregard

chrisling-dev (changes requested):

apps/frontend/core-dapp/src/features/deposit/DepositStore.ts
MeLlamoPablo: Interesting, I don't see how it's different javascript-wise, but I'll research it a bit

apps/frontend/core-dapp/src/features/deposit/DepositWarning.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+  :active {
+    opacity: 0.8;
+  }
+
+  // :focus for a11y, :not(:active) so we don't show on click
+  :focus:not(:active) {
+    border-color: ${({ theme }): string => theme.color.info};
+    border-radius: ${spacingIncrement(4)};
+  }
```
chrisling-dev, line 52: Should've mentioned earlier that the underline in figma is inaccurate, somehow their text rendering engine forces a lot of space between underline and text, it should just be a simple underline, the code below works perfectly fine:

```suggestion
  all: unset;
  color: ${({ theme }): string => theme.color.primaryLight};
  cursor: pointer;
  text-decoration: underline;
  transition: none;

  :hover {
    color: ${({ theme }): string => theme.color.darkPrimaryLight};
  }
```

apps/frontend/core-dapp/src/features/deposit/DepositWarning.tsx
```diff
@@ -0,0 +1,108 @@
...
+  icon: (
+    <Icon name="info" color="warning" width={spacingIncrement(20)} height={spacingIncrement(20)} />
+  ),
+  showIcon: true,
+  type: 'warning',
+})`
+  &&& {
+    .ant-alert-message {
+      font-size: ${({ theme }): string => theme.fontSize.xs};
+      font-weight: ${({ theme }): number => theme.fontWeight.medium};
```
chrisling-dev, line 18: Let's give it 16px line height

```suggestion
    .ant-alert-message {
      line-height: ${spacingIncrement(16)};
      font-size: ${({ theme }): string => theme.fontSize.xs};
      font-weight: ${({ theme }): number => theme.fontWeight.medium};
```

chrisling-dev (changes requested):
Could we try centering the text and icon like in the design?

apps/frontend/core-dapp/src/features/deposit/DepositStore.ts
chrisling-dev: nvm forget about this, I found it and it only matter when we're running in a tracked function, in this case what you're doing is fine

apps/frontend/core-dapp/src/features/deposit/DepositWarning.tsx
MeLlamoPablo: Huh so I thought I needed to publish `prepo-ui` to NPM for that to work, turns out I was just missing a `yarn install` lol

apps/frontend/core-dapp/src/features/deposit/DepositWarning.tsx
MeLlamoPablo: Huh, that makes it a lot simpler. Sometimes Figma does feel inaccurate in that it doesn't look exactly the same when you implement the design.

Also, I kept the `:focus` since we're using `all: unset`, so we need to give it a focus state for accessibility.

chrisling-dev: yea I read about Figma's engineering blogs back then, they have a custom text rendering engine solution which is different than actual browser implementation

apps/frontend/core-dapp/src/features/deposit/DepositStore.ts
```diff
@@ -106,17 +112,41 @@ export class DepositStore {
     return this.root.baseTokenStore.needToAllowFor(this.depositAmount, 'preCT')
   }
 
-  get globalNetDepositAmountInUsd(): string | undefined {
-    const depositAmount = this.root.depositRecordStore.globalNetDepositAmount
-    if (!depositAmount) return undefined
+  get globalDepositCapExceeded(): 'loading' | 'no' | 'already-exceeded' | 'exceeded-if-deposit' {
+    const { globalNetDepositAmount, globalNetDepositCap } = this.root.depositRecordStore
+    const { depositAmountBN } = this
```
chrisling-dev, line 118: should we add a production check here so it doesn't affect the current behaviour?

```suggestion
    // TODO: remove this when contract is updated
    if (isProduction()) return 'no'
```

chrisling-dev (approved):
LGTM after adding `isProduction` check

apps/frontend/core-dapp/src/features/deposit/DepositWarning.tsx
chrisling-dev: I've posted in FE channel to ask for copy of tooltip, for now let's leave it out and add on in a separate PR since it's hidden in production anyway

PR #887, opened 2023-02-15
chore: make import and module inheritance order consistent

ramenforbreakfast:
Our import and module inheritance ordering is not consistent. I went ahead and applied the following ordering rules:
- Imports are ordered based on alphabetical order + location hierarchy. Files immediately local to the contract go first, then our own packages, then external packages (openzeppelin).
- Module inheritance is alphabetical order, but interfaces come first.


XavierEkkel (approved):
LGTM

PR #890, opened 2023-02-16
refactor: use `toUsd` instead of `NumberFormat` directly

MeLlamoPablo:


apps/frontend/core-dapp/src/features/deposit/DepositSummary.tsx
```diff
@@ -18,10 +21,7 @@ const DepositSummary: React.FC = () => {
         {estimatedReceivedAmount === undefined ? (
           <Skeleton height="22px" width="64px" />
         ) : (
-          `$${Intl.NumberFormat(undefined, {
-            maximumFractionDigits: 2,
-            minimumFractionDigits: 2,
-          }).format(estimatedReceivedAmount)}`
+          toUsd(estimatedReceivedAmount)
```
chrisling-dev, line 24: is there any reason for this change? I feel like using NumberFormat gives us more flexibility and is a better solution since the NumberFormat is a built in JS api, the toUsd util is pretty outdated and uses some unconventional way to format numbers, which I have considered removing after every old pages are removed

chrisling-dev (changes requested):

apps/frontend/core-dapp/src/features/deposit/DepositSummary.tsx
MeLlamoPablo: No, I just thought our standard was using `toUsd` and this was an exception. I didn't realize that there was a conscious effort to migrate away from `toUsd`. Maybe we should mark `toUsd` `@deprecated`?

PR #894, opened 2023-02-17
chore:core: remove NFT score requirement code from DepositHook

ramenforbreakfast:
The allowlist removal from `DepositHook` is much more involved, so it will come in the following PR.


XavierEkkel (approved):
LGTM

PR #896, opened 2023-02-17
style: make DepositWarning's InlineTextButton hoverable on dark mode

MeLlamoPablo:
`primaryLight` and `darkPrimaryLight` are the same color on dark mode, which means that the `InlineTextButton` can't be visibly focused or hovered on dark mode.

Previous behavior:

https://user-images.githubusercontent.com/11708035/219605290-def80ab0-8b3f-4a51-bfeb-2001956637db.mp4

New behavior:

https://user-images.githubusercontent.com/11708035/219605335-6762f183-5611-447f-8021-09e7ed15d422.mp4


apps/frontend/core-dapp/src/features/deposit/DepositWarning.tsx
```diff
@@ -42,7 +42,7 @@ const InlineTextButton = styled.button`
 
   :hover,
   :focus {
-    color: ${({ theme }): string => theme.color.darkPrimaryLight};
+    color: ${({ theme }): string => theme.color.info};
```
chrisling-dev, line 45: nice catch but we don't use this pattern anywhere tho, for consistency purpose (e.g. deposit/withdraw link on portfolio page) I think we should keep darkPrimaryLight for now, and perhaps add another color pair for hovering that turns light purple in dark mode to darker purple

chrisling-dev: I don't mind keep it as is for now, and we'll add another ticket later to update all light purple hovering to make sure everything is consistent

chrisling-dev (changes requested):

apps/frontend/core-dapp/src/features/deposit/DepositWarning.tsx
MeLlamoPablo: Hovering is not the problem, it's focusing. We need to give some focus feedback for accessibility. The Deposit/Withdraw links also have this problem.

MeLlamoPablo: > and perhaps add another color pair for hovering that turns light purple in dark mode to darker purple

Also, I'm not sure if making it darker on focus is the way to go, since it may not be accessible for colorblind people. We may want a color that is more catchy. Alternatively, we could add a border to these elements on focus.

chrisling-dev (approved):
good point on accessibility

apps/frontend/core-dapp/src/features/deposit/DepositWarning.tsx
chrisling-dev: not a fan of border, I think this is ok then, LGTM

PR #902, opened 2023-02-18
feat: allow trading w usdc or preCT

chrisling-dev:
https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=6b4403e1ecc748a48a63e32a1f35e012&p=a05afb27388e449ebd9481676ceedb89&pm=s


apps/frontend/core-dapp/src/features/currencies/CurrencyRow.tsx
```diff
@@ -0,0 +1,110 @@
...
+  white-space: nowrap;
+  width: max-content;
+`
+
+const Balance = styled.p`
+  color: ${({ theme }): string => theme.color.neutral3};
+  font-size: ${({ theme }): string => theme.fontSize.sm};
+  font-weight: ${({ theme }): number => theme.fontWeight.medium};
+  line-height: 1.2;
+`
```
MeLlamoPablo, line 58: This is duplicated from `SlideUpRow`. Is this intentional? I think that, as a quick solution, we could export the components from `SlideUpRow` (even if the components are unrelated) so that at least the relationship is established in code. Then, if the components evolve in different ways we could decide to duplicate them.

apps/frontend/core-dapp/src/features/currencies/CurrencyRow.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+const CurrencyRow: React.FC<Props> = ({ hideBalance, onClick, token, selected }) => {
+  const { tokensStore } = useRootStore()
+  let balanceUI: React.ReactNode = <Skeleton width={60} height={16} />
+  const balance = tokensStore.getTokenBalance(token)
+
+  if (balance !== undefined)
+    balanceUI = <Balance>{compactNumber(+balance, { showUsdSign: true })}</Balance>
```
MeLlamoPablo, line 66: I would probably pull this into a child component that observes the `tokensStore`, like `<TokenBalance token={token} />`, considering that we don't use `tokenStore` or `balance` anywhere else in this component.

MeLlamoPablo, line 66: Also, this doesn't seem right for ETH (I have 2.15 ETH, not $2.15 ETH).

Screenshot 2023-02-20 at 10 31 02

We need to convert the ETH price to dollars (I'd argue that we also need to convert the USDC price to dollars in case it goes off-peg)

MeLlamoPablo (changes requested):
Looks great overall! Some small comments

apps/frontend/core-dapp/src/features/currencies/CurrencyRow.tsx
chrisling-dev: yeap intentional, SlideUpRow takes market entity and handle a bunch of stuffs like position direction, valuation and etc, so I duplicated this  rather than trying to make SlideUpRow reusable. It should've been renamed to MarketPositionRow really, but I reckon we could leave the refactor for separate PR/when we clean things up

apps/frontend/core-dapp/src/features/currencies/CurrencyRow.tsx
chrisling-dev: Good point

chrisling-dev: ahh yes you're right, I left this here to test that we can list non-erc20 token here (e.g. native eth) in a way that still keeps the observability of everything, it should've been removed)

MeLlamoPablo: After our sync today, I think we can merge this and deal with units later

MeLlamoPablo (approved):

apps/frontend/core-dapp/src/features/currencies/CurrencyRow.tsx
chrisling-dev: I've removed ETH for now

PR #903, opened 2023-02-18
fix:core: c4 audit m05

nmimran99:
fix: c4 audit m-05 ticket

Added a check to verify that `finalLongPayout` will not be set more than once.


apps/smart-contracts/core/test/unit/PrePOMarket.test.ts
```diff
@@ -226,14 +226,13 @@ describe('=> prePOMarket', () => {
       expect(await prePOMarket.getFinalLongPayout()).to.eq(TEST_CEILING_PAYOUT.sub(1))
     })
 
-    it('should correctly set the same value twice', async () => {
+    it('reverts if setting the value more than once', async () => {
```
XavierEkkel, line 229: nit:
```suggestion
    it('reverts if value set more than once', async () => {
```

XavierEkkel: and move this test below the role revert test to keep the right order of reverts

XavierEkkel (approved):
LGTM after nits

ramenforbreakfast (approved):
LGTM

PR #906, opened 2023-02-20
fix:core: withdraw amount limited by global limit

nmimran99:
fix:core: withdraw amount limited by global limit

This PR fixes the issue where a user can essentially withdraw an amount larger than the global withdraw limit if the withdraw is the first one in the period.


ramenforbreakfast (commented):
@nmimran99 Could you please add a test or modify a test so that it catches this? Clearly our tests were missing an important case so we'd want to add one to verify it given this fix.

ramenforbreakfast (changes requested):
Going to request changes here since we should have additional tests.

ramenforbreakfast:
Spoke about this on our call, but will add here an official reminder.
@nmimran99 needs to add a test case to ensure a user cannot deposit beyond the global limit just because a period has changed (which is what we did not test for and what caused the bug to get past).

apps/smart-contracts/core/test/unit/WithdrawHook.test.ts
```diff
@@ -351,6 +351,22 @@ describe('=> WithdrawHook', () => {
...
+      await withdrawHook
+        .connect(fakeCollateral.wallet)
+        .hook(user.address, user.address, TEST_USER_WITHDRAW_LIMIT, TEST_USER_WITHDRAW_LIMIT)
+      await withdrawHook
+        .connect(deployer)
+        .setGlobalWithdrawLimitPerPeriod(TEST_USER_WITHDRAW_LIMIT.sub(1))
+
+      const tx = withdrawHook
+        .connect(fakeCollateral.wallet)
+        .hook(user.address, user.address, TEST_USER_WITHDRAW_LIMIT, TEST_USER_WITHDRAW_LIMIT)
```
XavierEkkel, line 366:
```suggestion
        .hook(user.address, user.address, 0, 0)
```
should work, since the limit is already exceeded

XavierEkkel (approved):
LGTM after comments and after LGT @ramenforbreakfast

ramenforbreakfast (approved):
LGTM

PR #907, opened 2023-02-20
chore:core: add tests for recipient in hooks

nmimran99:
chore:core: add tests for recipient in hooks
This PR adds tests to hooks that can have the `funder` and `recipient` as 2 different users. 
The tests are focused on the parts where the difference comes in play.


apps/smart-contracts/core/test/unit/DepositHook.test.ts
```diff
@@ -156,6 +156,15 @@ describe('=> DepositHook', () => {
       expect(depositRecord.recordDeposit).calledWith(user.address, TEST_AMOUNT_AFTER_FEE)
     })
 
+    it('calls recordDeposit() if fee > 0 and funder != recipient', async () => {
+      expect(user.address).not.eq(recipient.address)
+      await depositHook
+        .connect(collateral.wallet)
+        .hook(user.address, user.address, TEST_AMOUNT_BEFORE_FEE, TEST_AMOUNT_AFTER_FEE)
```
ramenforbreakfast, line 163: I think this is a mistake, shouldn't the second parameter here be recipient.address? As well as the expect statement?

ramenforbreakfast (changes requested):
So we also should check that the contracts calling hooks pass the correct parameters, I see in the  `Collateral.test.ts` we've already done this (I guess you added it in the initial PR #882?), but if you look in the `PrePOMarket.test.ts`, we are still using the same account when testing hook parameters.

ramenforbreakfast (approved):
LGTM

apps/smart-contracts/core/test/unit/DepositHook.test.ts
```diff
@@ -156,6 +156,15 @@ describe('=> DepositHook', () => {
       expect(depositRecord.recordDeposit).calledWith(user.address, TEST_AMOUNT_AFTER_FEE)
     })
 
+    it('calls recordDeposit() if fee > 0 and funder != recipient', async () => {
+      expect(user.address).not.eq(recipient.address)
```
XavierEkkel, line 160:
```suggestion
      expect(user.address).not.eq(recipient.address)
      
```

apps/smart-contracts/core/test/unit/RedeemHook.test.ts
```diff
@@ -161,6 +158,13 @@ describe('=> RedeemHook', () => {
         expect(tokenSender.send).calledWith(user.address, 1)
       })
 
+      it('calls tokenSender.send() if fee > 0 and funder != recipient', async () => {
+        expect(user.address).not.eq(recipient.address)
```
XavierEkkel, line 162:
```suggestion
        expect(user.address).not.eq(recipient.address)
        
```

apps/smart-contracts/core/test/unit/RedeemHook.test.ts
```diff
@@ -161,6 +158,13 @@ describe('=> RedeemHook', () => {
         expect(tokenSender.send).calledWith(user.address, 1)
       })
```
XavierEkkel, line 160: add to the above test `and redeemer = recipient`

apps/smart-contracts/core/test/unit/RedeemHook.test.ts
```diff
@@ -161,6 +158,13 @@ describe('=> RedeemHook', () => {
...
 
+      it('calls tokenSender.send() if fee > 0 and funder != recipient', async () => {
```
XavierEkkel, line 161:
```suggestion
      it('calls tokenSender.send() if fee > 0 and redeemer != recipient', async () => {
```

apps/smart-contracts/core/test/unit/WithdrawHook.test.ts
```diff
@@ -153,6 +154,18 @@ describe('=> WithdrawHook', () => {
         expect(fakeTokenSender.send).calledWith(user.address, fee)
       })
```
XavierEkkel, line 156: add to the above test `and funder = recipient`

XavierEkkel (approved):
LGTM after nits

PR #912, opened 2023-02-21
feat: coming soon page

chrisling-dev:
https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=6b4403e1ecc748a48a63e32a1f35e012&p=a792370c4f3349ad9c4353cb97fb452d&pm=s


apps/frontend/core-dapp/src/utils/isProduction.ts
```diff
@@ -1 +1,2 @@
-export const isProduction = (): boolean => window.location.host === 'app.prepo.io'
+export const isProduction = (override?: boolean): boolean =>
+  override ?? (typeof window === 'undefined' || window.location.host === 'app.prepo.io')
```
MeLlamoPablo, line 2: What's the logic for having an `override`? I can't see why would we need that. Doesn't this mean that we won't be able to see the app ourselves?

MeLlamoPablo (changes requested):

apps/frontend/core-dapp/src/utils/isProduction.ts
chrisling-dev: the override allows us to force isProduction to return true, this way we can actually preview the changes without merging it, otherwise we can't verify the changes are accurate and ok while we're on PR previews.

Would start "whitelisting" each feature as we slowly release the entire thing

MeLlamoPablo: But if we merge this as it is, we can't develop locally unless we manually change it to `isProduction(false)`, no?

Why not just use `process.env.NODE_ENV === "production"`?

chrisling-dev: ahh should've mentioned that I wanted to get Xavier to review this from UI perspective before merging it, so we need to force isProduction to return true even in preview

After that we can "whitelist" each hidden feature as we release them by removing the "true"

`process.env.NODE_ENV === 'production'` would still return false in preview and not allow us to test Coming Soon page

But good suggestion to use process.env.NODE_ENV === 'production' for checking prod rather than using url

MeLlamoPablo: Ah, that makes sense. Approving then

MeLlamoPablo (approved):
LGTM after removing `isProduction(true)` -> `isProduction()`

apps/frontend/core-dapp/src/utils/isProduction.ts
chrisling-dev: thank you!

PR #914, opened 2023-02-21
chore:core: rename price fields to payout

nmimran99:
chore:core: rename price fields to payout

Rename all fields with `price` to `payout`.
Also updated in `markets.ts`


XavierEkkel (approved):
LGTM

XavierEkkel (approved):

PR #916, opened 2023-02-21
feat:core: enforce token order on market creation

nmimran99:
feat:core: enforce token order on market creation

Enforces token contract address less than the collateral contract address during market creation through 2 `require`s


apps/smart-contracts/core/helpers/index.ts
```diff
@@ -228,3 +235,56 @@ export async function getDeterministicMarketAddress(
   )
   return deterministicAddress
 }
+
+export async function generateLongShortSalts(
```
ramenforbreakfast, line 239: Providing context for Xavier here. Niv needs to define a long short salt generator to create a salt that would cause Collateral to be < Long and Short Token. Currently the generation functions are tied to a harness, which these old testcases don't use so this needs to be defined under `helpers/index.ts`

That being said, @nmimran99 I think there are two things that need to be changed:
1. `harnesses/base.ts` has its own `generateLongShortSalts` that is nearly identical, we should also change it to just call this function and pass in the `generateAddressLessThan` generator.
2. `generateAddressMoreThan` is a generalized function and could be used for multiple purposes beyond markets, this should be moved to `prepo-hardhat/packages/smart-contract-utils.ts` where `generateAddressLessThan` is.

ramenforbreakfast (approved):
LGTM besides change requested

apps/smart-contracts/core/test/unit/PrePOMarketFactory.test.ts
```diff
@@ -219,5 +226,41 @@ describe('=> PrePOMarketFactory', () => {
 
       expect(await prePOMarketFactory.getMarket(marketHash)).to.eq(prePOMarket.address)
     })
+
+    it('reverts if long token address > collateral', async () => {
```
XavierEkkel, line 230: nit: move these revert tests to be before all non-revert tests

packages/prepo-hardhat/src/smart-contracts-utils.ts
```diff
@@ -213,6 +213,35 @@ function generateAddressLessThan(
...
+     * Takes a random number and converts it to a string, using radix
+     * 36, encompassing 0-9 + 26 letters A-Z. Adds 1 to ensure an empty
+     * string is never generated.
+     * Substring is optional, used here to limit the size/complexity of
+     * the string generated for test purposes.
+     */
+    randomString = (Math.random() + 1).toString(36).substring(7)
+    currSalt = formatBytes32String(randomString)
+    currAddress = getCreate2Address(deployer, currSalt, hashedInitCode)
+  } while (BigNumber.from(currAddress).lte(lowerBoundAddressBN))
```
XavierEkkel, line 238: could you also change the function above to use `gte` instead of `gt`?

apps/smart-contracts/core/harnesses/base.ts
```diff
@@ -79,27 +79,20 @@ export abstract class Base {
   public async generateLongShortSalts(
     deployer: string,
     tokenNameSuffix: string,
-    tokenSymbolSuffix: string
+    tokenSymbolSuffix: string,
+    generator: (deployer: string, initCode: BytesLike, lowerBoundAddress: string) => Create2Address
```
XavierEkkel, line 83:
```suggestion
    generateLowerAddress: (deployer: string, initCode: BytesLike, upperBoundAddress: string) => Create2Address
```

packages/prepo-hardhat/src/smart-contracts-utils.ts
```diff
@@ -213,6 +213,35 @@ function generateAddressLessThan(
   }
 }
 
+function generateAddressMoreThan(
```
XavierEkkel, line 216: rename this to `generateHigherAddress` and the other one to `generateLowerAddress`

apps/smart-contracts/core/helpers/index.ts
```diff
@@ -228,3 +235,27 @@ export async function getDeterministicMarketAddress(
...
+  deployer: string,
+  collateral: string,
+  tokenNameSuffix: string,
+  tokenSymbolSuffix: string,
+  generator: (deployer: string, initCode: BytesLike, lowerBoundAddress: string) => Create2Address
```
XavierEkkel, line 244:
```suggestion
  generateAddress: (deployer: string, initCode: BytesLike, lowerOrUpperBoundAddress: string) => Create2Address
```

XavierEkkel (approved):
LGTM after comments

PR #919, opened 2023-02-22
feat: IAgree interface

ramenforbreakfast:


ramenforbreakfast:
Closing for now due to decision to not go through with on-chain agreements

PR #925, opened 2023-02-22
chore:core: add caller and remove user2

nmimran99:
chore:core: add caller and remove user2

Modify tests to use `caller` instead of `user`, and remove `user2` and exchange it for `user`


XavierEkkel (approved):
LGTM

PR #932, opened 2023-02-23
feat:core: Add DepositTradeHelper E2E Skeleton and WETH Auxiliary Changes

ramenforbreakfast:
This PR contains a skeleton test setup for `DepositTradeHelper` E2E testing and auxiliary modifications for WETH.

Since we need to do E2E testing, testing for WETH integration from here on out will be in a new test suite modeled after `ArbitrageTrading.test.ts` and will be E2E testing. There isn't much point to testing WETH wrapping not on a E2E basis since it is a simple process. We should just knock them out together by testing the wrapped version of these functions on a E2E basis, which we need to do prior to launching the helper anyway.

- The mock harness has been modified to accept a custom base token, this is so that we can pass in the real WETH contract for E2E testing, since there are WETH specific functions we need access to.

- Multiple helper functions were also modified to remove dependence on `USDC_DENOMINATOR`. While this should be changed in a more elegant way, for now to save time and reduce change complexity, I am just substituting in a 18 decimal equivalent.

To support an earlier initial contract roll-out, I will focus first on writing + E2E testing `wrapAndDeposit` and `withdrawAndUnwrap` first, and then we can redeploy it after we finish implementing and E2E testing `wrapAndDepositAndTrade`


ramenforbreakfast:
One more note, I think during the linting changes removal, we did not correctly remove `user` from the `ArbitrageTrading.test.ts`, and have added this fix here. I think we should probably add functional testing runs as well in our github actions job to be run whenever `core:` is specified.

apps/smart-contracts/core/test/integration/DepositTradeHelper.test.ts
```diff
@@ -0,0 +1,104 @@
...
+
+describe('=> DepositTradeHelper', () => {
+  let weth: ERC20
+  let core: MockCore
+  let deployer: SignerWithAddress
+  let governance: SignerWithAddress
+  let defaultMarketParams: PrePOMarketParams
+  let create2Deployer: Create2Deployer
+  const TEST_NAME_SUFFIX = 'preSPACEX 2000-10000 31-December 2022'
+  const TEST_SYMBOL_SUFFIX = 'preSPACEX_2000_10000_31_DEC_22'
```
XavierEkkel, line 30: nit: new naming format
```suggestion
  const TEST_NAME_SUFFIX = 'Fake Token ($5-10B)'
  const TEST_SYMBOL_SUFFIX = 'FAKE_5-10B'
```

XavierEkkel (approved):
LGTM after nit

PR #933, opened 2023-02-23
feat:core: add wrapAndDeposit function and tests

ramenforbreakfast:
Instead of initializing a separate IWETH9 variable, I have just wrapped it whenever its used to avoid needing to keep duplicate state.


apps/smart-contracts/core/test/integration/DepositTradeHelper.test.ts
```diff
@@ -114,4 +115,65 @@ describe('=> DepositTradeHelper', () => {
     )
     await snapshotter.saveSnapshot()
   })
+
```
XavierEkkel, line 118: add constructor tests (check base token is set based on collateral's base token, check approvals given)

apps/smart-contracts/core/test/integration/DepositTradeHelper.test.ts
```diff
@@ -114,4 +115,65 @@ describe('=> DepositTradeHelper', () => {
...
+      /**
+       * This uses rejecteWith rather than reverted because the transaction
+       * in this case will not even be evaluated and rejected by the provider.
+       */
+      await expect(
+        depositTradeHelper.connect(user).wrapAndDeposit(user.address, { value: ethToDeposit })
+      ).rejectedWith("sender doesn't have enough funds to send tx")
+    })
+
+    it('wraps and deposits ETH if funder == recipient', async () => {
```
XavierEkkel, line 137:
```suggestion
    it('wraps and deposits ETH if funder = recipient', async () => {
```

apps/smart-contracts/core/test/integration/DepositTradeHelper.test.ts
```diff
@@ -114,4 +115,65 @@ describe('=> DepositTradeHelper', () => {
     )
     await snapshotter.saveSnapshot()
   })
+
+  describe('# wrapAndDeposit', () => {
+    it('reverts if funder has insufficient ETH', async () => {
```
XavierEkkel, line 120: add test for 'reverts if base token deposit provides insufficient WETH' (e.g. if WETH were upgraded incorrectly or a non-WETH contract with the same interface were used)
this would use a fake of an ERC20

apps/smart-contracts/core/test/integration/DepositTradeHelper.test.ts
```diff
@@ -114,4 +115,65 @@ describe('=> DepositTradeHelper', () => {
...
+      /**
+       * Could just set to user's eth balance, because the gas cost
+       * would be deducted from the user's balance, but this is less
+       * fragile in the case the test net gas cost is set to 0.
+       */
+      const ethToDeposit = (await user.getBalance()).add(1)
+
+      /**
+       * This uses rejecteWith rather than reverted because the transaction
```
XavierEkkel, line 129:
```suggestion
       * This uses rejectedWith rather than reverted because the transaction
```

apps/smart-contracts/core/test/integration/DepositTradeHelper.test.ts
```diff
@@ -114,4 +115,65 @@ describe('=> DepositTradeHelper', () => {
...
+       * in this case will not even be evaluated and rejected by the provider.
+       */
+      await expect(
+        depositTradeHelper.connect(user).wrapAndDeposit(user.address, { value: ethToDeposit })
+      ).rejectedWith("sender doesn't have enough funds to send tx")
+    })
+
+    it('wraps and deposits ETH if funder == recipient', async () => {
+      const ethToDeposit = parseEther('1')
+      const ethBalanceBefore = await user.getBalance()
```
XavierEkkel, line 139: add expect that balance is significantly greater than ethToDeposit

apps/smart-contracts/core/test/integration/DepositTradeHelper.test.ts
```diff
@@ -114,4 +115,65 @@ describe('=> DepositTradeHelper', () => {
...
+        collateralBalanceBefore.add(await getCollateralAmount(core.collateral, ethToDeposit))
+      )
+      const receipt = await tx.wait()
+      expect(await user.getBalance()).eq(
+        ethBalanceBefore.sub(ethToDeposit).sub(receipt.gasUsed.mul(receipt.effectiveGasPrice))
+      )
+    })
+
+    it('wraps and deposits ETH if funder != recipient', async () => {
+      const ethToDeposit = parseEther('1')
```
XavierEkkel, line 159: can make this a const outside of these specific tests to avoid duplication

XavierEkkel (approved):
LGTM after comments

apps/smart-contracts/core/test/integration/DepositTradeHelper.test.ts
ramenforbreakfast: I will do this for `ethToDeposit`, but `ethBalanceBefore` would require adding an asynchronous beforeeach block since it is an asynchronous fetch.

PR #935, opened 2023-02-23
chore:core: add lint and prettier to shared and token

nmimran99:
chore:core: add lint and prettier to shared and token

This feature is adding scripts to run linting and prettier. 
To call them all we run `yarn l:all` command.
Other commands added:
```
"lint:prepo-shared-contracts": "cd '../../../packages/prepo-shared-contracts' && yarn c && yarn lint",
"lint:prepo-hardhat": "cd '../../../packages/prepo-hardhat' && yarn lint",
"lint:token": "cd ../token && yarn c && yarn lint",
"lint:all": "yarn lint && yarn lint:token && yarn lint:prepo-shared-contracts && yarn lint:prepo-hardhat", 
"prettier": "prettier --write './deploy/*.ts' './hardhat.config.ts' './harnesses/*.ts' './helpers/*.ts' './test/**/*.ts' './utils/*.ts'",
"prettier:prepo-shared-contracts": "cd '../../../packages/prepo-shared-contracts' && yarn l",
"prettier:prepo-hardhat": "cd '../../../packages/prepo-hardhat' && yarn l",
"prettier:token": "cd ../token && yarn l",
"prettier:all": "yarn prettier && yarn prettier:token && yarn prettier:prepo-shared-contracts && yarn prettier:prepo-hardhat", 
```

The choice to separate the calling for each repo to a separate script is to maintain order and ability to call them separately. 
The choice to add `yarn l:all` is to keep the behavior of `yarn l` uniform across the repos.


nmimran99:
Current implementation is using a new zyarn l:all` script. if that approach passes we should change the workflow to use `yarn l:all` instead of `yarn l`

ramenforbreakfast (commented):
I'm not sure if this is the way to go? This doesn't seem like a nice scalable solution given that this "all" command only exists inside the `core` repo. Is there a way we can do this where it sits inside the root of the `prepo-monorepo`?

Given that happens, I also don't think we should add commands referencing specific projects inside this file. I think the `package.json` of any specific proejct should just contain commands pertaining to that project.

The root can contain the command the run prettier/lint on all projects.

PR #936, opened 2023-02-23
feat:core: add snapshotter to long tests

nmimran99:
feat:core: add snapshotter to long tests

Adds snapshotter functionality to long tests that require a usage of a snapshot


apps/smart-contracts/core/test/unit/DepositRecord.test.ts
```diff
@@ -237,18 +239,22 @@ describe('=> DepositRecord', () => {
       )
     })
 
-    afterEach(() => {
+    after(() => {
```
ramenforbreakfast, line 242:
```suggestion
    afterEach(() => {
```
Good catch!, this still needs to happen after each block however.

apps/smart-contracts/core/test/unit/DepositRecord.test.ts
```diff
@@ -363,12 +369,15 @@ describe('=> DepositRecord', () => {
         userDepositBeforeSecondWithdrawal
       )
     })
+
+    after(() => {
```
ramenforbreakfast, line 373:
```suggestion
    afterEach(() => {
```
Please do this for all other fake resets as well!

apps/smart-contracts/core/test/unit/WithdrawHook.test.ts
```diff
@@ -638,4 +649,12 @@ describe('=> WithdrawHook', () => {
...
     })
   })
+
+  after(() => {
+    fakeCollateral.transfer.reset()
+    fakeCollateral.transferFrom.reset()
+    allowList.isIncluded.reset()
+    mockDepositRecord.recordWithdrawal.reset()
+    fakeTokenSender.send.reset()
+  })
```
ramenforbreakfast, line 659:
```suggestion
```
No need to reset all of these afterwards, these variables will not be relevant once exiting this test suite

ramenforbreakfast (approved):
LGTM besides comments!

apps/smart-contracts/core/test/unit/DepositRecord.test.ts
nmimran99: It fails with afterEach

apps/smart-contracts/core/test/unit/DepositRecord.test.ts
nmimran99: Fails with afterEach

apps/smart-contracts/core/test/unit/WithdrawHook.test.ts
```diff
@@ -395,6 +401,11 @@ describe('=> WithdrawHook', () => {
 
       await expect(tx).revertedWith('Global withdraw limit exceeded')
     })
+
+    afterEach(() => {
+      fakeTokenSender.send.reset()
+      mockTestToken.transferFrom.reset()
```
XavierEkkel, line 407: these aren't set anywhere except in `before()`, so I don't think this is necessary

could you also remove the unnecessary `mockTestToken.transferFrom.reset()` on line 139?

apps/smart-contracts/core/test/unit/DepositRecord.test.ts
```diff
@@ -238,17 +243,24 @@ describe('=> DepositRecord', () => {
...
       await setupDepositRecord()
       allowlist.isIncluded.whenCalledWith(caller.address).returns(true)
       await depositRecord
         .connect(caller)
         .recordDeposit(user.address, TEST_AMOUNT_ONE.add(TEST_AMOUNT_TWO))
+      await snapshotter.saveSnapshot()
+    })
+
+    beforeEach(() => {
+      allowlist.isIncluded.whenCalledWith(caller.address).returns(true)
```
XavierEkkel, line 263: this is already in `before`, so this can be removed

apps/smart-contracts/core/test/unit/DepositRecord.test.ts
```diff
@@ -363,12 +375,15 @@ describe('=> DepositRecord', () => {
         userDepositBeforeSecondWithdrawal
       )
     })
+
+    afterEach(() => {
+      allowlist.isIncluded.reset()
```
XavierEkkel, line 380: this can be removed

apps/smart-contracts/core/test/unit/DepositRecord.test.ts
```diff
@@ -68,8 +67,14 @@ describe('=> DepositRecord', () => {
...
-    beforeEach(async () => {
+    snapshotter.setupSnapshotContext('DepositRecord-recordDeposit')
+
+    before(async () => {
       await setupDepositRecord()
+      await snapshotter.saveSnapshot()
+    })
+
+    beforeEach(() => {
       allowlist.isIncluded.whenCalledWith(caller.address).returns(true)
```
XavierEkkel, line 78: this can be moved to `before`

apps/smart-contracts/core/test/unit/DepositRecord.test.ts
```diff
@@ -238,17 +243,24 @@ describe('=> DepositRecord', () => {
     })
 
     afterEach(() => {
```
XavierEkkel, line 245: this can be removed

XavierEkkel (approved):
LGTM after comments

apps/smart-contracts/core/test/unit/WithdrawHook.test.ts
nmimran99: Removed from line 139, but both resets are needed for the tests to pass

apps/smart-contracts/core/test/unit/DepositRecord.test.ts
nmimran99: that is also needed, it was a change I made after a call with @ramenforbreakfast and he confirmed it needs to be there.

apps/smart-contracts/core/test/unit/DepositRecord.test.ts
nmimran99: Same as 2 comments above, it is needed for the tests to pass and is located there after I worked with @ramenforbreakfast on this

apps/smart-contracts/core/test/unit/WithdrawHook.test.ts
ramenforbreakfast: The `mockTestToken.transferFrom.reset()` is needed because we verify it is `calledWith` certain parameters. I think the confusion here is that `reset()` is meant to reset values that it is being set with, which it does, but thats not the point here.

The point with resetting is resetting the history. If you don't reset after a test, an expect for a call to a fake contract could result in a false positive because it retains history from previous calls. It is important to remember that fake and smock contract histories aren't correlated with the state of the EVM and are not reset when the EVM is reset.

apps/smart-contracts/core/test/unit/DepositRecord.test.ts
ramenforbreakfast: Yes this is because of the `reset()` will reset what is returned as well. So we need to set the value that is being set every time because we are resetting the fake or smock to ensure its call history does not result in false positives.

This is really not a concern for efficiency because these are inconsequential synchronous calls.

PR #937, opened 2023-02-23
chore:core: c4 audit g15 optimization

nmimran99:
chore: c4 audit g15 optimization

Very small gas optimization, but saves 97 gas.


XavierEkkel (approved):
LGTM

PR #939, opened 2023-02-24
feat:core: add withdraw amount calculator

ramenforbreakfast:
So we need to add a calculator for withdrawal amounts, currently our helper methods are designed around calculating deposit amounts. I also went ahead and changed the names to reflect the difference between the helper functions.

I also identified a error with the name of one of the variables/functions, specifically `getBaseTokenAmountAfterFee` actually gets the `getBaseTokenAmountBeforeFee`. The logic and the way it is used is also as if it was the `amountBeforeFee`, so this was clearly a mistake and I have fixed the naming throughout.


XavierEkkel (approved):
LGTM

PR #942, opened 2023-02-24
chore:core: adding general linting functionality and solving linting errors

nmimran99:
chore:core: adding general linting functionality and solving linting errors

This PR adds a `scripts/general-lint.sh` file on the main repo. this file is being called by `yarn l` and is performing `yarn l` and `yarn lint` on:
1. core repo
2. token repo
3. prepo-shared-contracts 
4. prepo-hardhat


packages/prepo-shared-contracts/deploy/FixedUintValue.ts
```diff
@@ -1,4 +1,5 @@
 /* eslint-disable no-console */
+/* eslint-disable prettier/prettier */
```
XavierEkkel, line 2: why?

scripts/global-lint.sh
```diff
@@ -0,0 +1,35 @@
...
+run $SHARED_CONTRACTS "c"
+run $SHARED_CONTRACTS "l"
+run $SHARED_CONTRACTS "lint"
+
+# prepo-hardhat
+run $HARDHAT "l"
+run $HARDHAT "lint"
+
+
+
```
XavierEkkel, line 35:
```suggestion
```

scripts/global-lint.sh
```diff
@@ -0,0 +1,35 @@
...
+    if [ $status -eq 1 ]
+    then
+        exit 1
+    fi
+}
+
+# core
+run $CORE "c"
+run $CORE "l"
+run $CORE "lint"
```
XavierEkkel, line 18: lint is already within 'l'

package.json
```diff
@@ -23,6 +23,7 @@
         "dev": "turbo run compile && turbo run dev --parallel",
         "compile": "turbo run compile",
         "lint": "turbo run lint",
+        "l": "./scripts/global-lint.sh",
```
XavierEkkel, line 26:
```suggestion
        "l": "./scripts/global-lint.sh",
        "lint": "turbo run lint",
```

packages/prepo-hardhat/package.json
```diff
@@ -6,7 +6,9 @@
   "files": ["src/index.ts", "src/types.ts"],
   "scripts": {
     "build": "tsc",
-    "build:watch": "yarn build --watch"
+    "build:watch": "yarn build --watch",
+    "l": "yarn prettier --write ./src/*.ts",
```
XavierEkkel, line 10: include `yarn lint` within `l`

scripts/global-lint.sh
XavierEkkel: same for others (or it should be)

scripts/global-lint.sh
```diff
@@ -0,0 +1,35 @@
...
+    yarn --cwd $1 $2
+    status=$?
+    if [ $status -eq 1 ]
+    then
+        exit 1
+    fi
+}
+
+# core
+run $CORE "c"
```
XavierEkkel, line 16: compile seems unnecessary - can just run `yarn sl` (solidity linting)?
same for others below

packages/prepo-shared-contracts/deploy/FixedUintValue.ts
nmimran99: The prettier rules for eslint seem to overlap with the configuration of prettier. I disabled it through eslint so the warnings will cease to throw warnings on the 2 console.logs

scripts/global-lint.sh
nmimran99: if I dont run compile, eslint fails. also `yarn sl` is only running prettier for contracts

packages/prepo-shared-contracts/deploy/FixedUintValue.ts
nmimran99: I cant disable the specific lines if you think that is better

scripts/global-lint.sh
ramenforbreakfast: Ahhh right, because our files reference generated types, and so it will fail due to the missing import reference.

packages/prepo-shared-contracts/deploy/FixedUintValue.ts
nmimran99: THe idea is that we currently have 2 prettier engines work at the same time, the eslint one and the actual prettier engine. I looked over at the documentation and it states that there might be clashes if we're using a stand alone prettier engine and the eslient styling engine, which is build off of prettier too. 
As I see it and after a call with @ramenforbreakfast , there are 2 viable options:
1. Remove the prettier functionality from eslint all together in all the folder, since we are running prettier anyway, there is no reason to run the rule engine on it too.
2. Keep the same configuration as now, which is much less invasive. the ony clash we encountered until now is this file, so I dont think we necessarily have to change so many configurations just because we have a clash in 1 file. I think that for now, and to reduce complication, we should leave that disable in that file, and address this in the future if we encounter another iteration of the same problem.

nmimran99: reference:
https://prettier.io/docs/en/integrating-with-linters.html

ramenforbreakfast: Lets stick with option 2 for now, I'll create a ticket for 1!

ramenforbreakfast (approved):
LGTM

XavierEkkel (approved):
LGTM!

PR #943, opened 2023-02-24
refactor: remove `GasStore`

MeLlamoPablo:
This PR removes `GasStore` from `prepo-stores`.

I realized that there was something off with our gas estimation method, because Goerli has been congested recently and I had some transactions stuck. The `GasStore` wasn't actually doing anything useful for two reasons:

* We don't have an `AdvancedSettingsModal` anymore and the transaction speed (fast/normal/slow) can't be configured in-app
* The mechanism for estimating gas prices (`gasPriceCheckers`) was only available for Polygon Mainnet, and we were using a hardcoded gas price on Goerli, and the rest of the networks.

I don't think we want to suggest a gas price at all. We should let wallets do it for the user. I think every wallet should do it, but I can confirm that the changes in this PR work with:

* Metamask
* Gnosis Safe over WalletConnect
* Coinbase Wallet (mobile)

Note: this PR only removes gas price estimation, not gas limit estimation.


MeLlamoPablo:
> I totally agree with you, we made the GasStore because we were previously on polygon and always had to use a gas price higher than everyone else by 1.5 to get a tx through. Thanks for cleaning this up

Since polygon implemented EIP-1559 maybe this wouldn't be necessary on polygon anymore? Not that it matters, I'm just curious

PR #946, opened 2023-02-24
feat:core: add collateralization logic

ramenforbreakfast:
So ideally, a more complex implementation would actually subtract the pre-collateralization-adjusted amount from the `DepositRecord`, this is somewhat complicated. The current implementation approves the fee and passes `amountBeforeFee` and `amountAfterFee` to the hook, we want this fee (and rebate) to be based on the adjusted amount, so we'd also need to pass in a pre-collateralization-adjusted amount for the `DepositRecord` use.

I have opted for a much simpler implementation that simply passes the adjusted amounts to the `DepositHook` and by proxy the `DepositRecord`. This means that `globalNetDeposits` will not accurately reflect its marked down value, and in the event of a re-collateralization, `globalNetDeposits` will never be zeroed out.

This won't actually matter, because we will redeploy a `Collateral` contract as well as records in the event of a re-collateralization. Hence I went with the much simpler implementation.


apps/smart-contracts/core/test/unit/Collateral.test.ts
```diff
@@ -1195,6 +1231,25 @@ describe('=> Collateral', () => {
       expect(await baseToken.balanceOf(user.address)).to.eq(userBTBefore.add(expectedBT.sub(fee)))
     })
```
XavierEkkel, line 1233: add `reverts if collateralization factor reduces base token amount to 0` and `reverts if collateralization factor reduces fee to 0` tests

XavierEkkel (approved):
LGTM after comment

PR #947, opened 2023-02-25
feat:core: remove withdrawalsAllowed from WithdrawHook

ramenforbreakfast:
Also went ahead and removed user period related parameters that were still in our config scripts.


XavierEkkel (approved):
LGTM

PR #948, opened 2023-02-26
feat:core: add max period length

ramenforbreakfast:
This adds a maximum global period length and a minimum global withdraw limit per period.

Since it a flexible global period length was desired, the true minimum withdraw limit should be evaluated based on the max period. These amounts were chosen arbitrarily and appreciate input for desired amounts.

The `_minGlobalWithdrawLimitPerPeriod` was a little more complicated because the global withdraw limit is denominated in base tokens, which might not be necessarily 18 decimal places and cannot be set in stone. I have currently set the limit to be 1000 units of the base token.

Since we now have a minimum global withdraw limit, I also had to modify the `WithdrawHook.test.ts` to accomodate this. The change I made was meant to be as non-invasive as possible.

Many other places were modified to simply account for the fact that the `WithdrawHook` requires constructor parameters, since the global withdraw limit must be initialized to the minimum to be an effective floor.


apps/smart-contracts/core/contracts/WithdrawHook.sol
```diff
@@ -21,6 +21,9 @@ contract WithdrawHook is
   uint256 private _lastGlobalPeriodReset;
   uint256 private _globalAmountWithdrawnThisPeriod;
 
+  uint256 private constant _maxGlobalPeriodLength = 7 days;
```
XavierEkkel, line 24:
```suggestion
  uint256 public constant override MAX_GLOBAL_PERIOD_LENGTH = 7 days;
```
and remove getter

apps/smart-contracts/core/contracts/WithdrawHook.sol
```diff
@@ -21,6 +21,9 @@ contract WithdrawHook is
...
+  uint256 private immutable _minGlobalWithdrawLimitPerPeriod;
```
XavierEkkel, line 25:
```suggestion
```
just add the max period length in this PR, since there is more to discuss around min amounts per period

apps/smart-contracts/core/contracts/WithdrawHook.sol
```diff
@@ -31,6 +34,12 @@ contract WithdrawHook is
   bytes32 public constant SET_TREASURY_ROLE = keccak256("setTreasury");
   bytes32 public constant SET_TOKEN_SENDER_ROLE = keccak256("setTokenSender");
 
+  constructor(ICollateral collateral, uint256 baseTokenDecimals) {
+    _collateral = collateral;
+    _minGlobalWithdrawLimitPerPeriod = 10**baseTokenDecimals * 1000;
+    _globalWithdrawLimitPerPeriod = 10**baseTokenDecimals * 1000;
+  }
+
```
XavierEkkel, line 42:
```suggestion
```
unclear why collateral is suddenly being passed in

packages/prepo-constants/src/core.ts
```diff
@@ -5,3 +5,5 @@ export const FEE_DENOMINATOR = 1000000
 export const MARKET_FEE_LIMIT = 100000
 export const MAX_PAYOUT = parseEther('1')
 export const PERCENT_DENOMINATOR = 1000000
+export const MIN_GLOBAL_WITHDRAW_LIMIT_PER_PERIOD_IN_UNITS = 1000
```
XavierEkkel, line 8:
```suggestion
```

apps/smart-contracts/core/test/unit/WithdrawHook.test.ts
```diff
@@ -453,6 +471,12 @@ describe('=> WithdrawHook', () => {
       )
     })
 
+    it('reverts if > max period length', async () => {
```
XavierEkkel, line 474:
```suggestion
    it('reverts if set to > max period length', async () => {
```

XavierEkkel (approved):
LGTM after comments

apps/smart-contracts/core/contracts/WithdrawHook.sol
ramenforbreakfast: Ah yeah I think I initiall was sourcing the base token decimals from `Collateral`, but then switched over to just passing it in.

PR #949, opened 2023-02-27
feat: update collateral contract abi

chrisling-dev:
This PR updates core dapp to use arbitrum, the contracts addresses and fixes the things that got broken along with it


apps/frontend/core-dapp/src/features/deposit/DepositStore.ts
```diff
@@ -113,57 +118,62 @@ export class DepositStore {
...
+      depositFee === undefined ||
       this.depositAmountBN === undefined ||
-      feeDenominator === undefined
+      percentDenominator === undefined
     )
       return undefined
 
     return (
       this.depositAmountBN
         // TODO un-hardcode this when we have the fee enabled in our test contracts
```
MeLlamoPablo, line 137:
```suggestion
```

apps/frontend/core-dapp/src/lib/core-contracts.ts
```diff
@@ -18,72 +13,45 @@ export type CoreContracts = {
...
+  goerli: '0x92680aa94e19176ac5a2dd20a43573521d6b8949', // Dummy contract: The values can be modified by anyone.
+  arbitrumOne: '0xEe9e151C51758531FA2A9C550eFEB4c77845988d',
 }
 
 export const TOKEN_SENDER_ADDRESS: ExternalContract = {
-  // TODO: update to actual contract address
-  // A dummy contract for testing purposes only.
-  // The values can be modified by anyone.
-  goerli: '0xE8cCbbBe6dcC8b88484c8bC7e3bCfa1C431092d7',
+  goerli: '0xE8cCbbBe6dcC8b88484c8bC7e3bCfa1C431092d7', // Dummy contract: The values can be modified by anyone.
```
MeLlamoPablo, line 43: Maybe just remove these if we're not gonna support Goerli anymore? The addresses are stored in git anyway, in case we wanna restore them.

apps/frontend/core-dapp/jest.config.ts
```diff
@@ -12,6 +12,10 @@ process.env.TZ = 'UTC'
 const config: Config.InitialOptions = {
   ...base,
   roots: [fromRoot('.')],
+  // TODO: remove this when we have market contracts to test on arbitrum
+  // ignoring tradeStore now because a bunch of issues arises with trying to test trading w markets
+  // which doesnt exist on arbitrum yet
+  testPathIgnorePatterns: ['TradeStore.test.ts'],
```
MeLlamoPablo, line 18: What are the issues? Can we do `describe.skip` instead? This way the fact that we're ignoring a test is more discoverable, because Jest will report the number of skipped tests in the results.

MeLlamoPablo (approved):
LGTM after nits. Good job!

apps/frontend/core-dapp/jest.config.ts
chrisling-dev: So TradeStore requires a selectedMarket (which we just initialize a new MarketEntity) for most of the tests to test trading with markets. The tests are valid, they just don't work yet because we don't have any working Market contract on arbitrum, so I don't want to mock them to something that causes the tests to be meaningless when we add markets contracts later. but yea good suggestion to use describe.skip

PR #953, opened 2023-02-28
fix: allow non-mock base token in mock stack

ramenforbreakfast:
So there is a problem with our recent test from last week, specifically with trying to have a `Mock` harness that accepts a actual base token rather than a mock, for instances where we want to interact with already an already deployed `baseToken` like `WETH`. The problem is that typescript does not like it when types that do not exist on both methods are called, and when `reset()` is called on our base token in some tests, this causes tests to fail.

So how did this get past our github actions? For some reason, our github actions are showing that core tests are running when invoked via `core:`, yet this is deceiving and seems to be some sort of cached result that it is showing, and is not actually re-running the tests.

I have created a ticket for this to investigate this when we have more time, for now I am just having to run them all locally every time, which is more bearable now that our tests take 7 mins rather than 30+.

As for this mock harness problem, I have defined two types of mock harness that share the same mock base to reduce duplication. In addition, I have gone ahead and fixed `ArbitrageBroker.test.ts` to use a fixed permit deadline which we already needed to do anyway, because the old one was failing due to the new harness taking longer than 5 seconds to setup, since it used (last time stamp + 5).


XavierEkkel (approved):
LGTM

PR #954, opened 2023-02-28
feat: add stripped down IVault interface

ramenforbreakfast:
So Balancer has a interface package that I initially was using when i started trying to implement this. There were multiple problems that made this much less straightforward:
- Balancer uses its own openzeppelin contracts, which causes the compiler to complain because there are duplicate named references to contracts that are slightly different (not in any meaningful way though).
- Balancer's contracts have some circular imports, which meant I could not just isolate the `IVault` interface, and copy it over and source the openzeppelin import from our openzeppelin library. This is because the circular imports would point to the `IVault `interface from the package, and not the one I tried to modify.

So instead, I thought the most sensible way, was to copy the `IVault` interface, and isolate only the items I needed. Besides removing most of the interface, I did not edit anything other than the headlining dev comment summarizing the interface.


XavierEkkel (approved):
LGTM

PR #955, opened 2023-02-28
feat: getters and setters for wsteth pools

ramenforbreakfast:
So Balancer pools vary from Uniswap in that there is a separate contract for where the assets sit in a pool, and then there can be multiple pools that have a claim on those assets, this contract where the assets are stored is called the **vault**. 

The **pool** itself, is referenced by a `bytes32` id, and must be specified when creating a swap. However, the user is not meant to interact with the **pool** directly, you tell the **vault** how much you want, and then the **vault** will send your trade to the specified **pool**, ensuring that the **pool** matches the **vault**'s asset pair.

Note that this **vault**, is not doing any routing even though it has context of the **pool**'s attached to it, which is why you still have to specify the **pool**'s ID.


apps/smart-contracts/core/test/unit/DepositTradeHelper.test.ts
```diff
@@ -78,6 +88,53 @@ describe('=> DepositTradeHelper', () => {
...
 
+  describe('# setWstethPoolId', () => {
+    it('reverts if not owner', async () => {
+      await expect(depositTradeHelper.connect(user).setWstethPoolId(TEST_POOL_ID)).revertedWith(
+        'Ownable: caller is not the owner'
+      )
+    })
+
+    it('sets to non-zero id', async () => {
+      expect(await depositTradeHelper.getWstethPoolId()).not.eq(TEST_POOL_ID)
```
XavierEkkel, line 99: add expect that it's not zero

apps/smart-contracts/core/contracts/DepositTradeHelper.sol
```diff
@@ -115,6 +123,10 @@ contract DepositTradeHelper is
     IWETH9(address(_baseToken)).withdrawTo(recipient, wethAmount);
   }
 
+  function setWstethPoolId(bytes32 wstethPoolId) external override onlyOwner {
+    _wstethPoolId = wstethPoolId;
```
XavierEkkel, line 127: need to add event in interface + emit it here

apps/smart-contracts/core/test/unit/DepositTradeHelper.test.ts
```diff
@@ -78,6 +88,53 @@ describe('=> DepositTradeHelper', () => {
...
+
+      await depositTradeHelper.connect(deployer).setWstethPoolId(TEST_POOL_ID)
+
+      expect(await depositTradeHelper.getWstethPoolId()).eq(TEST_POOL_ID)
+    })
+
+    it('emits WstethPoolIdChange event', async () => {
+      const tx = await depositTradeHelper.connect(deployer).setWstethPoolId(TEST_POOL_ID)
+
+      expect(tx).emit(depositTradeHelper, 'WstethPoolIdChange').withArgs(TEST_POOL_ID)
```
XavierEkkel, line 134: this test wont pass

apps/smart-contracts/core/contracts/DepositTradeHelper.sol
```diff
@@ -15,17 +15,25 @@ contract DepositTradeHelper is
   ICollateral private immutable _collateral;
   IERC20 private immutable _baseToken;
   ISwapRouter private immutable _swapRouter;
+  IVault private immutable _wstethVault;
```
XavierEkkel, line 18: nit: change to use proper camel casing (`wstEth` rather than `wsteth` and `WstEth` rather than `Wsteth`) throughout all these files

XavierEkkel (approved):
LGTM after comments

PR #957, opened 2023-02-28
feat: add BalancerStore with wstETH prices (wip)

MeLlamoPablo:
https://www.notion.so/prepo/Create-Store-for-Balancer-s-Pool-ce325f2b29444cba920873b60f965344

This PR adds a `BalancerStore` that queries the [50% wstETH 50% ETH Balancer Pool on Arbitrum](https://app.balancer.fi/#/arbitrum/pool/0x36bf227d6bac96e2ab1ebb5492ecec69c691943f000200000000000000000316) to check the `wstETH` price denominated in `WETH`.

This store is a bit different in that it doesn't extend `ContractStore`. This is because the `querySwap` function on the `BalancerQueries` contract is not `view` [due to technical limiations](https://docs.balancer.fi/reference/contracts/query-functions.html), but it still needs to be called with `eth_call`. This makes it not compatible with our MultiCall, and thus the `BalancerStore` calls it directly via `coreProvider.call`.

The price is refreshed every 90 blocks (~30s on Arbitrum).

This PR only implements the data fetching mechanism. I'll send a subsequent PR to actually display the correct amount of ETH in the UI.


apps/frontend/core-dapp/src/stores/BalancerStore.ts
```diff
@@ -0,0 +1,131 @@
...
+      async (blockNumber) => {
+        const { wstEthPriceInEthBN } = this
+
+        if (
+          wstEthPriceInEthBN !== undefined &&
+          blockNumber !== undefined &&
+          // Only query every 90 blocks (~30s)
+          blockNumber % 90 !== 0
+        )
+          return
```
chrisling-dev, line 41: Wdyt if we rely on timestamp instead? I monitored how fast blocks are processed on arbitrum for awhile, it doesn't seem so stable sometimes. It basically generates 30-40 blocks every 1-2 seconds, sometimes more/less for short/longer interval

Because of that, I feel like it's better to use a timestamp to control this, for example on our Web3Store, if you check the frequency of handleNewBlock, it's actually being triggered 30-40 times every second (can check this by console.log in the function), but if you inspect the network request, we're only sending 1 request every second, this is because we've implemented a 1sec stall timeout when creating the provider.

I reckon we could use the same approach as provider which is to fire a call only if the previous call was fired more than 1sec ago

So I'd probably do something like this (I've tested this timestamp approach locally btw): 

```suggestion
      async () => {
        const now = new Date().getTime()
        if (this.lastCalled + SEC_IN_MS > now) return
        this.lastCalled = now
```

apps/frontend/core-dapp/src/stores/BalancerStore.ts
```diff
@@ -0,0 +1,131 @@
...
+const WETH_DECIMALS = 18
+
+const wstEthWethBalancerPoolIdByNetwork: Partial<Record<SupportedNetworks, string>> = {
+  arbitrumOne: '0x36bf227d6bac96e2ab1ebb5492ecec69c691943f000200000000000000000316',
+}
+
+export class BalancerStore {
+  private readonly balancerQueriesInterface = BalancerQueriesAbi__factory.createInterface()
+  private readonly root: RootStore
+  private wstEthPriceInEthBN: BigNumber | undefined = undefined
```
chrisling-dev, line 23: Should we keep this accessible to other stores for calculations?
```suggestion
  wstEthPriceInEthBN: BigNumber | undefined = undefined
```

apps/frontend/core-dapp/src/stores/BalancerStore.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  constructor(root: RootStore) {
+    this.root = root
```
chrisling-dev, line 26:
```suggestion
  private wstEthPriceInEthBN: BigNumber | undefined = undefined

  constructor(private root: RootStore) {
```

apps/frontend/core-dapp/src/stores/BalancerStore.ts
chrisling-dev: Also since there's still chance of getting rate limited, I'd use this timestamp approach to add a cool off period if an error is thrown, I'll provide example in the next comment

apps/frontend/core-dapp/src/stores/BalancerStore.ts
```diff
@@ -0,0 +1,131 @@
...
+      to: balancerQueriesAddress,
+      data,
+    })
+
+    const [decodedResponse] = this.balancerQueriesInterface.decodeFunctionResult(
+      'querySwap',
+      encodedResponse
+    )
+
+    return decodedResponse
```
chrisling-dev, line 129: Try catch this since call may throw error

```suggestion

    try {
      const encodedResponse = await coreProvider.call({
        to: balancerQueriesAddress,
        data,
      })

      const [decodedResponse] = this.balancerQueriesInterface.decodeFunctionResult(
        'querySwap',
        encodedResponse
      )

      return decodedResponse
    } catch (e) {
      runInAction(() => {
        // cool off for 10secs if rate limited
        this.lastCalled = new Date().getTime() + 10 * SEC_IN_MS
      })
      return this.wstEthPriceInEthBN
    }
```

apps/frontend/core-dapp/src/stores/BalancerStore.ts
```diff
@@ -0,0 +1,131 @@
...
+          wstEthPriceInEthBN !== undefined &&
+          blockNumber !== undefined &&
+          // Only query every 90 blocks (~30s)
+          blockNumber % 90 !== 0
+        )
+          return
+
+        const price = await this.getWstEthPrice()
+        if (price !== undefined) {
+          if (wstEthPriceInEthBN !== undefined && price.eq(wstEthPriceInEthBN)) return
```
chrisling-dev, line 45: Don't need to do this check here because mobx will memoize values for us, try removing this and console.log somewhere that is using this value and you'll see it won't trigger any changes if the values are same

```suggestion
```

chrisling-dev (changes requested):
LGTM after comments

apps/frontend/core-dapp/src/stores/BalancerStore.ts
chrisling-dev: also this way if we get rate limited, at least it will fail gracefully by returning the cached price, and wait for the rate limit to recover

chrisling-dev:
Follow up on rate limiting, if you still experience rate limiting issue after the changes above, it's probably not an issue with the balancer store itself but more of problem from web3store, I've created a ticket to investigate that now https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=6b4403e1ecc748a48a63e32a1f35e012&p=d8a2b41723e64d71994a65611b573253&pm=s

apps/frontend/core-dapp/src/stores/BalancerStore.ts
MeLlamoPablo: I would rather keep everything that we don't need to expose *right now* private, and then refactor if we do need to access it outside stores.

Also I wish Javascript had an easy way of doing `private set` :/

apps/frontend/core-dapp/src/stores/BalancerStore.ts
MeLlamoPablo: > also this way if we get rate limited, at least it will fail gracefully by returning the cached price, and wait for the rate limit to recover

I don't think we want to return the cached price in this function (since it is only called by the reaction to set the price), and in the future we will want to call this function with different wstETH amounts.

chrisling-dev: ahh gotcha, could it return undefined then? I'd like that when rate limited, it doesn't throw error that are uncaught

MeLlamoPablo: I've added the try/catch [here](https://github.com/prepo-io/prepo-monorepo/pull/957/files#diff-f136e478490ad9b496d51f6959d56d269d288555c8350d40258b7aa1b9e58ee4R46-R55). I don't want to add it on the function because we may want to reuse it somewhere else (i.e. in the withdraw page) and we may want to add our custom error handler there.

MeLlamoPablo: Also, I'm using a 30s interval which I think it's big enough to not need the extra 10s cooldown

apps/frontend/core-dapp/src/stores/BalancerStore.ts
MeLlamoPablo: Changed it to a timestamp approach every 30s. 3s seems excessively short, I don't think we need that much accuracy for displaying prices.

chrisling-dev: I think a shorter refresh is better tho. This lets large balances see their balance ticking every few seconds which is more exciting. Since the calls on Multicall are handled properly now (on average 4sec per call), I think we can use 3s here. I had the same configuration run locally for over 10mins without getting any rate limit error.

apps/frontend/core-dapp/src/stores/BalancerStore.ts
```diff
@@ -0,0 +1,136 @@
...
+  arbitrumOne: '0x82aF49447D8a07e3bd95BD0d56f35241523fBab1',
+}
+
+const WETH_DECIMALS = 18
+
+const wstEthWethBalancerPoolIdByNetwork: Partial<Record<SupportedNetworks, string>> = {
+  arbitrumOne: '0x36bf227d6bac96e2ab1ebb5492ecec69c691943f000200000000000000000316',
+}
+
+const getNextCallMinimumTime = addSeconds(30)
```
chrisling-dev, line 23: let's make this 3 secs, the average interval between each `.on('block')`'s callback is 3.9 - 4.1 secs. This means at max we'll only send out about 2 requests every 4 seconds, shouldn't get rate limited this way

chrisling-dev (approved):

apps/frontend/core-dapp/src/stores/BalancerStore.ts
MeLlamoPablo: > This lets large balances see their balance ticking every few seconds which is more exciting

Counterpoint: ticking numbers give me anxiety :) but I don't have a strong opinion here either way

PR #960, opened 2023-02-28
feat: modify wrapAndDeposit to swap weth for wsteth

ramenforbreakfast:
I have added comments to help provide some context to the swap parameters for Balancer pools as they differ greatly from the Uniswap convention.

Notice I have added a separate offchain trade param struct for the Balancer swap since although it shares elements with the Uniswap parameters, we don't need as much info and prevents confusion. The `amountOutMinimum` is actually just called `limit` in the Balancer struct, but I have changed it to the existing `amountOutMinimum` param used with Uniswap since that is a more descriptive name than `limit`.

I have also modified the TODO to reflect the change in what would need to figure out if we were to implement more complex non-base token tests for the new function.


ramenforbreakfast:
Updated to use `IVault`'s existing `queryBatchSwap` method which allows static queries of swaps without needing assets. This will become more important when writing `withdrawAndUnwrap` tests since the caller will not have `wstETH` to simulate a swap with.

apps/smart-contracts/core/test/integration/DepositTradeHelper.test.ts
```diff
@@ -150,14 +172,28 @@ describe('=> DepositTradeHelper', () => {
...
+     * that to the swap.
      */
+    before(async () => {
+      expectedWstETHFromDepositSwap = await getBalancerSingleSwapQuote(
+        TEST_DEPOSIT_AMOUNT,
+        weth.address,
+        wsteth.address
+      )
+      balancerParamsForDeposit = {
+        amountOutMinimum: expectedWstETHFromDepositSwap.mul(99).div(100),
```
XavierEkkel, line 192: add comment that this just assumes 1% slippage for testing purposes

apps/smart-contracts/core/test/integration/DepositTradeHelper.test.ts
```diff
@@ -168,26 +204,28 @@ describe('=> DepositTradeHelper', () => {
...
-          .connect(user)
-          .wrapAndDeposit(user.address, { value: (await user.getBalance()).add(1) })
+        depositTradeHelper.connect(user).wrapAndDeposit(user.address, balancerParamsForDeposit, {
+          value: (await user.getBalance()).add(1),
+        })
       ).rejectedWith("sender doesn't have enough funds to send tx")
     })
 
-    it('wraps and deposits ETH if funder = recipient', async () => {
+    it('wraps ETH and swaps for WstETH if funder = recipient', async () => {
```
XavierEkkel, line 213:
```suggestion
    it('swaps into and deposits wstETH if funder = recipient', async () => {
```

XavierEkkel (approved):
LGTM after comments

PR #961, opened 2023-03-01
feat: modify withdrawAndUnwrap for WstETH base token

ramenforbreakfast:
Some context around some things that might stand out in this PR:
- Removed `IWETH9` because this is not needed anymore since the Balancer vault is handling wrapping/unwrapping.
- Modified the `getBaseTokenAmountForWithdrawal` again. This is because the way we were calculating it, was similar to how we calculated the amount needed for a deposit, which required us to perform the reverse of what the contract does.
I did the same thing here, except it introduces sometimes, a 1 wei rounding error. I realized that since we just need the amount received, we can just replicate the contract logic and not introduce small rounding discrepancies.
- I added `await` to event emission checks, this is because I noticed that some were passing that shouldn't have, and realized that we were not `await`ing `expect(tx).emit(something)` assertions, which apparently are asynchronous observations. I have created a ticket to retroactively ensure this is being done in other tests.


apps/smart-contracts/core/contracts/DepositTradeHelper.sol
```diff
@@ -31,6 +30,7 @@ contract DepositTradeHelper is
     _wstethVault = wstethVault;
     collateral.getBaseToken().approve(address(collateral), type(uint256).max);
     collateral.approve(address(swapRouter), type(uint256).max);
+    collateral.getBaseToken().approve(address(wstethVault), type(uint256).max);
```
XavierEkkel, line 33: nit: group with the other base token approval

apps/smart-contracts/core/contracts/DepositTradeHelper.sol
```diff
@@ -143,8 +144,29 @@ contract DepositTradeHelper is
...
+      // Unwraps WETH into ETH directly to recipient
+      payable(recipient),
+      false
+    );
+    _wstethVault.swap(
+      wstethSwapParams,
+      wstethFundingParams,
+      balancerParams.amountOutMinimum,
+      balancerParams.deadline
+    );
```
XavierEkkel, line 169: since this is depending on external systems, might as well add a require here at the end that checks the `amountOutMinimum` was at least received

apps/smart-contracts/core/test/integration/DepositTradeHelper.test.ts
```diff
@@ -267,15 +267,20 @@ describe('=> DepositTradeHelper', () => {
...
-     * TODO: Add tests using a non-WETH base token, 1 to ensure `withdrawAndUnwrap`
-     * reverts due to the absence of WETH functions existing on the base token,
-     * and one to ensure that if for some reason, the WETH contract returns <
-     * WETH burned in ETH, the transaction reverts. This will require creating a
-     * custom interface that combines IWETH9 and ERC20 so that these functions can
-     * be mocked.
+     * TODO: Add tests using a non-WstETH base token, need a test to ensure `wrapAndDeposit`
+     * reverts because Balancer pool should revert due to Collateral returning the fake base
+     * token and not WstETH. Also need a test if when the Balancer contract converts
+     * < provided WETH into ETH, the transaction reverts. This will be complicated because
```
XavierEkkel, line 273: similar to comment in wrapAndDeposit, may not need this if have the suggested require statement

apps/smart-contracts/core/test/integration/DepositTradeHelper.test.ts
```diff
@@ -344,65 +388,70 @@ describe('=> DepositTradeHelper', () => {
...
       )
     })
 
-    it('withdraws WETH to DepositTradeHelper contract', async () => {
-      const expectedBaseTokenFromWithdrawal = await getBaseTokenAmountForWithdrawal(
-        core.collateral,
-        collateralBalanceBefore
-      )
-
+    it('withdraws WstETH to DepositTradeHelper contract', async () => {
```
XavierEkkel, line 403:
```suggestion
    it('withdraws wstETH to DepositTradeHelper contract', async () => {
```
similarly for elsewhere

XavierEkkel (approved):
LGTM after comments

PR #963, opened 2023-03-02
fix: rework withdraw limit around the setter and initialization

ramenforbreakfast:
This is the first of two PRs around reworking withdrawal limits. I will open with the agreed upon things we wanted:
- A fixed constant percent to serve as a dynamic minimum limit based on net deposits.
- A fixed constant raw limit to serve as an absolute floor.
- The settable limit to be raw and bounded by the above minimums.

However, one thing we did not take into consideration, is that just bounding the setting of the limit is not enough. The whole point of a dynamic limit is so that it is an effective protection when the size of deposits exceeds our initial baseline. However, this dynamic limit **should also apply to the limit currently set**. This is because the size of deposits can grow to the point where the existing set limit falls below the dynamic minimum.

Hence, we also need to re-evaluate the limit everytime the `hook` is called. This is not complicated, but provides context for why I established a shared `max` function, and made `getMinGlobalWithdrawLimitPerPeriod` an `internal` function for `hook` to use when determining if the current limit is < than the dynamic minimum.

As also stated in the previous PR submitted around changing the withdraw limits, the raw minimum limit must be a function of base token decimals, and thus needs to be in the constructor.

Test inputs had to be modified, because the initial tests were not designed around withdraw limits tied to the net deposits.

I have also removed the idempotency test, since effectively, this function is not idempotent due to its external dependence on net deposits.

The next PR will contain the logic for the `hook` to decide the limit based on the minimum and set limit.


XavierEkkel (approved):
LGTM after comments

PR #964, opened 2023-03-02
feat: rework withdraw limit around hook logic

ramenforbreakfast:
So as stated in the previous PR, the hook will need to re-evaluate withdraw limits due to the dynamic nature of the minimum.
The problem here, is that the `hook` tests are already extremely long, and reworking them to accomodate all the different cases of limits would have been a nightmare to both rework + review, due to the hook logic's complexity and the tests are currently written with a set limit in mind.

So instead, I have decided to make a `public` `getEffectiveGlobalWithdrawLimitPerPeriod()` function that will be used and tested instead, which given the simplicity of the change, still provides a sufficient guarantee testing wise. In addition, unlike `getMinGlobalWithdrawLimitPerPeriod`, the FE cares about the effective limit and not the minimum. They will most likely need it to reduce the amount of separate calls they will need to determine if a withdrawal satisfies current limits.


XavierEkkel (approved):
LGTM

PR #965, opened 2023-03-02
fix: add cool off to multicall in case we get rate limited

chrisling-dev:
https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=6b4403e1ecc748a48a63e32a1f35e012&p=d8a2b41723e64d71994a65611b573253&pm=s


packages/prepo-stores/src/MulticallStore.ts
```diff
@@ -74,6 +79,10 @@ export class MulticallStore {
           })
           // errors in here won't get caught by try catch
           .catch((error) => {
+            runInAction(() => {
+              // cool off 10 seconds if we get rate limited
+              this.lastCall = new Date().getTime() + 10 * SEC_IN_MS
```
MeLlamoPablo, line 84: I would personally use `addSeconds(Date.now(), 10)` rather than doing math with our `X_IN_X` constants because it can be easier to read, and less error prone. But it's not a strong opinion, feel free to disregard.

MeLlamoPablo, line 84: Also, I think `lastCall` is confusing, because we're not setting the time of the last call here, we're setting the time of the next call. Maybe we can rename it to `cooldownUntilTime` or something like that?

packages/prepo-stores/src/MulticallStore.ts
```diff
@@ -74,6 +79,10 @@ export class MulticallStore {
           })
           // errors in here won't get caught by try catch
           .catch((error) => {
+            runInAction(() => {
```
MeLlamoPablo, line 82: Also, I think we should be checking for actual 429 errors. If a call fails for other reasons (execution reverted because the blockchain data is not updated with whatever value) I think we want to re-try as soon as possible.

packages/prepo-stores/src/MulticallStore.ts
```diff
@@ -42,6 +44,9 @@ export class MulticallStore {
   }
 
   call(): void {
+    const now = new Date().getTime()
+    if (this.lastCall + SEC_IN_MS > now) return
+    this.lastCall = now
```
MeLlamoPablo, line 49:
```suggestion
    if (this.lastCall > now) return
    this.lastCall = now + SEC_IN_MS
```

Maybe this way we can keep things consistent in both places we assign `lastCall`

MeLlamoPablo (changes requested):

MeLlamoPablo:
> Follow up on rate limiting, if you still experience rate limiting issue after the changes above, it's probably not an issue with the balancer store itself but more of problem from web3store, I've created a ticket to investigate that now https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=6b4403e1ecc748a48a63e32a1f35e012&p=d8a2b41723e64d71994a65611b573253&pm=s
> 
> Fixed here: #965

Following up with this comment from #957 - I don't think this PR fixes the underlying problem, at least not in the right way. I made my own testing infura account and we're making a lot of requests:

Screenshot 2023-03-02 at 09 07 42

Screenshot 2023-03-02 at 09 07 34

400 requests in just the two minutes where I opened the app doesn't seem reasonable.

While I agree that we should merge this PR and properly handle 429 errors, I also think that we should fix our code so that we don't even attempt to make this many requests. Storming the RPC with requests and then cooling down on 429 doesn't seem like the right approach. I can try to help with this if you want.

packages/prepo-stores/src/MulticallStore.ts
chrisling-dev: thanks for the feedback, decided not to add another package for something so simple, will definitely do this if there's something else that require more complex time logic

chrisling-dev: renamed to `nextCall` with the buffer time included in it, so if current time < nextCall it will return and will til nextCall's time

packages/prepo-stores/src/MulticallStore.ts
chrisling-dev: unfortunately multicall will throw the error without the actual http code, it throws something like CALL_EXCEPTION so I added a check with that

chrisling-dev:
> 

but we don't use infura node no? On arbitrum we're interacting with `https://arb1.arbitrum.io/rpc`

Did you test this on arbitrum or goerli? And is this result still happening after this PR's changes?

chrisling-dev:
> > Follow up on rate limiting, if you still experience rate limiting issue after the changes above, it's probably not an issue with the balancer store itself but more of problem from web3store, I've created a ticket to investigate that now https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=6b4403e1ecc748a48a63e32a1f35e012&p=d8a2b41723e64d71994a65611b573253&pm=s
> > Fixed here: #965
> 
> Following up with this comment from #957 - I don't think this PR fixes the underlying problem, at least not in the right way. I made my own testing infura account and we're making a lot of requests:
> 
> Screenshot 2023-03-02 at 09 07 42
> 
> Screenshot 2023-03-02 at 09 07 34
> 
> 400 requests in just the two minutes where I opened the app doesn't seem reasonable.
> 
> While I agree that we should merge this PR and properly handle 429 errors, I also think that we should fix our code so that we don't even attempt to make this many requests. Storming the RPC with requests and then cooling down on 429 doesn't seem like the right approach. I can try to help with this if you want.

So I left mine on for 2 mins, the total network requests sent from browser was about 177 reqs within 2 mins, and this includes a bunch of requests for svg, font, other js files and at least half of these are `preflight` requests, so in reality it should be way less than 100 requests every 2 mins. In fact, we can kinda calculate this, we're sending on average 2 requests every 4 seconds, 1 req for multicall and the other for balancer's query, that gives us only about 60 reqs every 2 mins

I'd say 30reqs/min/client is pretty reasonable, were you testing w multiple tabs on?

Screenshot 2023-03-03 at 2 06 13 PM

MeLlamoPablo:
> but we don't use infura node no? On arbitrum we're interacting with `https://arb1.arbitrum.io/rpc`
> 
> Did you test this on arbitrum or goerli? And is this result still happening after this PR's changes?

No, I was using my own Infura node becuase before this PR the Arbitrum public RPC was unusable. I tested on Arbitrum.

After the PR it works well.

> So I left mine on for 2 mins, the total network requests sent from browser was about 177 reqs within 2 mins, and this includes a bunch of requests for svg, font, other js files and at least half of these are preflight requests, so in reality it should be way less than 100 requests every 2 mins. In fact, we can kinda calculate this, we're sending on average 2 requests every 4 seconds, 1 req for multicall and the other for balancer's query, that gives us only about 60 reqs every 2 mins
> 
> I'd say 30reqs/min/client is pretty reasonable, were you testing w multiple tabs on?

Yeah, on this PR I counted 35 requests per minute. I still think that's a bit on the higher end. Infura charges $0.0002 per request. That would be $0.42 per user per hour. 10k users on the app for just six minutes per day for a month would cost over $12k.

I know we don't use Infura, I'm just using these figures as an example.

packages/prepo-stores/src/MulticallStore.ts
MeLlamoPablo: What do you mean? We already have `date-fns` in the project

MeLlamoPablo (approved):
Anyway, approving since the change got fixed, we can follow up on the discussion on another PR if necessary

chrisling-dev:
> > but we don't use infura node no? On arbitrum we're interacting with `https://arb1.arbitrum.io/rpc`
> > Did you test this on arbitrum or goerli? And is this result still happening after this PR's changes?
> 
> No, I was using my own Infura node becuase before this PR the Arbitrum public RPC was unusable. I tested on Arbitrum.
> 
> After the PR it works well.
> 
> > So I left mine on for 2 mins, the total network requests sent from browser was about 177 reqs within 2 mins, and this includes a bunch of requests for svg, font, other js files and at least half of these are preflight requests, so in reality it should be way less than 100 requests every 2 mins. In fact, we can kinda calculate this, we're sending on average 2 requests every 4 seconds, 1 req for multicall and the other for balancer's query, that gives us only about 60 reqs every 2 mins
> > I'd say 30reqs/min/client is pretty reasonable, were you testing w multiple tabs on?
> 
> Yeah, on this PR I counted 35 requests per minute. I still think that's a bit on the higher end. Infura charges $0.0002 per request. That would be $0.42 per user per hour. 10k users on the app for just six minutes per day for a month would cost over $12k.
> 
> I know we don't use Infura, I'm just using these figures as an example.

ahh gotcha, I think refetching something every 3-4 seconds is pretty reasonable, after since everyone would be querying from different IP address anyway, the chance of a single IP getting limit rated would be pretty low after this PR. But anyway thanks for looking into this and for reviewing this!

PR #966, opened 2023-03-02
Feat/lido apr

chrisling-dev:
https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=6b4403e1ecc748a48a63e32a1f35e012&p=4561d545b67d40e9bf599f2ec330ffb4&pm=s


apps/frontend/core-dapp/src/pages/portfolio/deposit.tsx
```diff
@@ -1,16 +1,23 @@
...
+    <DepositPage apr={apr} />
   </>
 )
 
+export const getStaticProps: GetStaticProps = async () => {
+  const res = await fetch('https://stake.lido.fi/api/sma-steth-apr')
+  const apr = await res.json()
+  return {
+    props: { apr },
+  }
```
MeLlamoPablo, line 21:
```suggestion
  return {
    props: { apr },
    revalidate: 3600,
  }
```

apps/frontend/core-dapp/src/features/deposit/BalanceBox.tsx
```diff
@@ -0,0 +1,112 @@
...
+  gap: ${spacingIncrement(10)};
+  overflow: hidden;
+  padding: ${spacingIncrement(12)} ${spacingIncrement(16)};
+  position: relative;
+  width: 100%;
+`
+
+const LogoWrapper = styled.div`
+  left: 50%;
+  left: 0;
```
MeLlamoPablo, line 29:
```suggestion
  left: 0;
```

apps/frontend/core-dapp/src/features/deposit/BalanceBox.tsx
```diff
@@ -0,0 +1,112 @@
...
+              fill="#6264D9"
+              fillOpacity="0.1"
+            />
+          </svg>
+        </LogoWrapper>
```
MeLlamoPablo, line 106: The logo isn't showing for me on Firefox. It does show on Chrome.

Screenshot 2023-03-03 at 11 12 47

Also:

* Can we put the logo in a separate SVG and import it as `background: url()`? [Seems simpler. I did it like this on the Portfolio page](https://github.com/prepo-io/prepo-monorepo/blob/872f3adfe349a777e9dac247c14c10cc4453e682/apps/frontend/core-dapp/src/features/portfolio/PortfolioValue.tsx#L12-L14).
* If not, can we pull the SVG to a separate component on a separate file?

MeLlamoPablo (changes requested):
Looks great! Requesting changes due to the logo issue, otherwise LGTM

apps/frontend/core-dapp/src/features/deposit/BalanceBox.tsx
chrisling-dev: oh thanks for catching this, it was just a width issue haha

I prefer using SVG tbh (and we should be using svg whenever possible) because it renders the fastest as a simple html element rather than having to make another network request to fetch another static file then rendering it as an image. For example the portfolio page, if you fetch on production, sometimes it takes 0.5 ~ 1s to load the image if the cache is invalidated: 

Screenshot 2023-03-06 at 10 49 57 AM

MeLlamoPablo: > For example the portfolio page, if you fetch on production, sometimes it takes 0.5 ~ 1s to load the image if the cache is invalidated

I think this is Vercel being slow. In my case the same request took 285ms (which still feels quite high).

Also, I don't think making an extra HTTP request here is a big deal (especially with HTTP2), and it will probably be cache for the rest of the application lifetime. As a counterpoint, rendering a SVG also incurs in a runtime performance overhead because React needs to process it.

I think basing this decision on performance would be premature optimization, because either inlining the SVG or using it as a background URL probably has a tiny impact. However, I do agree with the position flexibility reasoning, so let's keep this as-is.

MeLlamoPablo (approved):

PR #967, opened 2023-03-02
feat: use ETH values for wstETH on the withdraw page

MeLlamoPablo:
https://user-images.githubusercontent.com/11708035/222498822-36789d62-f899-4fe4-b25e-8be0a7cd12c5.mp4

This PR updates the withdraw page to use the proper ETH values (preETH) calculated from the wstETH balance. The following is updated:

* The following balances are converted from wstETH and displayed in ETH
  * The preETH balance (near the MAX button)
  * The withdraw limit
  * The PPO reward ETH value
* Whenever the user types a preETH balance, it is internally converted back to wstETH.

   Then, this we request a quote from Balancer for this wstETH to convert it back to ETH again. This ensures that the "received ETH amount" that we show to the user is accurate and accounts for the swap price impact.
* If the user clicks MAX, we update the input dynamically if the ETH value of wstETH increases (which is likely). This ensures that we don't leave wstETH dust in the user wallet.

# FE PR Checklist

- [X] Self-reviewed each line of code, preempted questions about logic or decisions
- [X] Noted any intentional deviations from design
- [X] Tried hard to break it and failed, think like a malicious user trying to find edge cases to break the app
- [X] Tested on supported device sizes
    - [X] iPhone 12 Pro
    - [X] iPad Air
    - [X] 1920x1080 (how to add custom size: [https://umaar.com/dev-tips/51-add-new-device/](https://umaar.com/dev-tips/51-add-new-device/))
- [X] Tested all widths between 1920px wide to 390px wide and confirmed no weird breakages at widths in between


apps/frontend/core-dapp/src/stores/BalancerStore.ts
```diff
@@ -14,7 +14,8 @@ const wethAddressByNetwork: Partial<Record<SupportedNetworks, string>> = {
   arbitrumOne: '0x82aF49447D8a07e3bd95BD0d56f35241523fBab1',
 }
 
-const WETH_DECIMALS = 18
+export const WETH_DECIMALS = 18
+const WSTETH_DECIMALS = 18 // TODO pull from baseTokenStore?
```
MeLlamoPablo, line 18: @XavierEkkel I'm still confused about `baseTokenStore`. When we say base token, are we referring to WETH or wstETH?

XavierEkkel: base token (the asset that backs the prePO collateral tokens) is `wstETH`

apps/frontend/core-dapp/src/stores/BalancerStore.ts
```diff
@@ -44,10 +45,11 @@ export class BalancerStore {
         this.lastCalled = Date.now() as DateTimeInMs
 
         try {
-          const price = await this.getWstEthPrice()
-          if (price !== undefined) {
+          const oneWstEth = BigNumber.from(10).pow(18)
```
chrisling-dev, line 48: Maybe better to just use parseEther or at least have a reference to wstETH decimals?

```suggestion
          const oneWstEth = parseEther('1')
```

or 

```suggestion
          const oneWstEth = BigNumber.from(10).pow(WSTETH_DECIMALS)
```

apps/frontend/core-dapp/src/features/withdraw/WithdrawPage.tsx
```diff
@@ -6,25 +6,37 @@ import WithdrawWarning from './WithdrawWarning'
...
+      {insufficientLiquidity && (
+        <Alert
+          message={
+            <p>
+              There&apos;s not enough liquidity on the Balancer pool. Try withdrawing a smaller
+              amount.
+            </p>
+          }
+        />
+      )}
```
chrisling-dev, line 39: pending confirmation on this, might just wanna show Insufficient Liquidity in button + disable withdraw

apps/frontend/core-dapp/src/features/withdraw/WithdrawStore.ts
```diff
@@ -22,14 +25,81 @@ export type WithdrawLimit =
...
+    value: '',
+  }
+  private withdrawalMarketValueInEth:
+    | { status: 'not-queried' | 'not-enough-liquidity' }
+    | { status: 'queried'; value: BigNumber } = { status: 'not-queried' }
 
   constructor(public root: RootStore) {
     makeAutoObservable(this, {}, { autoBind: true })
+
+    const debouncedUpdatePrice = debounce(async (withdrawalAmountInWstEthBN: BigNumber) => {
```
chrisling-dev, line 41: could we move this function out of constructor so it's more organized + can reuse it

chrisling-dev (changes requested):
great approach with `withdrawalAmountInWstEthBN`, will wait for fixes to do another review

apps/frontend/core-dapp/src/features/withdraw/WithdrawStore.ts
```diff
@@ -105,17 +191,41 @@ export class WithdrawStore {
...
+    const { balanceOfSigner, tokenBalanceFormatInEth } = this.root.preCTTokenStore
+    const { withdrawalAmountInEth } = this
+    if (
+      tokenBalanceFormatInEth === undefined ||
+      withdrawalAmountInEth === '' ||
+      balanceOfSigner === undefined
+    )
+      return undefined
+
+    const withdrawalPercent = +withdrawalAmountInEth / +tokenBalanceFormatInEth
```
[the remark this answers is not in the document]
MeLlamoPablo: Great catch!

MeLlamoPablo:
Thanks for your review, addressed your comments, pending confirmation on the liquidity thing.

I also removed `WETH_DECIMALS` and `WSTETH_DECIMALS`. Replaced the latter with `baseTokenStore.decimalsNumber` and the former with a new `wethStore.decimalsNumber`. This way we reuse our current network detection mechanism.

apps/frontend/core-dapp/src/features/withdraw/WithdrawPage.tsx
chrisling-dev: okay so if insufficient liquidity, the button would show "Insufficient Liquidity" and disabled, but without this message

apps/frontend/core-dapp/src/features/withdraw/WithdrawStore.ts
```diff
@@ -105,17 +169,40 @@ export class WithdrawStore {
...
+      tokenBalanceFormatInEth === undefined ||
+      withdrawalAmountInEth === '' ||
+      balanceOfSigner === undefined
+    )
       return undefined
-    return this.withdrawalAmountBN.sub(this.withdrawalFeesAmountBN)
+
+    if (+tokenBalanceFormatInEth === 0) return BigNumber.from(0)
+
+    const withdrawalPercent = +withdrawalAmountInEth / +tokenBalanceFormatInEth
```
chrisling-dev, line 198: a better practice imo would be to use BN here instead of +string to prevent decimals loss:
`const percentBN = withdrawalAmountBN.mul(BigNumber.from(10).pow(18)).div(balanceOfSignerInEthBN)`

then 
`balanceOfSigner.mul(percentBN).div(BigNumber.from(10).pow(18))`

But feel free to disregard if you think there's little to no impact for this use case

chrisling-dev, line 198: btw I just realize dividing by token balance doesn't make sense. The withdrawalAmountInWstEth should just be withdrawalEthAmount / staticPriceFromBalancerStore, which would give us the right portion of wstETH to withdraw.

An example

- wstEthBalance: 0.5
- wstETH->ETH: 1.2
- balanceInEth: 0.6

Scenario w different withdrawal amount:
- Input 1.0 ETH:
   wstETHAmount = 1 / 1.2 = 0.8333 ✅ 
- Input 0.6 ETH:
   wstETHAmount = 0.6/1.2 = 0.5 ✅ 
- Input 0.3 ETH: 
   wstETHAmount = 0.3/1.2 = 0.3 ✅ 
- Input 0.1 ETH:
   wstETHAmount = 0.1/1.2 = 0.0833 ✅ 


Whereas using token balance will give us wrong output:
- Input 1 ETH:
   wstETHAmount = 1 / 0.6 * 0.6 = 1 ❌ 
- Input 0.6 ETH:
   wstETHAmount = 0.6 / 0.6 * 0.6 = 0.6 ❌ 


So the actual logic should be much simpler

```
if price === undefined return undefined
return withdrawalAmountInEthBN.div(priceBN) // this is the wstETHAmount to be used for quoting
```

Do remember to take care of decimals in the div part, I'd recommend having a reverse of `getWstEthAmountInEth` in balancerStore

chrisling-dev (changes requested):

apps/frontend/core-dapp/src/features/withdraw/WithdrawStore.ts
MeLlamoPablo: I don't think that my current code is incorrect, but I will try to see if I can simplify it nontheless.

> Whereas using token balance will give us wrong output:
> 
>     * Input 1 ETH:
>       wstETHAmount = 1 / 0.6 * 0.6 = 1 ❌
> 
>     * Input 0.6 ETH:
>       wstETHAmount = 0.6 / 0.6 * 0.6 = 0.6 ❌

The current code works by calculating how much does the user want to withdraw in terms of % of their balance (in ETH, because that's what the user sees)

So for:

* wstEthBalance: 0.5
* wstETH->ETH: 1.2
* balanceInEth: 0.6

The values would be:

* Input 1 ETH:
    wstETHAmount = 1 / 0.6 * 0.5 = 0.8333 ✅
* Input 0.6 ETH:
    wstETHAmount = 0.6 / 0.6 * 0.5 = 0.5 ✅

For the first case, if the user wants to withdraw 1 ETH while having 0.6 ETH. That means they're withdraing 166% of their portfolio. Their wstETH balance is 0.5, so 166% of 0.5 is 0.8333.

Now, if you look at my formula: 

```
1 / 0.6 * 0.5
```

And yours: 

```
1 / 1.2
```

You'll realize that both can be expressed as:

```
1 / <wstETH price in ETH>
```

This is because:

```
<wstETH price in ETH> = <ETH balance> / <wstETH balance> = 0.6 / 0.5 = 1.2
```

So in reality, my code is indirectly using your way too 😁 So yes, I agree that my code can be simplified to your formula.

chrisling-dev: oh yea that's right, the only thing is when you have 0 balance, it returns 0 so if I try to input 100 to test the withdraw function, the received amount shows 0. I think the correct UX would be to show what you'll receive even if your balance is 0, but disable button with "Insufficient Balance"

MeLlamoPablo: Great point about the balance. Updated!

apps/frontend/core-dapp/src/features/withdraw/WithdrawStore.ts
```diff
@@ -22,14 +25,59 @@ export type WithdrawLimit =
...
   constructor(public root: RootStore) {
     makeAutoObservable(this, {}, { autoBind: true })
+
+    const debouncedUpdatePrice = debounce(this.updateWithdrawalMarketValue.bind(this), 300)
+
+    reaction(
+      () => this.withdrawalAmountInWstEthBN,
+      async (withdrawalAmountInWstEthBN) => {
+        if (withdrawalAmountInWstEthBN === undefined) return
+        this.withdrawalMarketValueInEth = { status: 'not-queried' }
```
chrisling-dev, line 47: Could we remove this line so it doesn't show a loading skeleton everytime it reloads?

chrisling-dev (approved):
Ideally we'd want to remove the `= { status: 'not-queried' }`. But don't change in this PR to speed things up

apps/frontend/core-dapp/src/features/withdraw/WithdrawStore.ts
MeLlamoPablo: Yeah I thought about that. There are two reasons why `withdrawalAmountInWstEthBN` might change:

* The user changed the amount
* The wstETH price changed

Ideally, we would set `not-queried` on the former and update silently on the latter, but that would be more complex. For now, I'll just remove it.

MeLlamoPablo:
> Ideally we'd want to remove the `= { status: 'not-queried' }`. But don't change in this PR to speed things up

What would you replace it with?

chrisling-dev (approved):
LGTM

PR #969, opened 2023-03-03
chore: add awaits to event emission assertions

ramenforbreakfast:
Some caveats/context
- Some approval event assertions were removed like done in previous PRs, because `transferFrom` actually spends down the allowance via the `approve` function. So I now instead, verify by checking that `permit` was called.
- I also included changes to the `Collateral` amount require statement tests because I carelessly made that modification without re-running the tests.
- For the `ArbitrageBroker.test.ts` I realized that event emission tests weren't going to work, because for the `ArbitrageBroker` tests, I had earlier encountered a puzzling issue where the fakes were not operating correctly for that test suite in particular, unless the contract was being called statically. This is a problem because static calls do not emit observable events. I would need to significantly rework the test suite to accomodate actual mocked contracts rather than fakes. I have left this as a TODO since I already spent too long today debugging this again and don't have time to address this now.


XavierEkkel (approved):
LGTM

PR #971, opened 2023-03-03
Feat/deposit page layout

chrisling-dev:
https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=6b4403e1ecc748a48a63e32a1f35e012&p=a4f1550ca8f04a149a5b3da878847d25&pm=s


apps/frontend/core-dapp/src/components/PageCard.tsx
```diff
@@ -61,6 +62,16 @@ const Title = styled.h2`
...
+const TitleWrapper = styled.div`
+  position: relative;
+`
+
+const TitleTooltipWrapper = styled.div`
+  position: absolute;
+  right: -4px;
+  top: 50%;
+  transform: translate(100%, -50%);
+`
```
MeLlamoPablo, line 74: This feels a bit brittle. As I understand, what this wants to do is place the tooltip icon to the right of the title while preserving the title centered. I think there's an easier way. We can drop `TitleTooltipWrapper` and replace `TitleWrapper` with a `Flex`:

```tsx
<Flex ml={titleTooltip ? 18 : 0} gap={4}>
  <Title>{title}</Title>
  {titleTooltip && (
    <Tooltip overlay={titleTooltip}>
      <Icon name="info" color="neutral5" height="14" width="14" />
    </Tooltip>
  )}
</Flex>
```

Since the icon size is fixed (14) we can add a `margin-left` that's equivalent to icon size + gap (14 + 4 = 18) that pushes the `Flex` to the right, and the text is visually centered.

What do you think?

MeLlamoPablo (approved):
Cool! Left nit that's not really a blocker, feel free to disregard

apps/frontend/core-dapp/src/components/PageCard.tsx
chrisling-dev: Great suggestion but I think the position: absolute approach isn't brittle at all tho. It ensures both text and tooltip to always be centered regardless of size is pretty common approach for centering something. Will keep in mind your suggestion for next thing that requires as to not waste time on reimplementing something that will look/function the same anyway. Thanks for feedback!

PR #974, opened 2023-03-03
fix: ignore "block skew detected" error

MeLlamoPablo:
https://user-images.githubusercontent.com/11708035/222770084-a3a6ac3b-40c2-4574-9ab6-6550dbfb90e0.mp4

This error happens when the user leaves the tab and comes back after a while (I think it's 1000 blocks). I think it also could happen on Goerli but Arbitrum has much faster block times which makes it more likely to happen.


chrisling-dev (approved):
Nice catch!

PR #975, opened 2023-03-03
feat: add basic /terms page

MeLlamoPablo:
https://user-images.githubusercontent.com/11708035/222780677-a776daf6-74d4-4c9c-b129-72d62a0d1f7a.mp4

This PR adds a basic `/terms` page. We can use it as a starting point and keep iterating when the design is finalized.


apps/frontend/core-dapp/src/features/terms/TermsPage.tsx
```diff
@@ -0,0 +1,100 @@
...
+`
+
+const TermsPage: React.FC = () => {
+  const [read, setRead] = useState(false)
+
+  return (
+    <Flex flexDirection="column" gap={16}>
+      <Title>prePO Terms of Service</Title>
+      <TermsContent>
+        <Warning>By using this website, you agree to our terms and conditions.</Warning>
```
chrisling-dev, line 54: don't need this line

apps/frontend/core-dapp/src/features/terms/TermsPage.tsx
```diff
@@ -0,0 +1,100 @@
...
+        <TermsList
+          onScroll={(e): void => {
+            if (!(e.target instanceof HTMLUListElement)) return
+            const remainingScrollPx =
+              e.target.scrollHeight - e.target.scrollTop - e.target.offsetHeight
+
+            if (remainingScrollPx === 0) {
+              setRead(true)
+            }
```
chrisling-dev, line 63: let's add some buffer here since scrolling on some device with touch can be tricky, can do something like remainingScrollPx < 16

chrisling-dev (changes requested):
Note on page structure: I think besides /terms page we should also wrap <Component/> in `_app` with a "TermsWall" component, where it checks if user has agreed to terms (stored in local storage), show children (<Component />) else show <TermsPage />

Only one blocker for this PR, could we make the scrollbar look nicer, can refer to sale.prepo.io (connect wallet and see the Acknowledge Terms content): https://github.com/prepo-io/prepo-monorepo-private/blob/mini_sales/apps/frontend/mini-sales/src/styles/scrollbar.css

everything else LGTM noting that we'll store read state in localStore in a separate PR

chrisling-dev (approved):
Approving this now because most UI looks good, will pick up the rest in separate PR

PR #981, opened 2023-03-03
feat: price conversion logic based on twap reading

ramenforbreakfast:
This has also been tested separately against Mean-Finance's existing deployment (manually checking prices based on current readings), but I did not see really any point in adding an integration test that would not be tantamount to retesting their contract.


ramenforbreakfast:
Reworked to directly pass in the base amount (in terms of USDC), rather than calculating a price and converting the base amount separately.

Also made a correction to `baseAmount`, needs to be `uint128` since the oracle interface requires `uint128`.

PR #984, opened 2023-03-06
fix: tooltip copy

chrisling-dev:
https://www.notion.so/prepo/All-pending-tooltip-copy-0c15503a2e79400c80fed64bad3596e7?d=df90e3ca4845464fbef7c5a767e0945c

Some other minor changes:
- Added support for custom Inter font so the ↗ icon looks good
- Fixed our custom `Link` component so we don't have to keep styling next/link everytime


apps/frontend/core-dapp/src/components/Link.tsx
```diff
@@ -1,25 +1,41 @@
...
+  noUnderline?: boolean
 }
 
-const Link: React.FC<Props> = ({ className, href, target = '_self', children, scroll }) => (
+const Link: React.FC<Props> = ({
+  className,
+  href,
+  target = '_self',
+  children,
+  noUnderline,
```
MeLlamoPablo, line 30: I personally dislike boolean variable names that contain a negation in their name. For example I tend to replace `isBalanceInsufficient` with `isBalanceSufficient (= !isBalanceInsufficient)` or `isBalanceExceeded`

In this case I'd do:

```suggestion
  underline = true,
```

Since it's a new prop, we don't need to worry about breaking anything.

apps/frontend/core-dapp/src/components/Link.tsx
```diff
@@ -1,25 +1,41 @@
 import NextLink from 'next/link'
-import styled from 'styled-components'
+import styled, { css, SimpleInterpolation } from 'styled-components'
 
-const Anchor = styled.a`
-  :hover {
-    color: ${({ theme }): string => theme.color.primary};
-  }
+const Anchor = styled.a<{ noUnderline?: boolean }>`
```
MeLlamoPablo, line 4: Let's add a `$`. Also, I don't think this prop needs to be nullable since it's internal.

```suggestion
const Anchor = styled.a<{ $noUnderline: boolean }>`
```

Or if you wanna take my suggestion:

```suggestion
const Anchor = styled.a<{ $underline: boolean }>`
```

apps/frontend/core-dapp/src/components/PpoRewardSummaryRecord.tsx
```diff
@@ -21,19 +37,24 @@ export const PPORewardSummaryRecord: FC<{
...
+            {Intl.NumberFormat(undefined, {
+              maximumFractionDigits: 2,
+              minimumFractionDigits: 1,
+              notation: 'compact',
+              signDisplay: 'exceptZero',
+            }).format(+(ppoReward ?? '0'))}
+          </PPORewardAmount>
+          <Icon name="ppo-logo" height="12" width="12" />
+        </PPORewardAmountWrapper>
+        <p>({displayEth(+(fee ?? 0))} ETH)</p>
```
MeLlamoPablo, line 56: Are we not incorporating the ETH symbol inside `displayEth`? I think that defeats the point of `displayEth` then.

MeLlamoPablo (approved):

PR #987, opened 2023-03-06
feat: add redeem button on Positions for redeemed markets

MeLlamoPablo:
https://www.notion.so/prepo/Redeem-text-on-positions-9f804f81b3dc4af39a81985435a62d93

Screenshot 2023-03-06 at 17 46 12

This PR replaces the +/- button with a single "Redeem" button on the Portfolio page for markets that are resolved. It has the same effect as the - button.

Unfortunately it isn't easy to test this PR as there's no way to open a position. I tested it by hardcoding mocked positions and markets.


chrisling-dev (approved):
code wise and screenshot LGTM, will review actual implementation when we get to portfolio page

PR #988, opened 2023-03-06
feat: add mean-finance contracts

ramenforbreakfast:
To avoid needing to fix compatibility problems with their entire repo, I have just copied their contract over into a standalone hardhat directory that follows our format. We do not need to copy over their entire test suite since we are using their contract without any modification except for the name change. If i were to do so I would need to setup env configuration, modify things to match our github linting guards, and im sure various other issues that I'm not going to waste time on.

I will add a simple deployment script in a follow up PR, since theirs is a determinstic deployment script which we don't need.


XavierEkkel (approved):
LGTM if capitalize directory as `UniswapV3Oracle`

PR #990, opened 2023-03-07
feat: multiple fixes and replaced preUSDC with preWstETH

ramenforbreakfast:
Modified deployment scripts to use `wstETH` as a base token. Also added entries for a twap based price source rather than a fixed price source.


ramenforbreakfast:
https://github.com/prepo-io/prepo-monorepo/pull/992 also removes references to `FixedUintValue` that were missed in this PR

apps/smart-contracts/core/harnesses/base.ts
```diff
@@ -293,8 +313,6 @@ export abstract class Base {
       'getAllowedMsgSenders',
       'setAllowedMsgSenders'
     )
-    if (fixedPrice !== undefined)
-      await sendTxAndWait(await this.tokenSender.fixedPrice.connect(signer).set(fixedPrice))
```
XavierEkkel, line 297: what about setting values for twapPrice (`UniswapV3OracleUnitValue`)?

XavierEkkel (approved):
LGTM after question

apps/smart-contracts/core/harnesses/base.ts
ramenforbreakfast: This is in the PR #1007 due to it being very invasive in how the token sender is setup and decided to leave those changes there to prevent having to re-change things again later.

PR #991, opened 2023-03-07
refactor: replace set contract function with more generalized simple value setter

ramenforbreakfast:
Made a minor refactoring to our deployment setters. Prior we only used this function for linking contracts, but I have extended it to any setting of an individual value, since this is simpler to generalize.


XavierEkkel (approved):
LGTM

PR #993, opened 2023-03-07
chore: remove docs from interfaces we already have not deployed

ramenforbreakfast:
Removed docs from all interfaces for contracts we plan to deploy. However, I have kept comments within interfaces we have already deployed, for example `IAccountList` and any access control contracts or withdrawERCX contracts.

I have also removed comments from the `IUniswapV3Oracle` interface that is used by `UniswapV3OracleUintvalue`, since the actual oracle contract is already deployed with Mean Finance's own  extensive documentation, and the `UniswapV3OracleUintValue` has no need for this since it only needs the interface to store a reference to the oracle.


XavierEkkel (approved):
LGTM

PR #994, opened 2023-03-07
chore: uppercase UniswapV3Oracle folder

ramenforbreakfast:
Sorry, forgot to do this.


XavierEkkel (approved):
doesn't this also need to remove/replace the lowercase version?
LGTM after that

PR #997, opened 2023-03-07
feat: new contract addresses

chrisling-dev:


MeLlamoPablo (approved):

MeLlamoPablo:
Tested and every read value seems to work. I wasn't able to deposit, but that was true before this PR too, so merging.

PR #998, opened 2023-03-07
feat: use new ABI on `wrapAndDeposit`

MeLlamoPablo:
Depends on #967.

This PR updates the ABI of `DepositTradeHelper` and its usage on `DepositTradeHelperStore`.

`wrapAndDeposit` now requires [`OffChainBalancerParams`](https://github.com/prepo-io/prepo-monorepo/blob/ac59ab0d8ace0a93e90ebbe69fdef70b4eba4b4a/apps/smart-contracts/core/contracts/interfaces/IDepositTradeHelper.sol#L17-L20) which are the `amountOutMinimum` and `deadline`. That is, respectively, the minimum amount that we will guarantee to the user after slippage, and the trade deadline, which we set to 10 minutes.

To get this value, we need to quote the ETH value the user is depositing in wstETH and apply the slippage parameter.


chrisling-dev (approved):
LGTM!

PR #999, opened 2023-03-07
feat: withdraw and unwrap

chrisling-dev:
https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=6b4403e1ecc748a48a63e32a1f35e012&p=7e625b331ca840869c2751c609d06264&pm=s


apps/frontend/core-dapp/src/stores/DepositTradeHelperStore.ts
```diff
@@ -15,10 +22,13 @@ export class DepositTradeHelperStore extends ContractStore<RootStore, SupportedC
...
     recipient: string,
-    amountInEth: BigNumber
+    amountInEth: BigNumber,
+    fee: BigNumber
   ): Promise<{ success: boolean; error?: string }> {
     try {
-      const amountInWstEth = await this.root.balancerStore.quoteEthAmountInWstEth(amountInEth)
+      const amountInWstEth = await this.root.balancerStore.quoteEthAmountInWstEth(
+        amountInEth.sub(fee)
+      )
```
MeLlamoPablo, line 31: Good catch, thanks 👍

apps/frontend/core-dapp/src/stores/CollateralStore.ts
```diff
@@ -149,4 +171,34 @@ export class CollateralStore extends Erc20Store {
...
+        { Permit },
+        {
+          owner,
+          spender,
+          value,
+          nonce,
+          deadline,
+        }
+      )
+    )
```
MeLlamoPablo, line 202: We don't seem to be using `eth-permit` at all, maybe we can uninstall it?

apps/frontend/core-dapp/src/stores/CollateralStore.ts
```diff
@@ -149,4 +171,34 @@ export class CollateralStore extends Erc20Store {
...
+    value: BigNumber,
+    deadline: number
+  ): Promise<Signature | string> {
+    const { address: verifyingContract, name, signerNonces: nonce } = this
+    const { address: owner, signer, network } = this.root.web3Store
+    if (nonce === undefined || !verifyingContract || !name || !owner || !signer)
+      return 'Please try again shortly.'
+
+    return splitSignature(
+      await signer?._signTypedData(
```
MeLlamoPablo, line 186: Also, it's a bit odd that `_signTypedData` has `_`. Maybe we're using a method we shouldn't use? Is there any other way to do `signTypedData` in ethers?

apps/frontend/core-dapp/src/stores/DepositTradeHelperStore.ts
```diff
@@ -49,4 +59,69 @@ export class DepositTradeHelperStore extends ContractStore<RootStore, SupportedC
...
+  async withdrawAndUnwrap(
+    recipient: string,
+    amountIn: BigNumber,
+    amountOut: BigNumber
+  ): Promise<{ success: boolean; error?: string }> {
+    try {
+      const amountOutMinimum = this.root.advancedSettingsStore.getAmountAfterSlippage(amountOut)
+      const { address: wstETHAddress } = this.root.baseTokenStore
+
+      // impossible unless we have forget to set addresses
```
MeLlamoPablo, line 72:
```suggestion
      // impossible unless we have forgot to set addresses
```

apps/frontend/core-dapp/src/stores/DepositTradeHelperStore.ts
```diff
@@ -49,4 +59,69 @@ export class DepositTradeHelperStore extends ContractStore<RootStore, SupportedC
...
+
+      // button should be disabled
+      if (!this.root.web3Store.signer?.provider)
+        return { success: false, error: 'Wallet not connected.' }
+
+      let permit = {
+        deadline: 0,
+        v: 0,
+        r: formatBytes32String('JUNK_DATA'),
+        s: formatBytes32String('JUNK_DATA'),
```
MeLlamoPablo, line 84: If I'm understanding this right, if `needPermitForWithdrawAndUnwrap` is false, then this is the junk data that we feed to the blockchain. I assume that it doesn't get checked.

If so, can we just do empty bytes? For transaction inputs, zero bytes cost 4 gas and non-zero bytes cost 64 gas.

apps/frontend/core-dapp/src/features/withdraw/WithdrawStore.ts
```diff
@@ -258,21 +273,21 @@ export class WithdrawStore {
...
+    if (value === undefined) return
 
     try {
-      const withdrawalAmountInEthBN = await this.root.balancerStore.quoteWstEthAmountInEth(
-        withdrawalAmountInWstEthBN
-      )
-
-      if (withdrawalAmountInEthBN !== undefined) {
-        runInAction(() => {
+      const cachedValue = value
```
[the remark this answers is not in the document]
chrisling-dev: So old request won't affect new reqs, example:

- Type 1.2
- fires a request
- change to 2.2
- fires another request
- second request resolves first and update to output for  2.2
- first request resolves and update to output for 1.2
Now the last step will cause the output to be wrong

another thing it fixes is
- type 1.2
- fires a request
- change to 2.2
- fires another request
- first request resolves, update output for 1.2
- at this point UI is showing estimated value for 1.2 but input is currently 2.2, and it wont show skeleton because first request would've set the value to something
- depending on what happens to the second request, it gives really weird UX some times

Adding this cache ensure that output will only get updated if the current value is same as when request was fired

apps/frontend/core-dapp/src/stores/CollateralStore.ts
chrisling-dev: oh did I not uninstall it? pretty sure I uninstalled but let me check again

apps/frontend/core-dapp/src/stores/CollateralStore.ts
chrisling-dev: I copied this code from SC's util function, I thought it was weird as well and tried to look for another alternative function that gives the same output but this seems to be the only thing so far

apps/frontend/core-dapp/src/stores/DepositTradeHelperStore.ts
chrisling-dev: yeap so if needPermitForWithdrawAndUnwrap is false we'll just pass in a junk permit. Good suggestion let me try that

apps/frontend/core-dapp/src/features/withdraw/WithdrawStore.ts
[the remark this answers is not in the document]
MeLlamoPablo: I don't think this is how it works. `value` is a locally scoped variable on the `updateWithdrawalMarketValue`. `cachedValue` is also locally scoped to that function. Neither is accessible outside the function, so neither can be modified by an external dependency.

Different invocations of `updateWithdrawalMarketValue` will have their own `value` and `cachedValue`.

apps/frontend/core-dapp/src/stores/CollateralStore.ts
MeLlamoPablo: I looked and it's because it's [experimental on ethers v5](https://docs.ethers.org/v5/api/signer/#Signer-signTypedData) but [stable on ethers v6](https://docs.ethers.org/v6/api/providers/#Signer-signTypedData).

Do you think migrating to v6 is a big challenge? I think it would be preferable to using an experimental method for a sensitive function :/

apps/frontend/core-dapp/src/features/withdraw/WithdrawStore.ts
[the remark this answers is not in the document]
chrisling-dev: yes you're right but we're checking cachedValue against current withdrawal amount in the next few lines, so it's checking with each `updateWithdrawalMarketValue`'s own value against the current amount:

`this.withdrawalAmountAfterFeeInWstEthBN?.eq(cachedValue)`

apps/frontend/core-dapp/src/stores/CollateralStore.ts
chrisling-dev: ahh yea should probably migrate at some point but not now haha, could you create a ticket for migrating to v6?

apps/frontend/core-dapp/src/features/withdraw/WithdrawStore.ts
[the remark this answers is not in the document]
chrisling-dev: oh somehow I misunderstood the question as why we're checking the cached value, yea I assigned it to cachedValue cus that reads better IMO, but directly using value works for me, updated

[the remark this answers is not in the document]
MeLlamoPablo: We can also rename the parameter to `cachedValue` if that reads better, but having two variables confused me

PR #1003, opened 2023-03-08
Fix/ppo reward value calculation

chrisling-dev:
https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=6b4403e1ecc748a48a63e32a1f35e012&p=e3e9e26149d24dd191a594687c01142c&pm=s


apps/frontend/core-dapp/src/stores/TokenSenderStore.ts
```diff
@@ -30,6 +45,20 @@ export class TokenSenderStore extends ContractStore<RootStore, SupportedContract
...
+      this.scaledPrice === undefined ||
+      this.multiplierDenominator === undefined
+    )
+      return undefined
+    return this.scaledPrice.mul(this.multiplierDenominator).div(this.priceMultiplier)
+  }
+
+  get priceMultiplier(): BigNumber | undefined {
+    return this.getPriceMultiplier()?.[0]
+  }
```
MeLlamoPablo, line 60: Nit: these could be private since we don't use them outside

PR #1005, opened 2023-03-08
feat: handle price impact on deposit and withdraw

MeLlamoPablo:
Screenshot 2023-03-08 at 14 01 21

This PR calculates the price impact of a wstETH trade (either deposit or withdraw) and blocks trades with a price impact higher thah 1%. For reference, it takes > $5M on the Balancer pool to get to above 1% price impact, so it is a case we won't see very often.

The price impact is the difference in percent between the fair value of a trade (we calculate that by multiplying the ETH amount times the wstETH price) and the actual output offered by Balancer.

* On withdraw, we calculate the price impact when the user types the input, which is when we query balancer.
* On deposit, we actually don't show the Balancer rates on the front-end. We calculate the price impact after the user clicks "Deposit", which is right before signing the transaction. If the price impact is too high, we then show an error.

# FE PR Checklist

- [ ] Tested on supported device sizes
    - [ ] iPhone 12 Pro
    - [ ] iPad Air
    - [ ] 1920x1080 (how to add custom size: [https://umaar.com/dev-tips/51-add-new-device/](https://umaar.com/dev-tips/51-add-new-device/))
- [ ] Tested all widths between 1920px wide to 390px wide and confirmed no weird breakages at widths in between


apps/frontend/core-dapp/src/features/withdraw/WithdrawStore.ts
```diff
@@ -264,21 +268,30 @@ export class WithdrawStore {
...
+          if (
+            this.root.advancedSettingsStore.isPriceImpactTooHigh(withdrawalAmountInEth.priceImpact)
+          ) {
+            this.withdrawalMarketValueInEth = { status: 'price-impact-too-high' }
+          } else {
+            this.withdrawalMarketValueInEth = {
+              status: 'queried',
+              value: withdrawalAmountInEth.value,
+            }
+          }
```
chrisling-dev, line 286: I think we can undo the changes here, and just add a `highImpact` prop to `status: 'queried'`

Because technically you will still have an estimated received amount, it's just that we disable the button to not let you perform a bad trade.

```suggestion
          this.withdrawalMarketValueInEth = {
              status: 'queried',
              value: withdrawalAmountInEth.value,
              highImpact: this.root.advancedSettingsStore.isPriceImpactTooHigh(withdrawalAmountInEth.priceImpact)
            }
```

alternative would be to still pass value to when status is 'price-impact-too-high'

apps/frontend/core-dapp/src/features/withdraw/WithdrawStore.ts
```diff
@@ -264,21 +268,30 @@ export class WithdrawStore {
...
         })
       }
       // eslint-disable-next-line
     } catch (e: any) {
       if (
         e.code === 'CALL_EXCEPTION' &&
         Array.isArray(e.errorArgs) &&
-        e.errorArgs[0] === 'BAL#001'
+        (e.errorArgs[0] === 'BAL#001' || e.errorArgs[0] === 'BAL#526')
```
chrisling-dev, line 294: could we add a comment here with referece to balancer's doc
```suggestion
                // ref: https://docs.balancer.fi/reference/contracts/error-codes.html
                (e.errorArgs[0] === 'BAL#001' || e.errorArgs[0] === 'BAL#526')
```

apps/frontend/core-dapp/src/features/withdraw/WithdrawSummary.tsx
```diff
@@ -10,11 +10,19 @@ import { displayEth } from '../../utils/number-utils'
...
+    receivedAmountInEth,
+    withdrawalAmountInEth,
+    withdrawalFeesAmount,
+  } = withdrawStore
 
   // empty input or 0 input
-  if (withdrawalAmountInEth === '' || +withdrawalAmountInEth === 0) return null
+  const emptyInput = withdrawalAmountInEth === '' || +withdrawalAmountInEth === 0
+  const tradeNotPossible = priceImpactTooHigh || insufficientLiquidity
+  if (emptyInput || tradeNotPossible) return null
```
chrisling-dev, line 25: I think there'll be summary with estimated value for price impact too high state, lmkwyt

apps/frontend/core-dapp/src/stores/BalancerStore.ts
```diff
@@ -120,16 +137,28 @@ export class BalancerStore {
    * it to parse user inputs. To estimate the value of a ETH balance without
    * accounting for price impact, use {@link getEthAmountInWstEth}.
    */
-  quoteEthAmountInWstEth(ethAmount: BigNumber): Promise<BigNumber | undefined> {
-    const { wethAddress, wstEthAddress } = this
+  async quoteEthAmountInWstEth(ethAmount: BigNumber): Promise<Quote | undefined> {
+    const { wethAddress, wethDecimals, wstEthAddress } = this
```
chrisling-dev, line 141: would it be easier to just read decimalsNumber here? not a big deal tho, just trying to keep the file short and concise

```suggestion
    const { wethAddress, wstEthAddress } = this
    const { decimalsNumber } = this.root.wethStore
```

apps/frontend/core-dapp/src/stores/BalancerStore.ts
```diff
@@ -216,4 +250,28 @@ export class BalancerStore {
   static getTradeDeadline(): BigNumber {
     return BigNumber.from(getTradeDeadlineFromDate(Date.now()).getTime()).div(1000)
   }
+
+  private static calculatePriceImpact({
```
chrisling-dev, line 254: Would recommend to put this into a file in utils.ts because we'll probably need to reuse this function for open/close trade

apps/frontend/core-dapp/src/stores/DepositTradeHelperStore.ts
```diff
@@ -27,8 +27,16 @@ export class DepositTradeHelperStore extends ContractStore<RootStore, SupportedC
...
       }
 
-      const wstEthAfterSlippage =
-        this.root.advancedSettingsStore.getAmountAfterSlippage(amountInWstEth)
+      if (this.root.advancedSettingsStore.isPriceImpactTooHigh(amountInWstEth.priceImpact)) {
+        return {
+          success: false,
+          error: "Can't swap to wstETH: price impact too high. Try a smaller amount.",
+        }
+      }
```
chrisling-dev, line 35: We don't need this I reckon because:
- button is already disabled
- technically you could still place a trade, just with a bad price. So since our purpose with disabling button is just to provide a better UX, I feel like there's no need to also disable the underlying execution function

however don't think my suggestion will have any actual effect atm from UX perspective so feel free to disregard if you disagree

apps/frontend/core-dapp/src/stores/DepositTradeHelperStore.ts
```diff
@@ -27,8 +27,16 @@ export class DepositTradeHelperStore extends ContractStore<RootStore, SupportedC
...
+
+      const wstEthAfterSlippage = this.root.advancedSettingsStore.getAmountAfterSlippage(
+        amountInWstEth.value
+      )
```
chrisling-dev, line 39: curious why was passing amountInWstEth working before but we need to use amountInWstEth.value now?

chrisling-dev (changes requested):

apps/frontend/core-dapp/src/stores/DepositTradeHelperStore.ts
MeLlamoPablo: > button is already disabled

No, this is Deposit, not Withdraw. The button is not disabled because we don't request a quote from Balancer on deposits. Therefore, we need to stop the user if they're going to do a bad trade.

apps/frontend/core-dapp/src/stores/DepositTradeHelperStore.ts
MeLlamoPablo: The returned quote is now an object:

```typescript
type Quote = {
  priceImpact: number
  value: BigNumber
}
```

chrisling-dev: ahh gotcha, thanks!

apps/frontend/core-dapp/src/features/withdraw/WithdrawStore.ts
MeLlamoPablo: I'm a bit reluctant to keeping it on `queried`, because we rely on `queried` to know whether a trade is valid (like [here](https://github.com/prepo-io/prepo-monorepo/blob/617f4450f83ca8ea8b7821e3a86d7e118699df30/apps/frontend/core-dapp/src/features/withdraw/WithdrawStore.ts#L142) or [here](https://github.com/prepo-io/prepo-monorepo/blob/617f4450f83ca8ea8b7821e3a86d7e118699df30/apps/frontend/core-dapp/src/features/withdraw/WithdrawStore.ts#L154))

So I'll do the second suggestion instead and keep the value on `price-impact-too-high`. I'll also rename `queried` to `trade-valid` for more clarity.

chrisling-dev (approved):
LGTM

PR #1007, opened 2023-03-09
fix: separate token sender for deposits and withdrawals

ramenforbreakfast:
Before, I had originally designed the deployment process with one shared `TokenSender` in mind. This change accomodates multiple token senders with the following changes:

- Token senders are now meant to be hierarchically placed under a contract, since most likely we will be deploying multiple ones with different multipliers, possibly even price sources. This does not mean contracts cannot share them, it is just an organizational placement.

- Reorganized and renamed the `UniswapV3UintValue` contract. Originally it was placed as belonging to the singular TokenSender, but this was not a good decision in hindsight. Especially now that multiple TokenSenders will use the same oracle (since price manipulation will happen using different multipliers), it has been renamed to `UniswapV3OracleUintValue-PPO-USDC-ETH`, as a standalone item not belonging to any specific contract. The name was chosen to denote both the token pair sourcing, and what the `baseAmount` it is converting represents. In this case, it is a PPO price oracle, converting using a USDC/ETH price.

- Since we need to deploy multiple `TokenSenders`, it no longer makes sense to have a singular script to deploy the contract, since we need to automatically deploy multiple instances. I have figured out how to deploy a contract within a helper function and now the deployment process for peripheral contracts has been moved to `deployTokenSenderAndPeriphery` in `harnesses/prod.ts`. All deployments for peripheral contracts such as `AccountList`'s and the separate `TokenSender`'s have been moved here.

- `ConfigureTokenSignerViaSigner` was moved to a helper function so as to reduce code duplication, since we now need to configure 2 `TokenSenders`.

I understand this PR is quite large, but I did not see it feasible to separate these changes out given how entangled they were.


apps/smart-contracts/core/deploy/ConfigureStackViaSigner.ts
```diff
@@ -87,49 +60,60 @@ const deployFunction: DeployFunction = async function configureStackViaSigner(
...
   await core.configureDepositRecordViaSigner(
     signer,
     deploymentParameters.depositRecord.globalNetDepositCap,
     deploymentParameters.depositRecord.userDepositCap,
     deploymentParameters.depositRecord.allowedMsgSenders,
     deploymentParameters.depositRecord.bypasslist
   )
-  console.log('Configuring TokenSender via Signer...')
-  await core.configureTokenSenderViaSigner(
+  console.log('\nConfiguring ETH/USDC Oracle via Signer...')
```
XavierEkkel, line 110: nit:
```suggestion
  console.log('\nConfiguring USDC/ETH Oracle via Signer...')
```
since ETH is the quote asset

apps/smart-contracts/core/harnesses/prod.ts
```diff
@@ -58,27 +65,41 @@ export class ProdCore extends Base {
...
       DEPLOYMENT_NAMES.preWstETH.depositHook.name
     )
+    this.collateral.depositHook.tokenSender = await ethers.getContractOrNull(
+      DEPLOYMENT_NAMES.preWstETH.depositHook.tokenSender.name
+    )
+    if (this.collateral.depositHook.tokenSender) {
+      this.collateral.depositHook.tokenSender.allowedMsgSenders = await ethers.getContractOrNull(
+        DEPLOYMENT_NAMES.preWstETH.depositHook.tokenSender.allowedMsgSenders.name
+      )
+      this.collateral.depositHook.tokenSender.twapPrice = this.ppoUSDCtoETHOracle
```
XavierEkkel, line 82: if tokensender is non-null + set, then shouldn't it be made essential that these are also non-null?
otherwise deposithook will revert always until they are set

same for withdrawhook

XavierEkkel (approved):
LGTM after question

apps/smart-contracts/core/harnesses/prod.ts
ramenforbreakfast: We can, it really doesn't matter and it would be annoying to change. Essentially the convention right now is, contracts that will only have one instance of themselves are deployed first, then through the configuration script, repeated contracts that are assigned to others are then deployed.

If I made these enforced as non-null, I'd essentially have to deploy these upon the harnesses initialization, rather than post initialization.

The above issue doesn't occur also because the deposit hook isn't even connected to the collateral contract until the configuration script is run.

PR #1009, opened 2023-03-09
feat: dynamically initialize TokenSender contract for deposit and wit…

chrisling-dev:
…hdraw

https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=6b4403e1ecc748a48a63e32a1f35e012&p=019719da7f2740db9fb713c6f23bb849&pm=s

PR #1012, opened 2023-03-09
feat: don't show rewards if TokenSender doesn't have PPO

MeLlamoPablo:
https://www.notion.so/prepo/TokenSender-s-PPO-balance-check-4a16b38c980a4614a71191270c516c4b

Blocked by #1009


chrisling-dev (approved):
LGTM after mergint the fix I have in the blocking PR

PR #1014, opened 2023-03-10
feat: wrapAndDepositAndTrade with unit tests only

ramenforbreakfast:
**Updated as of 3/14/2023**

Since our `DepositTradeHelper` now revolves around a ETH => WstETH flow, there is not much purpose in keeping the original `depositAndTrade` and therefore, need for a `baseTokenPermit`. 

In addition, depositing is already done via `wrapAndDeposit`, so we can now just isolate the portion that takes the user's `Collateral` and trades with it into an internal function `_trade`. `_trade` is written in a way where it is agnostic since it will need to support bi-directional trading (collateral => any market, any market => collateral). 

I also decided to split out collateral permit and transferring into a more generic function `_permitAndTransfer` since we will need to process permits and transfers for both `Collateral` and `LongShortToken` for bi-directional trading.

Along with the implementation, is the reworked DepositTradeHelper unit tests, since `depositAndTrade` no longer exists, but we want to keep the majority of these unit tests. Comments are provided to explain some hacky things I do to get around the fact that the unit tests are not using an actual Balancer swap or `wstETH` token.


apps/smart-contracts/core/contracts/DepositTradeHelper.sol
```diff
@@ -37,66 +37,19 @@ contract DepositTradeHelper is
...
-
   function wrapAndDepositAndTrade(
+    address recipient,
+    OffChainBalancerParams calldata balancerParams,
     Permit calldata collateralPermit,
     OffChainTradeParams calldata tradeParams
-  ) external payable override nonReentrant {}
+  ) external payable override nonReentrant {
+    uint256 collateralAmount = _wrapAndDeposit(recipient, balancerParams);
+    _trade(collateralAmount, collateralPermit, tradeParams);
```
MeLlamoPablo, line 51: I wonder if we could skip the need for a `collateralPermit` by calling `_wrapAndDeposit` with `address(this)` and then doing `_collateral.transfer` instead of `_collateral.transferFrom`.

Would save a click from the user and maybe be more gas efficient? Then if `_trade` fails the whole function should revert so the user doesn't lose ETH, correct?

ramenforbreakfast: So @MeLlamoPablo the reason why we cannot do this is because we need to attribute deposits to the actual depositor due to user deposit caps. This creates a restriction on the recipient of the `Collateral`. If we were to do as you suggested, the `DepositTradeHelper` would be the one these deposits would be attributed to, creating inaccurate records in regards to how much each user has deposited.

MeLlamoPablo: Why doesn't `DepositHook` record the sender of the ETH, not the recipient? (which in most cases will be the same, I'm just wondering the practical difference)

ramenforbreakfast: Well we have to stick with one convention, you either tie the ownership of a deposit (note this does not necessarily mean the recipient, in our case it does) to the recipient or the funder (`msg.sender`). 

If you don't tie it to either, its impossible to be consistent on how you keep track of user deposits.

We don't record the funder, because the deposit record attributes a recorded deposit to the `recipient`. If we did it based on the funder, we would pass that instead. 

If we did it based on the funder, we wouldn't be able to use an intermediary contract since any call to `deposit` would need to be done from the user.

MeLlamoPablo: Thanks so much! That was very informative

ramenforbreakfast:
Update: While looking at the codebase answering Pablo's question, I did realize that there is a mistake in that we cannot just have the `Collateral` come from `msg.sender` like in the original `depositAndTrade`. This is because `wrapAndDepositAndTrade` needs the `Collateral` to come from the `recipient` attributed to those deposits.

I have now aligned the `Collateral` permit and the `transferFrom` source to be the `recipient` and not `msg.sender`.

I have also made the internal `_trade` function take in a `funder` and `recipient`, because for our standalone `trade` function, we would pass in `msg.sender` as the `funder`, instead of `recipient`, since the caller is assume to already have `Collateral` to perform the standalone trade.

XavierEkkel (approved):
contract changes LGTM

PR #1016, opened 2023-03-10
feat: cap the amount in CurrencyInput to 99k

MeLlamoPablo:
This PR adds a default cap to the user input in `CurrencyInput` of 99,999. If the user attempts to input something higher, it will revert to that.


chrisling-dev (approved):
LGTM

PR #1017, opened 2023-03-10
feat: terms wall with dummy terms of service

chrisling-dev:


apps/frontend/core-dapp/src/lib/rules.tsx
```diff
@@ -0,0 +1,18 @@
...
+  'I understand that this is a very important message that I should be aware of.',
+  {
+    id: 'rule_2',
+    content: (
+      <p>
+        I understand that this is a very <strong>important bolded message</strong> that I should be
+        aware of.
+      </p>
+    ),
+  },
```
MeLlamoPablo, line 11: I think this is unnecessary complexity. We could just make all rules an object with `id` and `content`. While the file would be longer, we would reduce the mental overhead and could remove the `typeof` check.

MeLlamoPablo (approved):
Awesome! I think the system looks great.

apps/frontend/core-dapp/src/lib/rules.tsx
chrisling-dev: my only problem with that is that I don't want to have to write so many boilerplate for each rule just to add a rule, while keeping the possibility of writing JSX while necessary (e.g. add link, bold text)

MeLlamoPablo: I don't think we should avoid boilerplate - it is not necessarily a bad thing. On the other hand, we should avoid creating complex systems that don't add value. For example, increasing complexity to avoid boilerplate doesn't add value if it doesn't "save" us from writing a lot of boilerplate. In this case, I don't expect the terms to change very often.

I think adding `id` and `content` is not really a lot of extra effort, but feel free to disregard since the PR is approved.

chrisling-dev: okay I'm sold, added changes

PR #1019, opened 2023-03-10
fix: crash in CurrencyInput

MeLlamoPablo:
This PR fixes a crash that happens when typing a space in the `CurrencyInput`. This is because `validateStringToBN` returns `true`, since the unary plus operator can handle spacds when converting a string to a number, but `BigNumber` will trhow an error when feeding it a space.


packages/prepo-utils/src/validateStringToBN.ts
```diff
@@ -1,5 +1,7 @@
 export const validateStringToBN = (input: string): boolean => {
   // empty string is treated as 0
   if (input === '') return true
+  // valid strings can't have spaces
+  if (input !== input.trim()) return false
```
chrisling-dev, line 5: simpler alternative solution IMO, but not blocker

```suggestion
  if (input.includes(' ')) return false
```

chrisling-dev (approved):
Nice catch!

packages/prepo-utils/src/validateStringToBN.ts
MeLlamoPablo: Disregarding this because `trim` removes more than just spaces. It removes any white space character and line breaks. I don't think that you can enter those in the input but it's good to be defensive.

PR #1020, opened 2023-03-12
refactor: replace static call of actual swap with quoting function

ramenforbreakfast:
When writing the integration tests for `DepositTradeHelper`, I decided to perform a small refactoring to simplify how we obtain uniswap quotes for testing verification. Previously, I did not do it this way, because at the time, I was hesitant to include another artifact just to make quoting easier, since we were already adding a lot of external artifacts to support the Uniswap periphery. In hindsight, I should have just done this from the start. This allows us to easily obtain uniswap quotes for our tests without actually needing the assets for the trade.


XavierEkkel (commented):
@ramenforbreakfast this needs rebasing

XavierEkkel (dismissed):
LGTM if tests pass

XavierEkkel (approved):

PR #1025, opened 2023-03-13
feat: some preliminary cleanup for early deposit launch

chrisling-dev:
https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=6b4403e1ecc748a48a63e32a1f35e012&p=f2a265d3c11d4a7f9bb552df81cf4628&pm=s


apps/frontend/core-dapp/src/pages/index.tsx
```diff
@@ -1,19 +1,17 @@
...
-      : {
-          destination: Routes.Deposit,
-          permanent: false,
-        },
+  redirect: isProduction()
+    ? undefined
+    : {
+        destination: Routes.Deposit,
+        permanent: false,
+      },
```
MeLlamoPablo, line 14: I think it's very confusing to use `isProduction` inside `getServerSideProps` because `isProduction` returns true if `window === undefined`, which means that it will always return true in server side code.

What was the reason for `isProduction` to check the host name instead of `NODE_ENV` again? I think we discussed this but I don't remember the conclusion.

In my opinion, checking the host name adds complexity because we have to handle server code separately.

In this instance, can we just always do the redirect and deal with it later? Also, I think this is backwards - on production we *do* want to redirect users from `/` to `/deposit` right?

MeLlamoPablo (approved):
Approving to avoid blocking, but make sure to address the comment.

apps/frontend/core-dapp/src/pages/index.tsx
chrisling-dev: oh no my res wasn't sent ytd,

I'm using isProduction check here because:
- in preview, NODE_ENV is set as "production" so we might not be able to test some of the behaviour
- for any production check we should only rely on one function `isProduction()`

HOWEVER, it seems like it's still not directing to /deposit in preview (works in localhost, and work as expected in production, but not in preview)

I'll update the logic in isProduction() to `process.env.NODE_ENV === 'production'` in a separate PR

So business logic wise this is correct, we don't want to redirect to Deposit yet until early deposit launch. So for now, if it's in production, it should stay on `/` and show the Coming Soon page, but in dev or preview it should go to `/deposit`

chrisling-dev: > What was the reason for isProduction to check the host name instead of NODE_ENV again? I think we discussed this but I don't remember the conclusion.

as for this, it was because we weren't sure what was the exact values Vercel used for NODE_ENV in different environment. But I've tested it with this PR so we should be able to do it in the next PR

MeLlamoPablo: > in preview, NODE_ENV is set as "production" so we might not be able to test some of the behaviour

We should be able to add preview-only variables on Vercel, and if we can't override `NODE_ENV` at least we'd be able to add something like `VERCEL_PREVIEW=true`

chrisling-dev: yea we should be able to modify NODE_ENV for preview, I'll work on this shortly

PR #1026, opened 2023-03-14
feat: allow external contracts to easily identify valid markets

ramenforbreakfast:
Our original plan involved basing fees on market payouts, but we are now opting for a simpler model of taking fees purely based on collateral input or output. However, I felt this was still a change that made sense and will extend this off of `main` rather than the `DepositTradeHelper` chain due to it no longer depending on this.

I have spent some time reworking the PrePOMarketFactory to simplify how we store the validity of the market and to make it easy for an external contract to know if it is a sanctioned contract. The old way of storing the salt hashes as a mapping isn't really useful to anybody and markets deployed can be inferred off-chain from events. All an external contract really needs is to know if this is a sanctioned market deployed from our factory.

There are also minor changes to the variable naming to conform to our new standard.


apps/smart-contracts/core/test/unit/PrePOMarketFactory.test.ts
```diff
@@ -174,6 +179,14 @@ describe('=> PrePOMarketFactory', () => {
       await expect(tx).revertedWith('shortToken address >= collateral')
     })
 
+    it('reverts if attempting to deploy duplicate market', async () => {
```
XavierEkkel, line 182:
```suggestion
    it('reverts if duplicate market', async () => {
```

XavierEkkel (approved):
LGTM after comments

PR #1029, opened 2023-03-14
feat: add PPO balance to BalanceBox

MeLlamoPablo:
Screenshot 2023-03-14 at 11 54 56

https://www.notion.so/prepo/Add-PPO-balance-to-the-balance-box-below-deposit-component-5ecd94b11b2944428e8f5847356f4694


apps/frontend/core-dapp/src/features/deposit/BalanceBox.tsx
```diff
@@ -103,6 +107,31 @@ const BalanceBox: React.FC<{ apr: unknown }> = ({ apr }) => {
...
+                display: 'flex',
+                alignItems: 'center',
+                justifyContent: 'center',
+                gap: spacingIncrement(4),
+              }}
+            >
+              <p>PPO Balance</p>
+            </div>
+          </Flex>
+          <Value>
```
chrisling-dev, line 124: could we add a text-align: right to this

apps/frontend/core-dapp/src/features/deposit/BalanceBox.tsx
```diff
@@ -58,10 +57,15 @@ const GreenColor = styled.span`
   font-weight: ${({ theme }): number => theme.fontWeight.semiBold};
 `
 
+const PurpleColor = styled.span`
+  color: #6264d9;
```
chrisling-dev, line 61: and this should use color from theme, and use light purple in dark mode

```suggestion
  color: ${({theme}): string => theme.color.primaryLight};
```

apps/frontend/core-dapp/src/features/deposit/BalanceBox.tsx
```diff
@@ -103,6 +107,31 @@ const BalanceBox: React.FC<{ apr: unknown }> = ({ apr }) => {
...
+          <Flex gap={8}>
+            <Icon name="ppo-logo" width="24" height="24" />
+            <div
+              style={{
+                display: 'flex',
+                alignItems: 'center',
+                justifyContent: 'center',
+                gap: spacingIncrement(4),
+              }}
+            >
```
chrisling-dev, line 120: You could just use the Flex component I think, not sure if it automatically centers everything by default tho

```suggestion
            <Flex
              // may not need alignItems and justifyContent if Flex already centers children
              justifyContent="center"
              alignItems="center"
              gap={4}
            >
```

chrisling-dev (changes requested):

apps/frontend/core-dapp/src/features/deposit/BalanceBox.tsx
MeLlamoPablo: Huh, so I couldn't find this color because on figma it's `#6264d9` and on our codebase it's `#6264d8` 😬

apps/frontend/core-dapp/src/features/deposit/BalanceBox.tsx
MeLlamoPablo: Done, but not sure what the point is if the parent has space-around?

chrisling-dev: this is so that when PPO amount text is long on galaxy fold, it doesn't look weird and the text is right align, so it looks 
```
  123.24K  |
         PPO  |
```

rather than
123.24k |
PPO       |

chrisling-dev (approved):
Thanks LGTM!

PR #1032, opened 2023-03-15
feat: add permit to LongShortToken

ramenforbreakfast:
This should have been done as a backlog item, but we need to ensure `LongShortToken` supports permits so the `DepositTradeHelper` can take a user's position tokens.

I saw no point in re-testing these functions (since they will be tested anyway in integration testing), but have added a test to ensure we are initializing the `DOMAIN_SEPARATOR` correctly. We did this for `PPO` but seemed to have missed these tests for `Collateral` as well.


XavierEkkel (approved):
contract LGTM

PR #1036, opened 2023-03-16
fix: prevent upgrade from being submitted if ABIs are the same

ramenforbreakfast:
While doing some preliminary work for planning how we would test the Arbitrage Bot, I wanted to solve this very annoying problem where OZ and `hardhat-deploy` don't really have mechanisms to determine whether or not to upgrade something even if the ABI hasn't changed. So I implemented a hack here where we can compare the ABI strings to determine whether to actually submit an upgrade, preventing us from submitting unnecessary upgrade proposals even when the contract hasn't changed.


XavierEkkel (approved):
LGTM

PR #1038, opened 2023-03-16
refactor: override balance in CollateralStore

MeLlamoPablo:
This PR makes a small refactor to `CollateralStore`. This store is a special case of `Erc20Store` in that we don't display the native units (wstETH), but computed units (ETH).

In order to use this properly in `TokensStore` we need to do additional logic, because by default `TokensStore` would call [`tokenBalanceFormat`](https://github.com/prepo-io/prepo-monorepo/blob/89bad28aa94e52621f86a9398ea59b59dbd246a6/apps/frontend/core-dapp/src/stores/TokensStore.ts#L83).

Since the vast majority of the times we display the collateral balance we do it in ETH, it makes sense to make it the default. So, this PR overrides `balanceOfSigner` in `Erc20Store` from `CollateralStore`. The neat part is that we don't need to implement our own `tokenBalanceFormat` this way.

In the rare case we need to access the underlying `wstETH` balance, we can call `balanceOfSignerInWstEth`.


apps/frontend/core-dapp/src/stores/CollateralStore.ts
```diff
@@ -145,17 +145,14 @@ export class CollateralStore extends Erc20Store {
...
+  get balanceOfSignerInWstEth(): BigNumber | undefined {
+    return super.balanceOfSigner
   }
 
-  get tokenBalanceFormatInEth(): string | undefined {
-    const { balanceOfSignerInEth } = this
-    const { decimalsNumber: wethDecimals } = this.root.wethStore
-    if (balanceOfSignerInEth === undefined || wethDecimals === undefined) return undefined
-    return utils.formatUnits(balanceOfSignerInEth, wethDecimals)
+  override get balanceOfSigner(): BigNumber | undefined {
```
chrisling-dev, line 152: Could we add a ts-doc here to explain that this is in wstETH? This approach looks good, my only concern is that we might be used to the fact that balanceOfSigner is for the actual balance of that token and accidentally misuse this.

chrisling-dev (approved):

PR #1039, opened 2023-03-16
fix: not being able to withdraw max

MeLlamoPablo:
This PR fixes a bug where there would still be `preWstEth` dust on the user's wallet after withdrawing MAX.

This is because when the user selects MAX, it selects 100% of its ETH balance which then gets converted to wstETH. During this conversion, due to price impact the amount we get is less than 100% of their wstETH.

The solution is to not convert the balance if MAX is selected, and instead just return the underlying wstETH balance.


chrisling-dev (approved):
LGTM, good to see that overriding balanceOfSigner with ETH value does make the code much more readable here, great approach 🔥

PR #1040, opened 2023-03-17
fix: multiple fixes, mainly around market deployments

ramenforbreakfast:
Made all `ConfigureStackViaSigner` fields undefined so as to not trigger transactions unnecessarily.

Also added a generic `hardhat-deploy` manual saver to assist with troubleshooting/fixes.


ramenforbreakfast:
Updated with additional minor fixes to ownership transfer, ensuring the nominee is not already nominated to prevent wasting gas on unnecessary txs. Added a TODO to refactor to allow it to also check if its owned _or_ nominated.

Also removed ArbitrageBroker from the role handover process since its not as straightforward.

ramenforbreakfast:
Moved `deployMarketHookAndPeriphery` to the renamed `deploy.ts` helper file, since there is no reason it should be contained within the `markets` task. Also added a `deployTokenSender` helper in there since this is an operation that will be necessary to perform often.

apps/smart-contracts/core/helpers/deploy.ts
```diff
@@ -0,0 +1,122 @@
...
+  }
+}
+
+export async function deployMarketHookAndPeriphery(
+  signer: SignerWithAddress,
+  hre: HardhatRuntimeEnvironment,
+  nameSuffix: string,
+  marketHookContractName: string,
+  marketAddress: string,
+  governanceAddress: string
```
XavierEkkel, line 63: pass in an optional `tokenSenderAddress` so that it can be set if it's already been deployed? (assuming the same one is used between markets for redeem)

apps/smart-contracts/core/helpers/deploy.ts
```diff
@@ -0,0 +1,122 @@
...
+): Promise<void> {
+  const { ethers } = hre
+  const hookName = `${nameSuffix}-${marketHookContractName}`
+  await deployNonUpgradeableContract(`${marketHookContractName}`, hookName, [], hre)
+  const hookAllowedMsgSendersName = `${nameSuffix}-${marketHookContractName}-AllowedMsgSenders`
+  await deployNonUpgradeableContract('AccountList', hookAllowedMsgSendersName, [], hre)
+  const hookAllowlistName = `${nameSuffix}-${marketHookContractName}-Allowlist`
```
XavierEkkel, line 70: consider passing in an optional `allowlistAddress` argument, so that it can be re-used across markets (no new deployment needed)

apps/smart-contracts/core/tasks/markets.ts
```diff
@@ -83,17 +46,25 @@ const prepareMarketPeriphery = async (
...
   )
-  await setContractIfNotAlreadySet(
+  await deployTokenSender(
+    signer,
+    hre,
+    `${nameSuffix}-RedeemHook-TokenSender`,
+    rewardTokenAddress,
+    18
+  )
+  const redeemHookTokenSender = await ethers.getContract(`${nameSuffix}-RedeemHook-TokenSender`)
```
XavierEkkel, line 63: if have `tokenSenderAddress` as an arg in `deployMarketHookAndPeriphery`, can call `deployTokenSender` first and then call `deployMarketHookAndPeriphery` after, which makes this `prepareMarketPeriphery` a bit neater (no value setting), and and makes `deployMarketHookAndPeriphery` more self-contained

XavierEkkel (approved):
LGTM after comments

apps/smart-contracts/core/helpers/deploy.ts
```diff
@@ -0,0 +1,122 @@
...
+    'getPrice',
+    'setPrice'
+  )
+  await setSingleValueIfNotAlreadySet(
+    signer,
+    tokenSender,
+    tokenSender.allowedMsgSenders.address,
+    'getAllowedMsgSenders',
+    'setAllowedMsgSenders'
+  )
```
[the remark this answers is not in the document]
ramenforbreakfast: No, because if it is not defined, something might have gone wrong and we want the script to fail so we are alerted.

PR #1059, opened 2023-03-17
feat: add arbitrum market details 

MeLlamoPablo:
Screenshot 2023-03-18 at 13 27 15

This PR adds:

* The Arbitrum market address + token addresses + pool addresses
* The Arbitrum market details (name, logo...)
  * I'm using the same logo as the network logo, but we may want to change it to a different, nicer one.
* Updates the CurrencyInput on CloseTrade to use `preETH`

The chart on the left is not working but I believe it is subgraph related right @chrisling-dev ?


chrisling-dev (commented):
LGTM code wise, approving shortly after checking preview with latest pool

chrisling-dev:
> The chart on the left is not working but I believe it is subgraph related right @chrisling-dev ?

Yeap, I'll check if uniswap has a subgraph on arbitrum that has the same interface

chrisling-dev (approved):
LGTM 🔥

PR #1070, opened 2023-03-19
feat: opentrade eth conversion

chrisling-dev:


MeLlamoPablo (approved):
LGTM! A lot of logic is similar to the `WithdrawStore` so maybe after launch we can try to clean it up and extract it.

chrisling-dev:
> LGTM! A lot of logic is similar to the `WithdrawStore` so maybe after launch we can try to clean it up and extract it.

ye ye I agree, definitely need some clean up everywhere lol

PR #1083, opened 2023-03-19
feat: implement launch checklist

MeLlamoPablo:
In addition to the items [here](https://www.notion.so/prepo/Checklist-for-transitioning-from-Early-Deposit-to-Platform-Launch-dcc83de00a2948c5b46a54c9c2fb9ee2), this PR also redirects `/deposit` -> `/portfolio/deposit` and `/withdraw` -> `/portfolio/withdraw`. This is necessary for the navbar "current page highlight" to work properly.

This PR introduces a `PLATFORM_LAUNCHED` flagin `apps/frontend/core-dapp/src/platform-launch.ts`. When we want to launch, the whole contents of that file should be replaced with:

```typescript
export const PLATFORM_LAUNCHED = true
```


XavierEkkel (approved):
LGTM after:
-ensuring that / goes to /trade (when PLATFORM_LAUNCHED = true)
-ensuring that clicking prePO logo also goes to / (not /deposit)
-hiding nav bar during terms overlay

PR #1085, opened 2023-03-20
fix: close position price conversion calculation

chrisling-dev:


apps/frontend/core-dapp/src/stores/SwapStore.ts
```diff
@@ -123,39 +123,36 @@ export class SwapStore {
...
+    fromAddress,
+    toAddress,
+    fee,
+  }: QuoteExactInputProps): Promise<
+    { cachedInAmount: BigNumber; output: BigNumber } | undefined
+  > => {
+    if (amountBN.eq(0)) return { cachedInAmount: amountBN, output: BigNumber.from(0) }
+
+    const quoterContract = new ethers.Contract(
+      UNISWAP_QUOTER_ADDRESS.mainnet ?? '', // all uniswap contracts has same address on all chains
```
XavierEkkel, line 137: not true necessarily, see https://docs.uniswap.org/contracts/v3/reference/deployments
`Integrators should no longer assume that they are deployed to the same addresses across chains and be extremely careful to confirm mappings below.`
(however it is the same for mainnet+arbitrum+testnets)

PR #1088, opened 2023-03-20
fix: price out of bound

chrisling-dev:
handle when long/short token price is out of bound


MeLlamoPablo (approved):
LGTM. Note that this will conflict with #1087, but let's prioritize this one

PR #1089, opened 2023-03-20
chore: update Uniswap router address and ABI to v3 [ON HOLD]

MeLlamoPablo:


MeLlamoPablo:
Seems like the ABIs are not compatible, so putting this on hold for after launch.

PR #1093, opened 2023-03-20
feat: minor copy things

chrisling-dev:
https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=6b4403e1ecc748a48a63e32a1f35e012&p=0eab35bd8fd344bb91fef93e6aa8e4c8&pm=s

and some minor adjustments on portfolio page


MeLlamoPablo (approved):
The tooltip change may make them less discoverable on mobile, so I don't entirely agree (but I think that it makes the UI look clean and nicer)

But we can address that in the future

chrisling-dev:
> The tooltip change may make them less discoverable on mobile, so I don't entirely agree (but I think that it makes the UI look clean and nicer)
> 
> But we can address that in the future

yea it was to address the issue where balances don't quite fit on smaller screens and it looked too compact, agree with your points tho, but will address that later

apps/frontend/core-dapp/src/features/trade/DepositMessage.tsx
```diff
@@ -0,0 +1,29 @@
...
+  padding: ${spacingIncrement(16)};
+`
+
+const DepositMessage: React.FC = () => (
+  <Wrapper>
+    <Text>
+      You need to deposit before trading. <Link href={Routes.Deposit}>Deposit now &rarr;</Link>
+    </Text>
+  </Wrapper>
+)
```
MeLlamoPablo, line 27: Actually, this message needs a conditional:

Screenshot 2023-03-20 at 14 52 42

MeLlamoPablo (changes requested):

PR #1096, opened 2023-03-20
feat: add PPO balance to portfolio

MeLlamoPablo:
Screenshot 2023-03-20 at 15 45 42


apps/frontend/core-dapp/src/features/portfolio/PpoBalance.tsx
```diff
@@ -0,0 +1,61 @@
+import React from 'react'
+import styled from 'styled-components'
+import { observer } from 'mobx-react-lite'
+import { Flex, spacingIncrement } from 'prepo-ui'
+import Skeleton from 'react-loading-skeleton'
```
chrisling-dev, line 5: should use the custo Skeleton component we have so it looks good on dark mode

apps/frontend/core-dapp/src/features/portfolio/Portfolio.tsx
```diff
@@ -22,7 +23,10 @@ const Portfolio: React.FC<{ apr: string }> = ({ apr }) => (
       desktop: 'fit-content',
     }}
   >
-    <PortfolioValue apr={apr} />
+    <Flex flexDirection="column" gap={10}>
```
chrisling-dev, line 26: could we add a min-width 264px?

chrisling-dev (changes requested):
LGTM after comments

apps/frontend/core-dapp/src/features/portfolio/Portfolio.tsx
MeLlamoPablo: What would that be useful for? The width is defined on the grid template.

MeLlamoPablo: Maybe when you were looking on the preview it was >264px, but I fixed it and now it's exactly 264 on desktop.

chrisling-dev (approved):
LGTM

PR #1097, opened 2023-03-20
feat: add observer with observer

chrisling-dev:


MeLlamoPablo (approved):
Wonder why did it work for me when I made the change...

PR #1106, opened 2023-03-21
feat: show fdv test with token price

chrisling-dev:


apps/frontend/core-dapp/src/features/trade/SlideUpRow.tsx
```diff
@@ -134,11 +135,18 @@ const SlideUpRow: React.FC<Props> = ({
         minWidth="0px"
       >
         <NameWrapper>
-          <Name>{position.market.name}</Name>
+          <Name>
+            {position.market.name}
+            {position.market.symbol !== undefined && (
+              <span style={{ fontSize: '16px' }}> (${position.market.symbol})</span>
+            )}
```
MeLlamoPablo, line 142: The font size doesn't look too different. Maybe we can simplify and have it in the name?

chrisling-dev: removed this

XavierEkkel (approved):
LGTM

PR #1107, opened 2023-03-22
[DO NOT MERGE] fix: simplify turbo pipeline with core dapp changes

chrisling-dev:


MeLlamoPablo (approved):
Awesome! The more native features we use, the better 🥳

PR #1116, opened 2023-03-22
feat: show deposit limit on deposit and trade

MeLlamoPablo:
https://user-images.githubusercontent.com/11708035/226987528-073c299f-f7c7-499c-a4ab-c4c93b3f9c5e.mp4

This PR refactors the deposit limit logic out of `DepositStore` so it is also reusable on `TradeStore` and displays the `DepositWarning` on "deposit and trade".


chrisling-dev (changes requested):
Haven't been able to test the flow out because deposit limit is reached, but noticed an obvious bug: the selected payment token shouldn't affect closing, closing should only be in preETH, right

apps/frontend/core-dapp/src/features/trade/TradeStore.ts
```diff
@@ -901,6 +903,19 @@ export class TradeStore {
     return this.root.depositStore.calculateDepositRewardValue(ppoRewardForOpenTrade)
   }
 
+  get depositLimit(): DepositLimit {
+    if (this.action !== 'open' || this.paymentToken.type !== 'native') {
```
chrisling-dev, line 907: do we need this check? Don't think it's relevant here because withdraw flow doesnt use deposit limit
```suggestion
    if (this.paymentToken.type !== 'native') {
```

apps/frontend/core-dapp/src/features/deposit/DepositLimitWarning.tsx
```diff
@@ -104,13 +102,16 @@ const DepositWarning: React.FC = () => {
...
-        </Box>
-      )
+      if (showWstEthInfo)
+        return (
+          <Box>
+            <p>Your ETH will be deposited as wstETH to earn you yield.</p>
+          </Box>
+        )
+
+      return null
```
chrisling-dev, line 112: Maybe this is cleaner?

```suggestion
        return showWstEthInfo ? (
          <Box>
            <p>Your ETH will be deposited as wstETH to earn you yield.</p>
          </Box>
          ) : null
```

chrisling-dev (changes requested):
minor comments, will test again when limit is increased, can't test anything now

apps/frontend/core-dapp/src/features/trade/TradeStore.ts
MeLlamoPablo: Same logic as with [this](https://github.com/prepo-io/prepo-monorepo/pull/1115#discussion_r1147258469). I'm happy to remove it, but personally it would be helpful for me.

MeLlamoPablo:
> Haven't been able to test the flow out because deposit limit is reached, but noticed an obvious bug: the selected payment token shouldn't affect closing, closing should only be in preETH, right

Why do you say that the payment token affects closing?

chrisling-dev:
> > Haven't been able to test the flow out because deposit limit is reached, but noticed an obvious bug: the selected payment token shouldn't affect closing, closing should only be in preETH, right
> 
> Why do you say that the payment token affects closing?

When testing locally (preview now uses NODE_ENV which has the value "production"), if I choose ETH, and go to Close tab, it will show as ETH in close tab. Similarly if I choose preETH and go to Close tab, it shows as preETH in close tab. Selecting a paymentToken shouldn't ever affect payment token in closing

chrisling-dev (dismissed):

MeLlamoPablo:
Superseded by #1142 and #1143

PR #1117, opened 2023-03-23
feat: chart by market and direction

chrisling-dev:
https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=6b4403e1ecc748a48a63e32a1f35e012&p=27155d9957344c97ad2d5e2f61f017d1&pm=s


apps/frontend/core-dapp/src/stores/entities/MarketEntity.ts
```diff
@@ -382,10 +381,10 @@ export class MarketEntity
     )
   }
 
-  get historicalData(): MarketHistoryData[] | undefined {
+  get historicalData(): SeparatedMarketHistoryData | undefined {
     const { data, type } = this.getHistoricalData(this.historicalDataDateRange)
     if (data === undefined) return undefined
-    return syncLatestData(data, type, this.getLatestPoolsData(type))
+    return syncLatestDatas(data, type, this.getLatestPoolsData(type))
```
MeLlamoPablo, line 387: Nit: the plural of "data" is "data". To avoid ambiguity, the singular can also be "datum".

Personally when this can lead to confusion I usually avoid the word "data" in favor of something more specific. In this case it could be "data point"

apps/frontend/core-dapp/src/utils/market-utils.ts
```diff
@@ -327,30 +303,37 @@ export const normalizePoolsData = (poolsData?: PoolsData): NormalizedPoolsData |
   }
 }
 
-export const syncLatestData = (
-  historicalDatas: MarketHistoryData[],
-  type: HistoricalDataQueryType,
-  latestDatas?: MarketHistoryData[]
+const syncLatestData = (
```
MeLlamoPablo, line 306: What does this function do? It is unclear to me from the name

chrisling-dev: Sync with latest data on subgraph. Uniswap's subgraph records historical data on hourly or daily basis. The latest "tick" can change depending on trading activity.

The strategy:
- fetch all historical data and cache them, so we don't have to keep querying the entire list of data ticks, which can be massive
- on every block, query the latest tick (as you can see in screenshot below, it queries only the first - latest tick, so the network request is super lightweight)
- sync new data to our cached data
- the sync function basically helps update the cached data by checking the timestamp of latest tick from subgraph against the latest tick in our cached data, it modifies the latest tick in cached data if latest tick from subgraph has the same timestamp, else it adds a new tick to our cached data

Screenshot 2023-03-24 at 10 47 30 AM

In other words, it syncs the latest tick from subgraph to our cached data such that, our cached data will always look like it just got fetched from subgraph

apps/frontend/core-dapp/src/stores/entities/MarketEntity.ts
chrisling-dev: renamed to syncPoolsData

apps/frontend/core-dapp/src/utils/market-utils.ts
MeLlamoPablo: Thank you! That's very helpful

MeLlamoPablo (approved):
LGTM after fixing CI

PR #1119, opened 2023-03-23
feat: add support for permit and swap (via Uniswap)

MeLlamoPablo:
https://user-images.githubusercontent.com/11708035/227313538-34cca5f6-fd26-40aa-8471-bed4122c968f.mp4

This PR adds support for approving tokens via [ERC2612 permits](https://eips.ethereum.org/EIPS/eip-2612) for their usage in Uniswap. Instead of calling the `exactInput` function in `SwapRouter`, it calls its `multicall` function which then calls `selfPermit` + `exactInput`.

Since all of our interactions via Uniswap involve using tokens that we control (and all of them implement ERC2612), we can also remove all approval-related logic. We won't need to approve tokens anymore.

Note that I researched the new [Universal Router + Permit2](https://blog.uniswap.org/permit2-and-universal-router) contracts introduced by Uniswap, which are the ones they use on their UI. However, they don't work for us, because as far as I know `Permit2` aims to supersede `ERC2612`, not complement it. Which means that the users would still need to approve their tokens at least once with `Permit2`. I haven't seen a way to make `ERC2612` signatures work via `Permit2`.


MeLlamoPablo:
Not reopening this because we will use the trade functions in `DepositTradeHelper`, so this isn't relevant anymore.

PR #1127, opened 2023-03-24
feat: handle dust in deposit cap

MeLlamoPablo:
image


apps/frontend/core-dapp/src/utils/balance-limits.ts
```diff
@@ -24,32 +26,28 @@ export function getBalanceLimitInfo({
...
-    return { status: 'loading' }
-  }
-
   let status: 'already-exceeded' | 'not-exceeded' | 'exceeded-after-transfer'
 
-  if (currentAmount.gte(cap)) {
+  // To calculate the cap, we subtract a tiny amount of ETH to account for dust.
+  // Due to slippage, it is unlikely that users will deposit the exact remaining
+  // amount. They will typically leave room for dust, therefore the cap will
+  // never technically be exceeded.
```
XavierEkkel, line 34: nit (to reduce confusion):
```suggestion
  // Due to slippage, it is unlikely that users will deposit the exact remaining
  // amount. They will typically leave room for dust, therefore the cap will
  // never technically be exceeded. Therefore, to calculate the cap, we
  // subtract a tiny amount of ETH to account for dust.
```

XavierEkkel (approved):
LGTM after nit

XavierEkkel (approved):

PR #1128, opened 2023-03-24
feat: handle ETH <-> wstETH price impact on open trade

MeLlamoPablo:
This PR handles the high price impact on balancer wstETH -> ETH pool on the open trade page (which can happen via deposit + trade).

High price impact can happen due to two factors: high trade size or low liquidity. It is unlikely that the former case would happen here because a high trade size, because a trade with a size big enough to cause high price impact on the current balance pool will also likely bring our `preWstEth/longShortToken` pools outside bounds, which we already prevent on the UI. However, we can't trust that the balancer pool will always stay liquid.

There is one case that this PR doesn't cover: high price impact on the `preWstEth/longShortToken` pools (via Uniswap). I believe we don't need to cover this because we control most of the liquidity and have deposit limits. Therefore:

* A high price impact due to a big trade size is unlikely because we have deposit limits in place.
* A high price impact due to low liquidity is unlikely because we will ensure that there is always enough liquidity.


chrisling-dev (dismissed):
LGTM after addressing the nit I sent to you via TG

apps/frontend/core-dapp/src/features/trade/open-trade/OpenTradeButton.tsx
```diff
@@ -24,7 +24,7 @@ const OpenTradeButton: React.FC = () => {
 
     if (selectedMarket.resolved) return 'Market Closed'
 
-    if (withinBounds === false) return 'Unprofitable Trade'
+    if (unprofitable === true) return 'Unprofitable Trade'
```
chrisling-dev, line 27:
```suggestion
    if (unprofitable) return 'Unprofitable Trade'
```

chrisling-dev (changes requested):
LGTM after minor nit

apps/frontend/core-dapp/src/features/trade/TradeStore.ts
```diff
@@ -112,16 +113,11 @@ export class TradeStore {
...
         if (!wstEthQuote) return
 
-        if (this.root.advancedSettingsStore.isPriceImpactTooHigh(wstEthQuote.priceImpact)) {
-          // TODO handle this case
-          return
-        }
-
         runInAction(() => {
-          this.openTradeIntermediateWstEthOut = wstEthQuote.value
+          console.log(wstEthQuote)
```
chrisling-dev, line 119: should this be removed?

apps/frontend/core-dapp/src/features/trade/TradeStore.ts
```diff
@@ -553,6 +548,20 @@ export class TradeStore {
     return depositAmountAfterSlippage.sub(depositFeeAmount)
   }
 
+  private get priceImpactTooHigh(): boolean | undefined {
+    if (this.paymentToken.type !== 'native') return false
```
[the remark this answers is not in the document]
chrisling-dev: I think that should be in separate PR, added a ticket here with details: https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=6b4403e1ecc748a48a63e32a1f35e012&p=a69f0f4745624060bcb40c40ab860916&pm=s

apps/frontend/core-dapp/src/features/trade/open-trade/OpenTradeSummary.tsx
```diff
@@ -31,7 +32,8 @@ const OpenTradeSummary: React.FC = () => {
   const renderValueUI = (): React.ReactNode => {
     if (loading) return <Skeleton height="22px" width="64px" />
     const unitPrice = selectedMarket.getUnitPriceString(tradingValuation)
-    if (!withinBounds) return <RedText>Unprofitable</RedText>
+    if (withinBounds === false) return <RedText>Unprofitable Trade</RedText>
+    if (ethWstEthPriceImpactTooHigh) return <RedText>Insufficient wstETH Liquidity</RedText>
```
chrisling-dev, line 36: Are these changes meant to be here? I'm pretty sure "Insufficient wstETH Liquidity" won't fit in the transaction summary, especially on smaller mobile screens. Should these changes be in OpenTradeButton instead of here?

chrisling-dev (changes requested):
LGTM after comments

apps/frontend/core-dapp/src/features/trade/TradeStore.ts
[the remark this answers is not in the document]
chrisling-dev: I think that should be in separate PR, added a ticket here with details: https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=6b4403e1ecc748a48a63e32a1f35e012&p=a69f0f4745624060bcb40c40ab860916&pm=s

apps/frontend/core-dapp/src/features/trade/TradeStore.ts
MeLlamoPablo: No, good catch, thanks!

apps/frontend/core-dapp/src/features/trade/open-trade/OpenTradeSummary.tsx
```diff
@@ -31,7 +32,8 @@ const OpenTradeSummary: React.FC = () => {
   const renderValueUI = (): React.ReactNode => {
     if (loading) return <Skeleton height="22px" width="64px" />
     const unitPrice = selectedMarket.getUnitPriceString(tradingValuation)
-    if (!withinBounds) return <RedText>Unprofitable</RedText>
+    if (withinBounds === false || ethWstEthPriceImpactTooHigh)
+      return <RedText>Unprofitable Trade</RedText>
```
chrisling-dev, line 36: Pretty sure it should only say Unprofitable in the summary

```suggestion
    if (withinBounds === false || ethWstEthPriceImpactTooHigh)
      return <RedText>Unprofitable</RedText>
```

chrisling-dev (changes requested):
LGTM after comment

chrisling-dev (approved):
LGTM

PR #1031, opened 2023-03-15
feat: add fee transfer logic and tests to wrapAndDepositAndTrade

ramenforbreakfast:
To save time, I did not see much purpose in doing both integration and unit tests for this as this is a very straightforward add-on and only added integration tests to ensure fees and rebates are sent to the appropriate accounts and in the right amounts.

I also plan to ensure fees and rebates are sent inside the standalone trade functions to ensure they are using the right fee percentage.

Like the other internal functions defined, the collateral fee capture was moved into an internal function since it will be used multiple times as well as placed at different points in a transaction depending on the direction of the trade (Collateral => Position, Position => Collateral).

PR #1132, opened 2023-03-27
feat: tradeForPosition implementation + tests

ramenforbreakfast:
Updated to include changes + tests around the fee transfer logic changes introduced in PR 1021 and 1031.


XavierEkkel (approved):
LGTM after similar test comments as in https://github.com/prepo-io/prepo-monorepo/pull/1031

PR #1134, opened 2023-03-27
feat: tradeForCollateral implementation and tests

ramenforbreakfast:
So while this function mostly just uses the existing internal functions, there are some quirks to the flow for LongShort => Collateral:

- Since there will be multiple markets, and multiple Long & Short tokens, we need to give the SwapRouter approval to take those Long and Short tokens from us, in addition to the existing approval flow of taking the tokens from the user. We cannot just tell the SwapRouter to take tokens directly from the user since swaps must be funded by the msg.sender.

- I debated whether to have a more complex system where we check to ensure the validity of long and short tokens by forcing the caller to pass in a market, and then we check that market's validity. But, since we don't really need to know the market (because we aren't using the floor payout to determine the rebate), and the trade will implicitly only work if there is an existing Collateral/LongShort pool, I opted for a simpler method of just forcing the very first caller to give the approval and check if it's already done so subsequent callers don't need to.

- Because we need to capture the Collateral fee post-trade, we cannot directly send the output of the swap to the recipient, we must send the Collateral to the contract first, capture the fee, and then transfer it to the recipient.

- You might be wondering why the approval/permit tests here are slightly different. This is because Long and Short tokens need to be deterministic in their deployment, we do not mock them. This means I cannot observe their permit call. I also found out that you cannot assert that an event was not called with specific arguments, you can only assert that the event wasn't called at all. Therefore, I have opted for just comparing allowances which I think is satisfactory given these are unlimited approvals.


ramenforbreakfast:
Reduce the total collateral deployed by governance initially by 1200 => 120 because it was causing issues with later tests where the initial supplying of `wstETH` had exhausted the pool to the point where I could not get enough `wstETH` to mint even 2 ethers of positions for a user.

apps/smart-contracts/core/test/integration/DepositTradeHelper.test.ts
```diff
@@ -973,6 +974,463 @@ describe('=> DepositTradeHelper', () => {
...
+          sqrtPriceLimitX96: 0,
+        }
+
+        /**
+         * Pass in a JUNK_PERMIT, since we do not check if a user already
+         * has given approval on-chain, we expect the FE to simply pass in
+         * a JUNK_PERMIT to tell us whether to skip approvals. With the
+         * allowance between the SwapRouter and the DepositTradeHelper, we
+         * must detect whether to perform an approval on-chain.
+         */
```
XavierEkkel, line 1180: could you add a test that if a non-junk permit is passed in it will be processed even if it already has the necessary approval? then you can move this comment there (where it will be most relevant)

apps/smart-contracts/core/test/integration/DepositTradeHelper.test.ts
```diff
@@ -973,6 +974,463 @@ describe('=> DepositTradeHelper', () => {
...
+            permitForFunder,
+            uniswapParamsForTrade
+          )
+
+        expect(await core.markets[TEST_NAME_SUFFIX].longToken.balanceOf(funder.address)).eq(
+          funderPositionTokenBefore.sub(TEST_POSITION_AMOUNT)
+        )
+        expect(await core.markets[TEST_NAME_SUFFIX].longToken.balanceOf(recipient.address)).eq(
+          recipientPositionTokenBefore
+        )
```
XavierEkkel, line 1215: this is confusing because funder is supposed to be the recipient, but here it's treating them as different

please do some renaming so that this makes sense

same for other recipient = funder tests

XavierEkkel (approved):
LGTM after comments and comments from https://github.com/prepo-io/prepo-monorepo/pull/1031

PR #1139, opened 2023-03-28
refactor: small improvements to TradeStore

MeLlamoPablo:
Continuation of #1130 - GitHub won't let me re-open that one


apps/frontend/core-dapp/src/stores/TokensStore.ts
```diff
@@ -78,6 +78,16 @@ export class TokensStore {
     return token.erc20.tokenBalanceFormat
   }
 
+  static parseUnits(token: Token, units: string | undefined): BigNumber | undefined {
```
chrisling-dev, line 81: I think `units` is very confusing since it's usually used for values like `gwei`, `wei` (e.g.  https://eth-converter.com/). The right name for this variable should be `amount` or `value`

```suggestion
  static parseUnits(token: Token, amount: string | undefined): BigNumber | undefined {
```

chrisling-dev (changes requested):

chrisling-dev (approved):
LGTM

PR #1141, opened 2023-03-29
feat: shared AllowedMsgSenders and allowlist

ramenforbreakfast:
To accomodate the varied optional parameters for each configuration function that you requested from the last PR and I agreed were a definite necessary change, I went with the convention of a `options` object that is passed in for optional parameters. This seems to be common TS convention to avoid the problem of having multiple positional optional parameters.

I seem to have forgotten to pass along the common shared optional parameters with both hooks, to the redeem hook configurator.

These optional parameters allow for the passing in of an existing `allowlist` and `allowedMsgSenders`.


XavierEkkel (approved):
LGTM

PR #1142, opened 2023-03-29
refactor: extract deposit limit logic

MeLlamoPablo:
This PR extracts the deposit logic from `DepositStore` so that it can be reused by other stores, and decouples the `DepositLimit` component from the mobx state so that each caller can pass their own deposit limits.


apps/frontend/core-dapp/src/features/deposit/DepositStore.ts
```diff
@@ -1,25 +1,12 @@
...
-      amountEth: string
-      capEth: string
-      remainingEth: string
-      status: 'already-exceeded' | 'exceeded-after-transfer'
-      type: 'user-limit' | 'global-limit'
-    }
-
 type ExceededAfterTransfer = BalanceLimitInfo & { status: 'exceeded-after-transfer' }
 
 function isExceededAfterTransfer(
```
chrisling-dev, line 12: should delete this code

chrisling-dev (approved):
LGTM after comment

PR #1143, opened 2023-03-29
feat: use deposit limit logic for deposit and trade

MeLlamoPablo:
Blocked by #1142

https://user-images.githubusercontent.com/11708035/226987528-073c299f-f7c7-499c-a4ab-c4c93b3f9c5e.mp4

This PR adds the deposit limit logic to `TradeStore` and shows the `DepositLimitWarning` on `OpenTrade`


apps/frontend/core-dapp/src/features/trade/TradeStore.ts
```diff
@@ -902,6 +904,19 @@ export class TradeStore {
...
+    if (this.action !== 'open' || this.paymentToken.type !== 'native') {
+      return { status: 'not-exceeded' }
+    }
+
+    return getDepositLimit({
+      connected: this.root.web3Store.connected,
+      depositAmount: this.openTradeAmountBN,
+      depositRecordStore: this.root.depositRecordStore,
+      isNetworkSupported: this.root.web3Store.isNetworkSupported,
+    })
```
chrisling-dev, line 917: one minor suggestion I have for this is to just make a util function in depositRecordStore which already takes care of `connected, limits, and isNetworkSupported`, and we just do `depositRecordStore.getDepositLimitInfo(this.openTradeAmountBN)`

but that can be in the previous PR or separate PR

apps/frontend/core-dapp/src/features/trade/TradeStore.ts
```diff
@@ -814,7 +815,7 @@ export class TradeStore {
     })
   }
 
-  get tradeDisabled(): boolean {
+  get openTradeDisabled(): boolean {
```
chrisling-dev, line 818: 👍

chrisling-dev (approved):
LGTM, minor suggestion irrelevant to this PR

apps/frontend/core-dapp/src/features/trade/TradeStore.ts
MeLlamoPablo: I think this is fine as it is now. This would add an intermediate step and I don't think it has a clear benefit.

The current way:

```mermaid
graph LR;
    A["TradeStore: depositLimit"]-->C["utils: getDepositLimit"];
    B["DepositStore: depositLimit"]-->C["utils: getDepositLimit"];
```

The proposed way:


```mermaid
graph LR;
    A["TradeStore: depositLimit"]-->D["DepositRecordStore: getDepositLimit"];
    B["DepositStore: depositLimit"]-->D["DepositRecordStore: getDepositLimit"];
    D["DepositRecordStore: getDepositLimit"]-->C["utils: getDepositLimit"];
```

This adds an intermediate step which makes the code harder to reason about and only saves us 3 lines of code on `TradeStore` and 3 on `DepositStore`. If we were calling `utils: getDepositLimit` from a lot of places I would agree, but in this case I don't think the change is necessary.

PR #1145, opened 2023-03-29
refactor: extract open trade fees

MeLlamoPablo:
This PR extracts all fee-related logic from `TradeStore` into `open-trade-utils.ts`

This is beneficial because:

* It is easier to test, as the utility functions are self-contained with less dependencies than `TradeStore`, so we can also write self-contained, single purpose tests.
* The internal values (like `amountAfterDepositSlippageInWstEth`) don't need to be exposed to `TradeStore`, so there's less pollution on the `TradeStore` namespace, which will reduce the mental load of working with this file.


apps/frontend/core-dapp/src/stores/DepositTradeHelperStore.ts
```diff
@@ -20,7 +21,10 @@ const getPermitDeadlineFromDate = addDays(1)
 const getPermitDeadline = (): number =>
   Math.floor(getPermitDeadlineFromDate(Date.now()).getTime() / 1000)
 
-export class DepositTradeHelperStore extends ContractStore<RootStore, SupportedContracts> {
+export class DepositTradeHelperStore
+  extends ContractStore<RootStore, SupportedContracts>
+  implements DepositAndTradeFeesHolder
```
chrisling-dev, line 26:
```suggestion
export class DepositTradeHelperStore extends ContractStore<RootStore, SupportedContracts>
```

apps/frontend/core-dapp/src/stores/CollateralStore.ts
```diff
@@ -23,7 +24,7 @@ type Withdraw = CollateralAbi['functions']['withdraw']
 const TOKEN_SYMBOL = 'preETH'
 const TOKEN_DECIMALS = 18
 
-export class CollateralStore extends Erc20Store {
+export class CollateralStore extends Erc20Store implements DepositFeeHolder {
```
chrisling-dev, line 27: ditto

```suggestion
export class CollateralStore extends Erc20Store {
```

apps/frontend/core-dapp/src/features/trade/open-trade-utils.ts
```diff
@@ -0,0 +1,215 @@
+import { BigNumber } from 'ethers'
+
+type Props = {
+  collateralStore: {
+    balanceOfSignerInWstEth: BigNumber | undefined
+    depositFee: BigNumber | undefined
+    percentDenominator: BigNumber | undefined
+  }
```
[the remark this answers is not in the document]
MeLlamoPablo: This is not relevant after ee7b042cb437d42dddd9d4c87c2397f37c7ae863 because `CollateralStore` and `DepositTradeHelperStore` depend on the new `DepositFeeHolder` and `DepositAndTradeFeesHolder` types, respectively.

I did this so that the IDE can properly track usages of this method. Otherwise, my IDE thinks `depositAndTrade` fee is not used anywhere in the codebase and suggests to remove it.

apps/frontend/core-dapp/src/stores/DepositTradeHelperStore.ts
MeLlamoPablo: See [this comment](https://github.com/prepo-io/prepo-monorepo/pull/1145#discussion_r1157138474) as for why are they necessary

chrisling-dev (changes requested):

apps/frontend/core-dapp/src/features/trade/open-trade-utils.ts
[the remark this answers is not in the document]
chrisling-dev: in which file does your IDE thinks depositAndTradeFee isn't used?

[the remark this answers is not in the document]
chrisling-dev: thanks for the video explanation, well technically we don't even need Pick, we could just do 

```
collateralStore: CollateralStore
depositTradeHelperStore: DepositTradeHelperStore
```

since typescript is just a "sugarcoat" on top of js, Picking the necessasry methods doesn't mean other methods aren't passed as props, so we might as well just use the entire store as type and save a few lines of code

apps/frontend/core-dapp/src/features/trade/open-trade-utils.ts
```diff
@@ -0,0 +1,215 @@
...
+
+type Props = {
+  collateralStore: Pick<
+    CollateralStore,
+    'balanceOfSignerInWstEth' | 'depositFee' | 'percentDenominator'
+  >
+  depositTradeHelperStore: Pick<
+    DepositTradeHelperStore,
+    'depositAndTradeFee' | 'percentDenominator' | 'tradeFee'
+  >
```
chrisling-dev, line 13: so after watching your video explanation, I think we should just do this:

```suggestion
type Props = {
  collateralStore: CollateralStore
  depositTradeHelperStore: DepositTradeHelperStore
```

PR #1147, opened 2023-03-29
chore: add `TokenSender` to `MarketEntity` via `DynamicContract`

MeLlamoPablo:
This PR adds `RedeemHook` to `MarketEntity`, and then uses `getTokenSender` on `RedeemHook` to add `TokenSender` to `MarketEntity`.

Since this involves two queries to find out the address (`PrePOMarket.getRedeemHook -> RedeemHook.getTokenSender`), I introduced `DynamicContractStore` to reduce the overhead of initializing a dynamic contract. Previously we'd need to add a reaction that initializes the dynamic entity when the address is resolved. This is cumbersome, and also forces us to handle the optional case on the entity. Now, we simply pass the address as a function (`() => this.getAddress()`) and the `DynamicContractStore` will initialize the contract when it is available.


XavierEkkel (approved):
LGTM

chrisling-dev (commented):
Great approach with DynamicContractStore!

PR #1149, opened 2023-03-30
fix: increase dust threshold

MeLlamoPablo:
Screenshot 2023-03-30 at 11 28 02

This PR increases the dust threshold for the deposit and withdraw limits from `1M wei` to `0.0001 ETH`. This will consider the limits as exceeded if less than this amount is remaining.

If the remaining amount is below `0.0001 ETH`, it usually doesn't make sense to deposit or withdraw because the gas costs would be higher than the deposited or withdrawn amount.


XavierEkkel (approved):
LGTM

PR #1155, opened 2023-03-31
feat: modify RedeemHook to let allowlist not be configured

ramenforbreakfast:


apps/smart-contracts/core/test/unit/RedeemHook.test.ts
```diff
@@ -138,6 +138,14 @@ describe('=> RedeemHook', () => {
       ).revertedWith('Redeemer not allowed')
     })
 
+    it('ignores allowlist if not set', async () => {
+      expect(await msgSendersAllowlist.isIncluded(deployer.address)).to.be.true
+      expect(await allowlist.isIncluded(user.address)).to.eq(false)
```
XavierEkkel, line 143:
```suggestion
      expect(await msgSendersAllowlist.isIncluded(deployer.address)).true
      expect(await allowlist.isIncluded(user.address)).eq(false)
```

apps/smart-contracts/core/test/unit/RedeemHook.test.ts
```diff
@@ -138,6 +138,14 @@ describe('=> RedeemHook', () => {
...
+      await redeemHook.setAccountList(ZERO_ADDRESS)
+
+      await redeemHook.connect(deployer).hook(user.address, user.address, 1, 1)
```
XavierEkkel, line 146: add assert section below with check that allowlist isnt called

XavierEkkel (approved):
LGTM after comments

PR #1158, opened 2023-03-31
fix: redeem UI fixes

MeLlamoPablo:


apps/frontend/core-dapp/src/stores/entities/MarketEntity.ts
```diff
@@ -457,7 +457,7 @@ export class MarketEntity
   }
 
   get longTokenPrice(): number | undefined {
```
XavierEkkel, line 459: dont need the changes in this file anymore

XavierEkkel (approved):
LGTM after comment

PR #1161, opened 2023-03-31
fix: wrong balance limit calculation

MeLlamoPablo:


apps/frontend/core-dapp/src/utils/__tests__/balance-limits.test.ts
```diff
@@ -21,6 +21,22 @@ describe('getBalanceLimitInfo', () => {
...
+        currentAmount: parseEther('99.99985'),
+      })
+
+      if (result.status !== 'not-exceeded') {
+        throw new Error('Invalid status')
+      }
+
+      expect(result.amountEth).toBe('99.99985')
+      expect(result.capEth).toBe('100.0')
+      expect(result.remainingEth).toBe('0.00015')
```
XavierEkkel, line 38: should use the dust threshold constant directly, otherwise this test could break or become a false positive later if the dust threshold changes

XavierEkkel (approved):
LGTM after comment

PR #1171, opened 2023-04-03
fix: hardcoded valuation

MeLlamoPablo:
On #1166 we hardcoded the valuation on the "Market Expiry" tooltip because it wasn't showing on production. I suspect this was due to the `Trans` component (but can't confirm since it was working for me locally). Since we don't i18n our app yet, we can remove `Trans` from this definition.


chrisling-dev (approved):
Yeap seems like an issue with Trans, LGTM

PR #1172, opened 2023-04-03
feat: subgraph for tracking all relevant pools

chrisling-dev:
https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=6b4403e1ecc748a48a63e32a1f35e012&p=10346db6d83548c1b2dbc4edb5d7fdae&pm=s


apps/subgraphs/prepo-arbitrum/schema.graphql
```diff
@@ -21,6 +28,31 @@ type Market @entity {
...
+  longShortToken: LongShortToken!
+  "collateral token of this pool"
+  collateralToken: CollateralToken!
+  "0 or 1"
+  collateralTokenPosition: BigInt!
+  "token0 per token1"
+  token0Price: BigDecimal!
+  "token1 per token0"
+  token1Price: BigDecimal!
+  "The sqrt(price) of the pool after the swap, as a Q64.96"
```
MeLlamoPablo, line 48:
```suggestion
  "The sqrt(price) of the pool as a Q64.96"
```

"After the swap" doesn't make a lot of sense in this context, right?

apps/subgraphs/prepo-arbitrum/src/mappings/factory.ts
```diff
@@ -20,12 +31,57 @@ export function handleMarketCreation(event: MarketCreation): void {
...
+
+  const collateralTokenPosition = token0.type == TokenType.collateral ? 0 : 1
+
+  const poolAddress = event.params.pool.toHexString()
+  const pool = new Pool(poolAddress)
+  pool.longShortToken = collateralTokenPosition === 0 ? token1.id : token0.id
+  pool.collateralToken = collateralTokenPosition === 0 ? token0.id : token1.id
+  pool.collateralTokenPosition = BigInt.fromI32(collateralTokenPosition)
+  pool.token0 = token0.id
+  pool.token1 = token1.id
```
MeLlamoPablo, line 78: What's the point of storing `token0` and `token1` if we already know which is which? Can't we just store `longShortToken` and `collateralToken`?

Asking out of ignorance, so feel free to disregard.

MeLlamoPablo (approved):

apps/subgraphs/prepo-arbitrum/src/mappings/factory.ts
chrisling-dev: so we use token0 and token1 for ease of calculating price of both tokens. So we don't have to keep doing things like 

```
if(collateralTokenPosition == 0) {
  token0 = collateralToken.id
  token1 = longShortToken.id
  calculateSqrtPrice(token0, token1)
}
```

and can just do
```
const prices = calculateSqrtPrice(pool.token0, pool.token1)
```

The code block above is just an example, the actual implementation could be very different

apps/subgraphs/prepo-arbitrum/schema.graphql
chrisling-dev: Yea good point, I think I copied this description from uniswap's subgraph's Swap entity which made sense in that context, thanks for pointing out

PR #1173, opened 2023-04-03
feat: open trade using `DepositTradeHelper`

MeLlamoPablo:
This PR redirects open position trades via `tradeForPosition` of `DepositTradeHelper`, as long as the `TRADE_VIA_HELPER_ENABLED` feature flag is enabled (currently only on preview).

This allows to:

* Approve using permits to save a transaction.
    * This only needs to be done once. The approval will be remembered for the future.
* Collect the prePO fees


apps/frontend/core-dapp/src/features/trade/TradeStore.ts
```diff
@@ -500,6 +493,17 @@ export class TradeStore {
     )
   }
 
+  get needPermitForOpenTrade(): boolean | undefined {
+    if (!TRADE_VIA_HELPER_ENABLED) return false
+
+    if (!this.root.web3Store.connected) return false
```
chrisling-dev, line 499: Do we need this line? The function shouldn't matter if not connected right?

apps/frontend/core-dapp/src/stores/DepositTradeHelperStore.ts
```diff
@@ -94,6 +96,71 @@ export class DepositTradeHelperStore extends ContractStore<RootStore, SupportedC
...
+      if (needsPermit) {
+        const collateralPermit = await this.root.collateralStore.getPermitSignature(
+          depositTradeHelperAddress,
+          ethers.constants.MaxUint256,
+          deadline
+        )
+
+        if (typeof collateralPermit === 'string') {
+          return { success: false, error: collateralPermit }
+        }
```
chrisling-dev, line 133: This looks weird to me, can't we make it throw an error if there's an error and handle it properly in a catch block?

apps/frontend/core-dapp/src/stores/DepositTradeHelperStore.ts
```diff
@@ -94,6 +96,71 @@ export class DepositTradeHelperStore extends ContractStore<RootStore, SupportedC
...
+        permit,
+        {
+          amountOutMinimum:
+            this.root.advancedSettingsStore.getAmountAfterSlippageForTrades(positionTokenAmountOut),
+          deadline,
+          positionToken,
+          // This parameter defines the maximum price the trade can push the
+          // pool to. It is designed to protect against price impact, but we
+          // don't need it because we handle price impact in the UI, and we
+          // already define an amountOutMinimum.
```
chrisling-dev, line 153: I think can leave this comment out, been studying a lot about uniswap V3 lately and most of the examples in their docs use 0 for sqrtPriceLimitX96 anyways

chrisling-dev (changes requested):
The main thing I don't agree with is error handling of permit in tradeForPosition, lmkwyt!

apps/frontend/core-dapp/src/features/trade/TradeStore.ts
MeLlamoPablo: Indeed, I just copied it from `needApprovalForOpenTrade` but it doesn't have any effect here.

apps/frontend/core-dapp/src/stores/DepositTradeHelperStore.ts
MeLlamoPablo: Done, also used this pattern on other places btw.

chrisling-dev: oh I meant it's weird that we're checking for error from a string `typeof collateralPermit === 'string'`, don't we either return an object with error or just throw error? or does this have to be string because of ether's sign function?

MeLlamoPablo: [Here](https://github.com/prepo-io/prepo-monorepo/blob/29036b0e410cf70a2e3b64e65ea29a4181c33067/apps/frontend/core-dapp/src/stores/entities/Erc20Permit.entity.ts#L67-L68) we return an error string instead of throwing the error. I agree that it's weird and we should change it, but that's beside the scope of this PR I think. I didn't write this code.

chrisling-dev (approved):
LGTM

PR #1177, opened 2023-04-04
refactor: allow no return type if function is an expression

MeLlamoPablo:
This PR changes our eslint config to enable the `allowExpressions` flag on the `@typescript-eslint/explicit-function-return-type` rule:

```js
'@typescript-eslint/explicit-function-return-type': ['error', {
  allowExpressions: true,
}],
```

It's good to have this rule enabled because it forces us to type our functions, which makes our code more robust and prevents our functions from accidentally returning the wrong thing. However, sometimes the return value of a function can already be inferred from the context. For example, we know that `onClick` needs to return `void`, so manually adding `: void` doesn't add any value and adds frictions to our DX.

With `allowExpressions: true`, the rule will not error in these cases, while still erroring when we forget the return type of standalone functions.


chrisling-dev (approved):
LGTM!

PR #1178, opened 2023-04-04
refactor: use Fraction for fees

MeLlamoPablo:
This PR uses `Fraction` for `depositFee` and `withdrawFee` in `CollateralStore` and `depositAndTradeFee` and `tradeFee` In `DepositTradeHelperStore`.

`Fraction` allows us to abstract away the `percentDenominator`, which makes the code that consumes the fees simpler and more readable. Plus, they link the two related values (numerator and denominator) to ensure that we never use the one without the other.


chrisling-dev (approved):
LGTM!

PR #1179, opened 2023-04-04
feat: handle multiple `TokenSender`s to calculate PPO rebates

MeLlamoPablo:
✅ After merging this PR we should be able to QA and deploy Deposit and Trade

------

This PR correctly handles multiple `TokenSender`s. When doing deposit and trade, there are two fees: the `deposit` fee and the `depositAndTrade` fee. The PPO rebates corresponding to those come from multiple `TokenSender`s.

In order to avoid handling this in `TradeStore`, which would introduce more complexity, I added a `FeeStore`. Now, instead of each store querying the fees, calculating the amount, the PPO rebate and its ETH value, each store would simply create one `FeeStore` for each type of fee. When creating the `FeeStore`, the caller store simply links it to a wstETH value and a token sender, and `FeeStore` does the rest.

My intention is, if this proposal is accepted, to migrate al fees to `FeeStore`, which should simplify PPO reward calculations everywhere.

`FeeStore` also logs a debug table on the console which will help us confirm that the values are correct:

Screenshot 2023-04-04 at 19 00 25

Screenshot 2023-04-04 at 19 00 57

Also, this PR introduces the `TRADE_VIA_HELPER_ENABLED` flag (currently enabled only on previews) so that we don't accidentally display PPO rewards for normal open trades. This fee is not applicable until we implement trading via `DepositTradeHelper` in #1173


apps/frontend/core-dapp/src/features/trade/TradeStore.ts
```diff
@@ -568,36 +569,45 @@ export class TradeStore {
...
    *
    * If the user is paying with collateral (preWstEth), then this value is
    * exact. If the user is paying with native ETH, then this value is estimated,
    * since it depends on the Balancer quote.
    *
    * This does not include fees paid to Balancer or Uniswap.
    */
-  private get openTradeFeeInWstEthBN(): BigNumber | undefined {
-    const { openTradeAmountAfterDepositSlippageInWstEthBN, paymentToken } = this
+  private get openTradeFeesInWstEthBN(): FeeStore[] {
```
chrisling-dev, line 580: technically the returned value isn't BigNumber nor in wst eth so this should be renamed to openTradeFees right?
```suggestion
  private get openTradeFees(): FeeStore[] {
```

apps/frontend/core-dapp/src/features/trade/TradeStore.ts
```diff
@@ -615,15 +625,19 @@ export class TradeStore {
    * We shouldn't pass this value to the blockchain.
    */
   private get openTradeAmountForQuoteInWstETHBN(): BigNumber | undefined {
-    const { openTradeAmountAfterDepositSlippageInWstEthBN, openTradeFeeInWstEthBN } = this
+    const { openTradeAmountAfterDepositSlippageInWstEthBN, openTradeFeesInWstEthBN } = this
 
     if (
       openTradeAmountAfterDepositSlippageInWstEthBN === undefined ||
-      openTradeFeeInWstEthBN === undefined
+      openTradeFeesInWstEthBN === undefined
```
chrisling-dev, line 632: and should this check be removed since it will never be undefined? or updated to something like `openTradeFees.length === 0` ?

apps/frontend/core-dapp/src/stores/entities/FeeEntity.ts
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  static getTotalFeeInWstEth(fees: readonly FeeStore[]): BigNumber | undefined {
+    let total = BigNumber.from(0)
+
+    for (let i = 0; i < fees.length; i++) {
+      const fee = fees[i]
+      if (fee.feeAmountInWstEth === undefined) return undefined
+      total = total.add(fee.feeAmountInWstEth)
+    }
```
chrisling-dev, line 73: would prefer that we use `forEach` if possible, for better code readability and also less room for mistake
```suggestion
    fees.forEach((fee) => {
      if (fee.feeAmountInWstEth === undefined) return undefined
      total = total.add(fee.feeAmountInWstEth)
    })
```

apps/frontend/core-dapp/src/stores/entities/FeeEntity.ts
```diff
@@ -0,0 +1,118 @@
...
+      if (rewardValue === undefined) return undefined
+      totalEth = totalEth.add(rewardValue)
+
+      debugInfo.push({
+        tokenSender: `${fee.tokenSender.address?.slice(0, 6)}`,
+        fee: `${toPercent(fee.fee).toFixed(2)} %`,
+        ppo: `${formatEther(reward)} PPO`,
+        eth: `${formatEther(rewardValue)} ETH`,
+      })
+    }
```
chrisling-dev, line 102: ditto if we could use forEach

apps/frontend/core-dapp/src/stores/entities/FeeEntity.ts
```diff
@@ -0,0 +1,118 @@
...
+      const rewardValue = fee.tokenSender.calculateRewardValueInEth(reward)
+      if (rewardValue === undefined) return undefined
+      totalEth = totalEth.add(rewardValue)
+
+      debugInfo.push({
+        tokenSender: `${fee.tokenSender.address?.slice(0, 6)}`,
+        fee: `${toPercent(fee.fee).toFixed(2)} %`,
+        ppo: `${formatEther(reward)} PPO`,
+        eth: `${formatEther(rewardValue)} ETH`,
+      })
```
chrisling-dev, line 101: I like this debugInfo, and think we should keep it, but could we hide it with isProduction check?

chrisling-dev (changes requested):
Nice work! I like this approach with FeeStore, only some minor comments

apps/frontend/core-dapp/src/stores/entities/FeeEntity.ts
MeLlamoPablo: Mm, not sure about that. In the past we've had the fee appear to be wrong in production and debugging that was a bit tedious. If we've had that it would have been easy to inspect how the fee was being calculated.

My intention was to enable it on production for a few days and hide it or remove it some days after. What do you think?

apps/frontend/core-dapp/src/stores/entities/FeeEntity.ts
MeLlamoPablo: That wouldn't work because the `return undefined` is intended to return from the entire `getTotalFeeInWstEth`, and here it would return from the inline `forEach` function instead.

I tried to do `for..of` for better readability but our [`no-restricted-syntax` ESLint rule](https://eslint.org/docs/latest/rules/no-restricted-syntax) prevents that.

apps/frontend/core-dapp/src/features/trade/TradeStore.ts
MeLlamoPablo: Good catch!

apps/frontend/core-dapp/src/features/trade/TradeStore.ts
MeLlamoPablo: Actually I added `| undefined` to the type instead, because I think that an empty array is not a good indication that the fees are loading (could be that there are no fees). So thanks for the feedback.

apps/frontend/core-dapp/src/stores/entities/FeeEntity.ts
chrisling-dev: oh you're right, it needs to return, np!

chrisling-dev (approved):
LGTM, a PR like this could be split into smaller PRs for better reviewability, I'd probably do:  
- First PR for FeeStore without reward stuff, this way the first PR would only ever have fee stuff
- Second PR for PPO reward

that way both PRs would be much easier to review IMO

PR #1182, opened 2023-04-05
feat: stronger unique tx id to prevent duplicate tx when both actions…

chrisling-dev:
… happen in same blockchain tx

https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=6b4403e1ecc748a48a63e32a1f35e012&p=892f4d9bec3345ef804c4dfa9eb21a1a&pm=s


MeLlamoPablo (approved):
LGTM!

Curious about what happens when you make a change like this? Does The Graph drop all data and reconstruct it from scratch? Or is it smart and knows what to update exactly?

chrisling-dev:
> LGTM!
> 
> Curious about what happens when you make a change like this? Does The Graph drop all data and reconstruct it from scratch? Or is it smart and knows what to update exactly?

it drops all old data and reconstruct them. Basically every time we make a change and deploy that change, regardless of how small that change is, The Graph will re-index every block of the blockchain

PR #1183, opened 2023-04-05
feat: position costbasis

chrisling-dev:
https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=6b4403e1ecc748a48a63e32a1f35e012&p=dab08ef020de4a439f1667af41ef287a&pm=s


apps/subgraphs/prepo-arbitrum/schema.graphql
```diff
@@ -97,3 +97,16 @@ type Token @entity {
...
+  "template: {USER_ADDRESS}-{LONG_SHORT_TOKEN_ADDRESS}}"
+  id: ID!
+  "average cost per token - for PNL calculation"
+  costBasis: BigDecimal!
+  "address who owned this position"
+  ownerAddress: String!
+  "token of which this position is tracking"
+  longShortToken: LongShortToken!
+  "balance when cost basis was last updated"
+  balance: BigDecimal!
```
MeLlamoPablo, line 111: Is this balance in ETH, wstETH or in position tokens? Could you clarify in the comment?

apps/subgraphs/prepo-arbitrum/schema.graphql
```diff
@@ -97,3 +97,16 @@ type Token @entity {
   "e.g. Long, Short, Collateral, CollateralBase"
   type: String!
 }
+
+type Position @entity {
+  "template: {USER_ADDRESS}-{LONG_SHORT_TOKEN_ADDRESS}}"
+  id: ID!
+  "average cost per token - for PNL calculation"
+  costBasis: BigDecimal!
```
MeLlamoPablo, line 105: Same here, I assume this is in wstETH?

apps/subgraphs/prepo-arbitrum/src/utils/math.ts
```diff
@@ -0,0 +1,42 @@
+import { BigDecimal, BigInt } from '@graphprotocol/graph-ts'
+import { ONE_BI, ZERO_BD, ZERO_BI } from './constants'
+
+export function exponentToBigDecimal(decimals: BigInt): BigDecimal {
+  let bd = BigDecimal.fromString('1')
+  for (let i = ZERO_BI; i.lt(decimals); i = i.plus(ONE_BI)) {
+    bd = bd.times(BigDecimal.fromString('10'))
+  }
+  return bd
+}
```
MeLlamoPablo, line 10: I was confused a bit by the naming of this function. Maybe we could name it something like `tenToThePowerOf` so it reads like `tenToThePowerOf(exchangeDecimals)`?

I assume that what this function is trying to do is $10^{exchangeDecimals}$, right?

apps/subgraphs/prepo-arbitrum/src/utils/math.ts
```diff
@@ -0,0 +1,42 @@
...
+
+// ref: https://github.com/Uniswap/v3-subgraph/blob/bf03f940f17c3d32ee58bd37386f26713cff21e2/src/utils/pricing.ts#L48
+export function sqrtPriceX96ToTokenPrices(
+  sqrtPriceX96: BigInt,
+  token0Decimals: BigInt,
+  token1Decimals: BigInt
+): BigDecimal[] {
+  const num = sqrtPriceX96.times(sqrtPriceX96).toBigDecimal()
+  const denom = BigDecimal.fromString(Math.pow(2, 192).toString())
```
MeLlamoPablo, line 34: `Math.pow(2, 192)` is larger than `Number.MAX_SAFE_INTEGER`. Do you think this could cause issues? I assume not, because the Uniswap code does the same thing, but it feels weird to me.

apps/subgraphs/prepo-arbitrum/schema.graphql
chrisling-dev: oh my bad this isnt necessary, added this during debug, will remove

apps/subgraphs/prepo-arbitrum/schema.graphql
chrisling-dev: price is always in wstETH

apps/subgraphs/prepo-arbitrum/src/utils/math.ts
chrisling-dev: nope we've been using the same logic in the past year on testnet, also this code was copied over from Uniswap's subgraph code

apps/subgraphs/prepo-arbitrum/src/utils/math.ts
chrisling-dev: yeap good suggestion

PR #1184, opened 2023-04-05
feat: close trade using `DepositTradeHelper`

MeLlamoPablo:
This PR redirects close position trades via `tradeForCollateral` of `DepositTradeHelper`, as long as the `TRADE_VIA_HELPER_ENABLED` feature flag is enabled (currently only on preview).

It also displays the PPO rewards corresponding to the `tradeFee` on the close trade page.


apps/frontend/core-dapp/src/features/trade/TradeStore.ts
```diff
@@ -347,6 +354,19 @@ export class TradeStore {
     )
   }
 
+  get closePositionNeedPermit(): boolean | undefined {
+    if (!TRADE_VIA_HELPER_ENABLED) return false
+
+    if (!this.selectedPosition || !this.root.web3Store.connected) return false
+    if (this.closePositionAmount === undefined) return undefined
+    if (this.selectedMarket?.resolved) return false
```
chrisling-dev, line 362: Would this change affect UX? If not we should group them for readability, lmkwyt
```suggestion
    if (!this.selectedPosition || !this.root.web3Store.connected || this.selectedMarket?.resolved) return false
    if (this.closePositionAmount === undefined) return undefined
```

apps/frontend/core-dapp/src/stores/DepositTradeHelperStore.ts
```diff
@@ -96,6 +97,71 @@ export class DepositTradeHelperStore extends ContractStore<RootStore, SupportedC
...
+        permit,
+        {
+          amountOutMinimum:
+            this.root.advancedSettingsStore.getAmountAfterSlippageForTrades(collateralAmountOut),
+          deadline,
+          positionToken: positionToken.address,
+          // This parameter defines the maximum price the trade can push the
+          // pool to. It is designed to protect against price impact, but we
+          // don't need it because we handle price impact in the UI, and we
+          // already define an amountOutMinimum.
```
chrisling-dev, line 154: can remove this

chrisling-dev (approved):
LGTM, approving since comments aren't blocker

apps/frontend/core-dapp/src/features/trade/TradeStore.ts
MeLlamoPablo: Good catch, I don't think it does affect anything

PR #1187, opened 2023-04-06
feat: remove multiplier from tokensender and other minor changes

ramenforbreakfast:
This PR contains only the removal of the price multiplier and a few surface-level changes including:
- `scaledPriceLowerBound` => `priceLowerBound` since there is no scaling
- Change `outputTokenDecimalsFactor` to `outputTokenUnit` to be consistent and agnostic to the way a token unit is being used.
- Change `setPrice` to `setPriceOracle`. While making changes I found myself very irritated by the fact that in the `TokenSender` logic, we need to get the `price` and store it, yet the oracle contract itself is named `price`. I find this to be very confusing and strongly believe it is better to use `priceOracle` instead. I have made this change

While this means the lower bound will now only be applied to the raw price, I think this makes much more sense because we should be basing our protection from a low price read against the price directly, rather than a scaled amount which is harder to reason with.


apps/smart-contracts/core/contracts/TokenSender.sol
```diff
@@ -14,68 +14,55 @@ contract TokenSender is
...
   bytes32 public constant override WITHDRAW_ERC20_ROLE =
     keccak256("withdrawERC20");
 
   constructor(IERC20 outputToken, uint256 outputTokenDecimals) {
     _outputToken = outputToken;
-    _outputTokenDecimalsFactor = 10**outputTokenDecimals;
+    _outputTokenUnit = 10**outputTokenDecimals;
   }
 
   function send(address recipient, uint256 unconvertedAmount)
```
XavierEkkel, line 37:
```suggestion
  function send(address recipient, uint256 inputAmount)
```
since the scaling already happens prior

packages/prepo-shared-contracts/contracts/interfaces/ITokenSender.sol
```diff
@@ -5,37 +5,25 @@ import "./IUintValue.sol";
...
 interface ITokenSender {
-  event PriceChange(IUintValue price);
+  event PriceOracleChange(IUintValue oracle);
 
-  event PriceMultiplierChange(uint256 priceMultiplier);
-
-  event ScaledPriceLowerBoundChange(uint256 scaledPrice);
+  event PriceLowerBoundChange(uint256 price);
 
   function send(address recipient, uint256 unconvertedAmount) external;
```
XavierEkkel, line 12:
```suggestion
  function send(address recipient, uint256 inputAmount) external;
```

apps/smart-contracts/core/test/unit/TokenSender.test.ts
```diff
@@ -63,169 +61,112 @@ describe('=> TokenSender', () => {
     it('sets output token from constructor', async () => {
       expect(await tokenSender.getOutputToken()).to.eq(outputToken.address)
     })
-    it('does not set price', async () => {
-      expect(await tokenSender.getPrice()).to.eq(ZERO_ADDRESS)
+    it('does not set price oracle', async () => {
```
XavierEkkel, line 64: nit:
```suggestion
    it('doesn't set price oracle', async () => {
```
same elsewhere

apps/smart-contracts/core/test/unit/TokenSender.test.ts
```diff
@@ -281,38 +219,23 @@ describe('=> TokenSender', () => {
...
+      await tokenSender.connect(deployer).setPriceLowerBound(2)
 
       await tokenSender.connect(deployer).send(user.address, 1)
 
       expect(outputToken.transfer).not.called
     })
 
-    it("doesn't transfer if scaled price = lowerBound", async () => {
+    it("doesn't transfer if price = lowerBound", async () => {
       priceOracle.get.returns(1)
```
XavierEkkel, line 238: add expect that lowerBound = 1

XavierEkkel (approved):
LGTM after comments

apps/smart-contracts/core/test/unit/TokenSender.test.ts
XavierEkkel, line 64: nit: missing empty lines between each of these tests

PR #1188, opened 2023-04-06
feat: add limit logic to TokenSender and perform half-way refactor of tests

ramenforbreakfast:
In addition to importing the module and adding the necessary roles for `TokenSender`, I performed some not-invasive refactoring to the test suite here since I needed to write tests here and the current state of these tests leaves much to be desired (not coverage wise, just readability/consistency-wise).

I went ahead and added snapshots to it, and added 4 test cases to ensure that the amounts are recorded under the caller and not the recipient, and to ensure it also reverts correctly, but did not test it beyond that since that is redundant.


packages/prepo-shared-contracts/contracts/interfaces/ITokenSender.sol
```diff
@@ -27,5 +27,9 @@ interface ITokenSender {
 
   function SET_ALLOWED_MSG_SENDERS_ROLE() external view returns (bytes32);
 
+  function SET_ACCOUNT_RESET_PERIOD_ROLE() external view returns (bytes32);
```
XavierEkkel, line 30:
```suggestion
  function SET_ACCOUNT_LIMIT_RESET_PERIOD_ROLE() external view returns (bytes32);
```

apps/smart-contracts/core/contracts/TokenSender.sol
```diff
@@ -73,6 +80,22 @@ contract TokenSender is
     super.setAllowedMsgSenders(allowedMsgSenders);
   }
 
+  function setAccountResetPeriod(uint256 accountResetPeriod)
+    public
+    override
+    onlyRole(SET_ACCOUNT_RESET_PERIOD_ROLE)
+  {
+    super.setAccountResetPeriod(accountResetPeriod);
```
XavierEkkel, line 88:
```suggestion

  function setAccountLimitResetPeriod(uint256 period)
    public
    override
    onlyRole(SET_ACCOUNT_LIMIT_RESET_PERIOD_ROLE)
  {
    super.setAccountLimitResetPeriod(period);
```

similarly change the argument name in the interface itself to reflect this new convention

apps/smart-contracts/core/contracts/TokenSender.sol
```diff
@@ -73,6 +80,22 @@ contract TokenSender is
...
+  }
+
+  function setAccountLimitPerPeriod(uint256 accountLimitPerPeriod)
```
XavierEkkel, line 91:
```suggestion
  function setAccountLimitPerPeriod(uint256 limitPerPeriod)
```

apps/smart-contracts/core/contracts/TokenSender.sol
```diff
@@ -26,6 +28,10 @@ contract TokenSender is
     keccak256("setPriceLowerBound");
   bytes32 public constant override SET_ALLOWED_MSG_SENDERS_ROLE =
     keccak256("setAllowedMsgSenders");
+  bytes32 public constant override SET_ACCOUNT_RESET_PERIOD_ROLE =
+    keccak256("setAccountResetPeriod");
```
XavierEkkel, line 32:
```suggestion
  bytes32 public constant override SET_ACCOUNT_LIMIT_RESET_PERIOD_ROLE =
    keccak256("setAccountLimitResetPeriod");
```

apps/smart-contracts/core/test/unit/TokenSender.test.ts
```diff
@@ -71,11 +53,24 @@ describe('=> TokenSender', () => {
       expect(await tokenSender.SET_PRICE_ORACLE_ROLE()).to.eq(id('setPriceOracle'))
       expect(await tokenSender.SET_PRICE_LOWER_BOUND_ROLE()).to.eq(id('setPriceLowerBound'))
       expect(await tokenSender.SET_ALLOWED_MSG_SENDERS_ROLE()).to.eq(id('setAllowedMsgSenders'))
+      expect(await tokenSender.SET_ACCOUNT_RESET_PERIOD_ROLE()).to.eq(id('setAccountResetPeriod'))
```
XavierEkkel, line 56:
```suggestion
      expect(await tokenSender.SET_ACCOUNT_LIMIT_RESET_PERIOD_ROLE()).to.eq(id('setAccountResetPeriod'))
```

apps/smart-contracts/core/test/unit/TokenSender.test.ts
```diff
@@ -71,11 +53,24 @@ describe('=> TokenSender', () => {
...
+      expect(await tokenSender.SET_ACCOUNT_LIMIT_PER_PERIOD_ROLE()).to.eq(
+        id('setAccountLimitPerPeriod')
+      )
       expect(await tokenSender.WITHDRAW_ERC20_ROLE()).eq(id('withdrawERC20'))
     })
   })
 
   describe('# setPriceOracle', () => {
+    beforeEach(async () => {
```
XavierEkkel, line 65: wouldnt this (and similarly below) actually be slower than before, because these beforeEachs aren't snapshotted? why not keep in the snapshotted before block?

ramenforbreakfast: Im going to undo that change and just assign them all at the beginning, at the time I was trying to avoid assigning them all prior to setter tests, because that means we could arrive at a false positive. Say a function should on role X, but since role X, Y, and Z are assigned at the beginning, we would not be able to spot a bug where that function actually relied on role Y, since all of them were assigned.

However, this really doesn't matter anymore, because I forgot we are shifting over to a model where the deployer is automatically assigned all roles.

apps/smart-contracts/core/contracts/TokenSender.sol
ramenforbreakfast: I think you mean `accountLimitResetPeriod` for the input name? This breaks from our convention of always having the input be named after the variable itself, in fact I think there is one setter here that breaks that convention.

ramenforbreakfast:
Updated with changes from previous PRs to the TokenSender naming and various comments

apps/smart-contracts/core/test/unit/TokenSender.test.ts
```diff
@@ -186,15 +175,103 @@ describe('=> TokenSender', () => {
...
 
-  describe('# send', () => {
+  describe('# setAccountLimitResetPeriod', () => {
+    it('reverts if not role holder', async () => {
+      expect(
+        await tokenSender.hasRole(
+          await tokenSender.SET_ACCOUNT_LIMIT_RESET_PERIOD_ROLE(),
+          user.address
+        )
+      ).to.eq(false)
```
XavierEkkel, line 185:
```suggestion
      ).eq(false)
```
similarly for rest of file

apps/smart-contracts/core/test/unit/TokenSender.test.ts
```diff
@@ -269,8 +346,58 @@ describe('=> TokenSender', () => {
       expect(outputToken.transfer).not.called
     })
 
+    it('reverts if sent amount > account limit and caller != recipient', async () => {
+      priceOracle.get.returns(OUTPUT_TOKEN_UNIT) // 1 for 1
+      const outputAmount = await calculateExpectedOutput(TEST_LIMIT.add(1))
+      expect(outputAmount).to.be.eq(TEST_LIMIT.add(1))
```
XavierEkkel, line 352:
```suggestion
      expect(outputAmount).eq(TEST_LIMIT.add(1))
```
same for rest of file

apps/smart-contracts/core/test/unit/TokenSender.test.ts
```diff
@@ -269,8 +346,58 @@ describe('=> TokenSender', () => {
...
+      await outputToken.connect(deployer).mint(tokenSender.address, TEST_LIMIT.add(1))
+
+      await expect(tokenSender.connect(deployer).send(deployer.address, TEST_LIMIT.add(1)))
+        .revertedWithCustomError(tokenSender, 'AccountLimitExceeded')
+        .withArgs(deployer.address, TEST_LIMIT.add(1))
+    })
+
+    it("adds sent amount to caller's total if <= account limit and caller != recipient", async () => {
```
XavierEkkel, line 371:
```suggestion
    it("adds sent amount to caller's total if = account limit and caller != recipient", async () => {
```
and separate test for `<` case

same for similar test below

apps/smart-contracts/core/test/unit/TokenSender.test.ts
```diff
@@ -269,8 +346,58 @@ describe('=> TokenSender', () => {
...
+      priceOracle.get.returns(OUTPUT_TOKEN_UNIT)
+      const outputAmount = await calculateExpectedOutput(TEST_LIMIT)
+      expect(outputAmount).to.be.eq(TEST_LIMIT)
+      const totalSentBefore = await tokenSender.getAmountThisPeriod(user.address)
```
XavierEkkel, line 375: add expect that this is 0
same for test below

XavierEkkel (approved):
LGTM after comments

PR #1190, opened 2023-04-06
fix: font too bold on Safari

MeLlamoPablo:
It looks like the addition of the variable font makes the font bolder than it should on Safari. The effect is especially bad when I open safari on an HDMI monitor instead of the MacBook Retina screen.

I think the easiest solution is to just remove the usage of the variable font.

#### Before (Safari)

Screenshot 2023-04-06 at 12 52 29

#### After (Safari)

Screenshot 2023-04-06 at 12 51 41

#### Before (Firefox, for comparison)

Screenshot 2023-04-06 at 12 56 48

#### Before (Safari)

Screenshot 2023-04-06 at 12 53 34

#### After (Safari)

Screenshot 2023-04-06 at 12 54 59

#### Before (Firefox, for comparison)

Screenshot 2023-04-06 at 12 57 18


chrisling-dev (approved):
Interesting, I added the custom Inter font because the "default" Inter font doesn't support the top right error icon and was showing an empty box, somehow it works for me now, could be caching issue with previous font. LGTM! Thanks for fixing this

PR #1200, opened 2023-04-07
fix: minor ux issue on portfolio page

chrisling-dev:
Fixes some minor UX issue:
- adding a tx would reflect in the new tx history (by increasing one hour of end timestamp)
- show use absolute number for displaying amount (close event can have negative amount)
- date format function has bug: the format function is returning min inaccurately


apps/frontend/core-dapp/src/features/history/HistoryTransaction.tsx
```diff
@@ -122,7 +122,10 @@ const HistoryTransaction: React.FC<Props> = ({ transaction }) => {
 
           {metadata.direction && <PositionLabel positionType={metadata.direction} />}
         </Flex>
-        <Label>{displayEth(+amount)}</Label>
+        {/** amount can be negative value when closing, we need to make it absolute number while maintaing it's type as string
```
MeLlamoPablo, line 125:
```suggestion
        {/** amount can be negative value when closing, we need to make it absolute number while maintaing its type as string
```

MeLlamoPablo (approved):

apps/frontend/core-dapp/src/features/history/HistoryTransaction.tsx
chrisling-dev: will address in next PR to avoid another build

PR #1202, opened 2023-04-07
chore: use `NEXT_PUBLIC_VERCEL_ENV` to check for production

MeLlamoPablo:
This PR switches from our custom `NEXT_PUBLIC_STAGE` variable to `NEXT_PUBLIC_VERCEL_ENV` to determine if we're on production. This is better because it's a native Vercer variable so we don't have to worry about defining our own.

`NEXT_PUBLIC_VERCEL_ENV` is:

* `production` on production
* `preview` on preview
* `undefined` on the local dev server


chrisling-dev (approved):
LGTM

PR #1205, opened 2023-04-07
refactor: extract `SimulatorCard` from `TradeSimulator`

MeLlamoPablo:
This PR extracts a pure reusable `SimulatorCard` component out of `TradeSimulator`. `SimulatorCard` doesn't depend on the root store and therefore I'll be able to reuse it on the custom simulator page.

It also removes the two `useEffects` that were syncing `marketEntryPrice <-> entryPrice` and `marketExitPrice <-> exitPrice`, and replaces them with a new abstraction: `useFormValue`.

This hook handles the syncing, the dirty detection and the resetting internally, and thus they don't need to happen inside `TradeSimulator`. In my opinion, state syncing is one of the biggest sources of complexity in React components (and UI in general), so I think it's important that this complexity is isolated away and unit tested.

`useFormValue` behaves similarly to a Formik value. You pass a `baseValue` and the hook returns it, until the user modifies it, in which case it becomes dirty.

* If the `baseValue` changes, and the value is not dirty, it will be synced.
* If the `baseValue` changes, and the value is dirty, it will not be synced (the user value will be preserved)


chrisling-dev (approved):
Great PR! LGTM

PR #1206, opened 2023-04-07
feat: add basic simulator page

MeLlamoPablo:
Screenshot 2023-04-14 at 18 43 18

This PR adds a basic simulator page that uses default valuation and payout ranges and lets you customize them in the settings. The default entry and exit price are mock values and not meaningful.

The page lives on `/simulator` without a feature flag but there are no links to it anywhere.

Regarding react-hook-form, I went back and forth but eventually decided not to include it because even though it would handle some complexity (replace `useFormValue`) it would also introduce other kinds of frictions. We can discuss this more in-depth if you're curious.

The design was altered a bit: the settings icon was removed and the settings are always visible. I made it this way for ease testing, but we decided that it was a simpler UI so we kept it that way.

To do:

- [ ] Support trade size
- [ ] Support billions/millions switch for valuation
- [ ] Have a market selector to be able to select a preset (valuation + payout) from those of a market. If the user edits a value, it switches to "Custom".
- [ ] If you select a market, it should auto fill the entry price to the correct value. If you change the entry or exit price, it should mark it as dirty and be able to reset. If you're on "Custom" it doesn't make sense to mark those values as dirty, and maybe we can use halfway as the entry point.
- [ ] Persist custom settings in local storage?
- [ ] Anything else I missed?


apps/frontend/core-dapp/src/components/DirectionRadio.tsx
```diff
@@ -3,60 +3,88 @@ import styled, { Color, css, FlattenSimpleInterpolation } from 'styled-component
...
   font-weight: ${({ theme }): number => theme.fontWeight.semiBold};
-  line-height: 24px;
   margin-bottom: 0;
   ${removeUserSelect}
+
+  ${({ $size, $selected, theme }): FlattenSimpleInterpolation => {
+    if ($size === 'sm') {
+      return css`
+        font-size: ${theme.fontSize.sm};
+        line-height: ${spacingIncrement(14)};
```
chrisling-dev, line 72: border radius for small size should be 12px so it's not fully rounded. I don't think we use fully rounded elements anywhere

chrisling-dev (changes requested):
Thoughts on design, looking at code now

The input width seems wrong in preview? and width changes as I type 

Thoughts on design, feels weird that nothing is in the center (no main CTA), if we're going with this approach, I think the simulator should be in the center and settings to the right of it (like trade page)

Screenshot 2023-04-15 at 8 09 58 AM

apps/frontend/core-dapp/src/features/simulator/SimulatorSettings.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+const NumberInput: React.FC<{
+  isValid: boolean
+  max: number
+  min: number
+  suffix: string
+  value: FormValue<string>
+}> = ({ isValid, max, min, suffix, value: { value, setValue } }) => (
+  <Input
+    error={!isValid}
```
chrisling-dev, line 73: Should add block to keep it fill the width

```suggestion
  <Input
    block
    error={!isValid}
```

apps/frontend/core-dapp/src/features/simulator/SimulatorSettings.tsx
```diff
@@ -0,0 +1,156 @@
...
+      if (e.key === '-') {
+        e.preventDefault()
+        return
+      }
+
+      if (ALLOWED_DECIMAL_SEPARATORS.includes(e.key) && e.target instanceof HTMLInputElement) {
+        e.preventDefault()
+        setValue(`${e.target.value}.`)
+      }
+    }}
```
chrisling-dev, line 89: There's probably some weird behaviour about this, try typing 10. in the valuation input and it just clears the input. Actually just typing "." or "," will empty the input and cause the other input to be inValid

I think the inputs could just be integer?

apps/frontend/core-dapp/src/features/simulator/SimulatorPage.tsx
```diff
@@ -0,0 +1,159 @@
...
+  return [+valuationMin >= 0 && +valuationMin <= 1000, +valuationMax >= 0 && +valuationMax <= 1000]
+}
+
+const allValid = ([min, max]: RangeValidity): boolean => min && max
+
+const SimulatorPage: React.FC = () => {
+  const [direction, setDirection] = useState<'long' | 'short'>('long')
+
+  const entryPrice = useFormValue<number | undefined>(0.5)
+  const exitPrice = useFormValue<number | undefined>(0.6)
```
chrisling-dev, line 101: default exit should be max profit scenario, 0.9?

apps/frontend/core-dapp/src/features/simulator/SimulatorPage.tsx
```diff
@@ -0,0 +1,159 @@
...
+}): RangeValidity {
+  if (Number.isNaN(+valuationMin) || Number.isNaN(+valuationMax)) {
+    return [Number.isNaN(+valuationMin), Number.isNaN(+valuationMax)]
+  }
+
+  if (+valuationMin >= +valuationMax) {
+    return [false, false]
+  }
+
+  return [+valuationMin >= 0 && +valuationMin <= 1000, +valuationMax >= 0 && +valuationMax <= 1000]
```
chrisling-dev, line 92: both of these seems to be doing almost exactly the same thing except hard coded min max, should we just make it a single `validateRange({max, min, maxCap, minCap})` ?

chrisling-dev (changes requested):

chrisling-dev (commented):
Also should settings card be full width on mobile? Not sure if this would look too small on mobile
Screenshot 2023-04-15 at 9 32 55 AM

packages/prepo-ui/src/components/Icon/icon-components/application/SettingsIcon.tsx
```diff
@@ -10,6 +10,7 @@ const SettingsIcon: React.FC<Props> = ({
 }) => {
   return (
     <svg
+      color={color}
```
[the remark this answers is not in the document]
MeLlamoPablo: Reverted the change, we don't need the settings icon anymore

apps/frontend/core-dapp/src/features/simulator/SimulatorSettings.tsx
MeLlamoPablo: > I think the inputs could just be integer?

I think they should allow decimals

> There's probably some weird behaviour about this, try typing 10. in the valuation input and it just clears the input. Actually just typing "." or "," will empty the input and cause the other input to be inValid

Fixed this by switching to type="text"

chrisling-dev (changes requested):
Estimated Profit seems wrong

Estimated Profit should always be same as Max Profit in max profit scenario for any parameters set. Try changing payout range to anything other than 10 - 90 and the max profit wouldn't be same as estimated profit. For example, if we do 30% - 50%, it means that the cheapest entry would be 0.3 per long token, and max profit would be selling at 0.5 per long token, 66.667% profit, which is correct for Max Profit but wrong for Estimated Profit

chrisling-dev:
> Estimated Profit seems wrong
> 
> Estimated Profit should always be same as Max Profit in max profit scenario for any parameters set. Try changing payout range to anything other than 10 - 90 and the max profit wouldn't be same as estimated profit. For example, if we do 30% - 50%, it means that the cheapest entry would be 0.3 per long token, and max profit would be selling at 0.5 per long token, 66.667% profit, which is correct for Max Profit but wrong for Estimated Profit

Seems to be a state management issue, if you change the valuation range after changing payout range, then drag the slider again, the Estimated Profit becomes correct again, but if you "reset", the Estimated Profit is wrong again?

packages/prepo-ui/src/components/Simulator/Simulator.tsx
```diff
@@ -142,8 +124,16 @@ const Simulator: React.FC<SimulatorProps> = ({
...
+    distance: entryDistance,
+  })
+  const exitPrice = mapDistanceToPrice({
+    data,
+    direction,
+    distance: exitDistance,
+  })
 
   const entryValuation = getValuationForPrice(entryPrice)
   const exitValuation = getValuationForPrice(exitPrice)
```
chrisling-dev, line 139:
```suggestion

  const entryLongPrice = direction === 'long' ? entryPrice : 1 - entryPrice
  const exitLongPrice = direction === 'long' ? exitPrice : 1 - exitPrice
  const entryValuation = getValuationForPrice(entryLongPrice)
  const exitValuation = getValuationForPrice(exitLongPrice)
```

packages/prepo-ui/src/components/Simulator/Simulator.tsx
```diff
@@ -50,84 +52,64 @@ const TrackProgress = styled.div`
...
+  data,
+  direction,
+  distance,
+}: {
+  data: SimulatorDataProps | undefined
+  direction: 'long' | 'short'
+  distance: number | undefined
+}): number | undefined {
+  if (!data || distance === undefined) return undefined
+  const position = direction === 'long' ? distance : 1 - distance
```
chrisling-dev, line 65: probably don't want to do this here because this distance is already "reverted" (e.g. regardless of direction, best entry is always 0)

So if I'm shorting, and simulating with entry at best rate (distance: 0), I want to get payout at payout[0], if you do 1 - distance here I will get payout[1]


```suggestion
  const position = direction === 'long' ? distance : 1 - distance
```

packages/prepo-ui/src/components/Simulator/Simulator.tsx
```diff
@@ -50,84 +52,64 @@ const TrackProgress = styled.div`
   position: absolute;
 `
 
+export function mapDistanceToPrice({
+  data,
+  direction,
+  distance,
+}: {
+  data: SimulatorDataProps | undefined
+  direction: 'long' | 'short'
```
chrisling-dev, line 61: so you can remove the need for direction here and remove everywhere that is passing direction to this function (e.g. including in SimulatorCard)

chrisling-dev (changes requested):
The estimated profit is wrong for Short

I have suggested approach to correct the prices:

`entryPrice` and `exitPrice` in `Simulator.tsx` should always be the actual price to the selected direction, since that helps with more accurate pnl calculation. So we should remove `const position = direction === 'long' ? distance : 1 - distance` in mapDistanceToPrice

Now, removing that will cause valuation shown on handles to be wrong, which is expected because now prices are accurate to their selected direction (e.g. best entry for long is payoutRange[0] and best entry for short is payoutRange[1]), but calculateValuation function takes `longTokenPrice` and specified in the function's parameter

So we'll do another variable to compute long price for both entry and exit

apps/frontend/core-dapp/src/features/simulator/SimulatorCard.tsx
```diff
@@ -81,71 +60,86 @@ const MarketNotAvailable = styled.span`
...
+        entryPrice={mapDistanceToPrice({
+          data,
+          direction,
+          distance: entryDistance,
+        })}
+        exitPrice={mapDistanceToPrice({
+          data,
+          direction,
+          distance: exitDistance,
+        })}
```
chrisling-dev, line 142: After removing `direction` in `mapDistanceToPrice`: 

```suggestion
        entryPrice={mapDistanceToPrice({ data, distance: entryDistance })}
        exitPrice={mapDistanceToPrice({ data, distance: exitDistance })}
```

chrisling-dev (changes requested):
This PR has been understandably, pretty complex to review, so I'm suggesting a few improvements on code readability

packages/prepo-ui/src/components/Simulator/Simulator.tsx
```diff
@@ -50,84 +52,64 @@ const TrackProgress = styled.div`
...
+  onChangeExitDistance,
 }) => {
   const trackRef = useRef<HTMLDivElement>(null)
   const { color } = useTheme()
 
-  // value between 0 - 1 to represent position/offset on slider.
-  const [entryX, setEntryX] = useState<number>()
-  const [exitX, setExitX] = useState<number>()
+  const entryX = direction === 'long' ? entryDistance : 1 - entryDistance
+  const exitX = direction === 'long' ? exitDistance : 1 - exitDistance
```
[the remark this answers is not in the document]
MeLlamoPablo: Great catch! The code is already prepared to handle the case of `entryX` and `exitX` being undefined, so what I need to do is just default to undefined if the distance is undefined.

MeLlamoPablo:
@chrisling-dev thanks for the fantastic feedback! Your suggestions make a lot of sense.

I didn't have the chance to give you an overview of the latest refactor in 4041d87. You were able to understand it well, I just wanted to talk about the rationale behind doing so:

The Simulator, when it was made for the trade page, was designed with the assumption that the valuation and price ranges wouldn't change. With these assumptions, having the source of truth be `entryPrice` and `exitPrice` makes sense.

However, on the simulator page this assumption no longer holds true, which is what caused the bug you detected:

> Estimated Profit seems wrong
> 
> Estimated Profit should always be same as Max Profit in max profit scenario for any parameters set. Try changing payout range to anything other than 10 - 90 and the max profit wouldn't be same as estimated profit. For example, if we do 30% - 50%, it means that the cheapest entry would be 0.3 per long token, and max profit would be selling at 0.5 per long token, 66.667% profit, which is correct for Max Profit but wrong for Estimated Profit

One solution would've been to sync the prices when the price range changes, but this added more complexity. Instead I decided to change the source of truth to be something different. It couldn't have been `entryX` and `exitX` because this wouldn't change when you changed direction, requiring also more syncing. This is why I introduced the "distance" concept, because it is independent of the price range and the direction.

chrisling-dev (changes requested):
build is still failing, could you look into this? @MeLlamoPablo

MeLlamoPablo:
Fixed, sorry, for some reason errors weren't popping up on my IDE

chrisling-dev (approved):
LGTM!

PR #1207, opened 2023-04-07
refactor: condense DepositTradeHelper fees and add multipliers

ramenforbreakfast:
This PR is quite large, mostly due to the fact that I
1. Had to modify 3 different functions for `DepositTradeHelper` as opposed to the hooks, since if I did not modify them to use the new fee structure, it would not compile.
2. We have to write extensive tests for fee and rebate corner cases for all 3 functions, since while they use the same internal function, they use it a different parts and pass different values.
3. Tests specifically for when the rebate is 0 due to rounding were laborious to write and are quite verbose.

Since we for `wrapAndDepositAndTrade` have existing unit tests, I wrote the majority of those tests in the existing unit test suite. For `tradeForCollateral` and `tradeForPosition`, these tests were directly written as integration tests and fee/rebate structure tests have been added and existing ones slightly modified.

So `tests/unit/DepositTradeHelper.test.ts` contains the modified existing unit tests for `wrapDepositAndTrade`
`tests/integration/DepositTradeHelper.test.ts` contains the modified integration tests for `tradeForCollateral` and `tradeForPosition`

I also switched the fee and rebate logic to not use nesting conditionals as per previous PR requests.

Note: You will notice in the `wrapDepositAndTrade` unit test, that I decided to not use `nonZeroTradeParams` anymore, this is because for `wrapAndDepositAndTrade` for fee and rebate tests, we do not care because these are not even real swaps and all we care about is the `Collateral` amount prior to the swap, so we don't need real swap parameters anyway.


apps/smart-contracts/core/contracts/interfaces/IDepositTradeHelper.sol
```diff
@@ -66,10 +64,7 @@ interface IDepositTradeHelper {
 
   function setWstethPoolId(bytes32 wstethPoolId) external;
 
-  function setTradeFeePercent(uint256 tradeFeePercent) external;
-
-  function setDepositAndTradeFeePercent(uint256 depositAndTradeFeePercent)
-    external;
+  function setHelperFeePercent(uint256 helperFeePercent) external;
```
XavierEkkel, line 67:
```suggestion
  function setTradeFeePercent(uint256 tradeFeePercent) external;
```
since it's only for trade and the trade part of deposit+trade, not the wrap/unwrap functions

apps/smart-contracts/core/contracts/interfaces/IDepositTradeHelper.sol
```diff
@@ -81,9 +76,7 @@ interface IDepositTradeHelper {
 
   function getWstethPoolId() external view returns (bytes32);
 
-  function getTradeFeePercent() external view returns (uint256);
-
-  function getDepositAndTradeFeePercent() external view returns (uint256);
+  function getHelperFeePercent() external view returns (uint256);
```
XavierEkkel, line 79:
```suggestion
  function getTradeFeePercent() external view returns (uint256);
```

apps/smart-contracts/core/contracts/interfaces/IDepositTradeHelper.sol
```diff
@@ -27,9 +27,7 @@ interface IDepositTradeHelper {
 
   event WstethPoolIdChange(bytes32 wstethPoolId);
 
-  event TradeFeePercentChange(uint256 percent);
-
-  event DepositAndTradeFeePercentChange(uint256 percent);
+  event HelperFeePercentChange(uint256 percent);
```
XavierEkkel, line 30:
```suggestion
  event TradeFeePercentChange(uint256 percent);
```

apps/smart-contracts/core/test/unit/DepositTradeHelper.test.ts
```diff
@@ -530,9 +467,10 @@ describe('=> DepositTradeHelper', () => {
         expect(core.tokenSender.send).not.called
       })
 
-      it("doesn't call token sender if set and fee = 0", async () => {
+      it("doesn't call token sender if fee amount > 0 and multiplier = 0", async () => {
```
XavierEkkel, line 470:
```suggestion
      it("doesn't call token sender if multiplier = 0", async () => {
```

apps/smart-contracts/core/test/unit/DepositTradeHelper.test.ts
```diff
@@ -546,26 +484,70 @@ describe('=> DepositTradeHelper', () => {
...
+        core.tokenSender.send.returns()
+
+        await depositTradeHelper
+          .connect(user1)
+          .wrapAndDepositAndTrade(user2.address, junkBalancerParams, junkPermit, junkTradeParams)
+
+        expect(core.tokenSender.send).not.called
+      })
+
+      it('calls token sender if fee amount > 0 and scaled fee > 0', async () => {
```
XavierEkkel, line 533:
```suggestion
      it('calls token sender if scaled fee > 0', async () => {
```

apps/smart-contracts/core/test/unit/DepositTradeHelper.test.ts
```diff
@@ -546,26 +484,70 @@ describe('=> DepositTradeHelper', () => {
...
+        expect(await depositTradeHelper.getTokenSender()).not.eq(ZERO_ADDRESS)
+        expect(await depositTradeHelper.getHelperFeePercent()).gt(0)
+        expect(await depositTradeHelper.getAmountMultiplier()).gt(0)
```
XavierEkkel, line 536: add expect that scaled fee > 0

apps/smart-contracts/core/test/unit/DepositTradeHelper.test.ts
```diff
@@ -546,26 +484,70 @@ describe('=> DepositTradeHelper', () => {
...
       })
 
-      it('calls token sender if set and fee > 0', async () => {
-        const nonZeroTradeParams = <IDepositTradeHelper.OffChainTradeParamsStruct>{
-          positionToken: FAKE_POSITION_TOKEN_ADDRESS,
-          deadline: TEST_DEADLINE,
-          amountOutMinimum: parseEther('1'),
-          sqrtPriceLimitX96: parseEther('2'),
-        }
+      it("doesn't call token sender if scaled fee rounds to 0 from multiplier < 1", async () => {
```
XavierEkkel, line 487: the setup is confusing for this test - why not just set an amount of 1 wei and a multiplier of 1 wei as well? then it will definitely round to 0 without needing to think of the setup

XavierEkkel (approved):
LGTM after comments

apps/smart-contracts/core/test/unit/DepositTradeHelper.test.ts
ramenforbreakfast: I'm not sure if the setup would be that much more simpler, because I can't just simply set the "amount" to 1. The amount is the fee from the collateral deposit, so I'd still need to:
- Calculate an amount of Collateral to mint that would result in a fee of 1
- Calculate the corresponding amount of `wstETH` to the contract ahead of time and mint it to the contract

So all that would change is instead of using the exact fee that would result in a rounding of 0 (MULTIPLIER_UINT - 1), we'd replace that with 1 wei, it wouldn't actually reduce lines, maybe a shorter comment explanation from 2 lines to 1 line.

PR #1209, opened 2023-04-08
feat: condense market hooks into one contract

ramenforbreakfast:


apps/smart-contracts/core/contracts/test/TestMarketHook.sol
```diff
@@ -5,7 +5,7 @@ import "../interfaces/IHook.sol";
 import "../interfaces/IPrePOMarket.sol";
 
 /// @notice this contract is a test contract made to test a use case where the redeem hook only takes a part of the fee
```
XavierEkkel, line 7:
```suggestion
/// @notice this contract is a test contract made to test a use case where the market hook only takes a part of the fee
```

XavierEkkel (approved):
LGTM after nit

PR #1217, opened 2023-04-12
feat: pnl estimation for close position

chrisling-dev:


MeLlamoPablo (approved):
Nice!

MeLlamoPablo (changes requested):
Something seems to have broken with the simulator for open position. It doesn't suggest the max value anymore. Could you check it out?

Screenshot 2023-04-12 at 20 33 33

chrisling-dev:
> Something seems to have broken with the simulator for open position. It doesn't suggest the max value anymore. Could you check it out?
> 
> Screenshot 2023-04-12 at 20 33 33

I am not able to reproduce this, could you provide a screen recording on how you got to this state?

MeLlamoPablo (approved):
Hmm weird, yesterday I got that state just by going to the preview URL. I also didn't get that state on main on localhost. So maybe it's unrelated?

Unblocking this, but could you keep an eye on prod after merging?

PR #1218, opened 2023-04-12
feat: use mst-gql generated store for core dapp graph

chrisling-dev:
https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=6b4403e1ecc748a48a63e32a1f35e012&p=a199f2b4d18f4d4887945356f69ba908&pm=s


apps/frontend/core-dapp/src/features/portfolio/PortfolioStore.ts
```diff
@@ -1,12 +1,10 @@
...
 import { RootStore } from '../../stores/RootStore'
-import { Transaction, Position as PositionFromGraph } from '../../types/user.types'
-import { KNOWN_HISTORY_EVENTS, KNOWN_HISTORY_EVENTS_MAP } from '../history/history-utils'
+import { Position as PositionFromGraph } from '../../types/user.types'
 import { PositionEntity } from '../../stores/entities/Position.entity'
-import { supportedContracts } from '../../lib/supported-contracts'
-import { FilterType } from '../../components/Filter/filter.constants'
+import { TransactionModelType } from '../../../generated/mst-gql/core-dapp/TransactionModel'
+
+export type Transaction = Required<Partial<TransactionModelType>>
```
MeLlamoPablo, line 7: Why do we need `Required` over `Partial`? Don't they cancel out?

apps/frontend/core-dapp/src/stores/graphs/CoreGraphStoreV2.ts
```diff
@@ -0,0 +1,85 @@
...
+    })
+
+    makeAutoObservable(this)
+  }
+
+  /**
+   * When any of the filters change, the query is replaced with new query variables.
+   */
+  // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+  get transactionsQuery() {
```
MeLlamoPablo, line 32:
```suggestion
  get transactionsQuery(): ReturnType<typeof QueryStoreCreator.Type['queryTransactions']> {
```

This is not super readable either so feel free to disregard, but I like it better than a rule disable.

Not a blocker.

apps/frontend/core-dapp/src/stores/graphs/CoreGraphStoreV2.ts
```diff
@@ -0,0 +1,85 @@
...
+      },
+    } = this.root.filterStore
+
+    // only query subgraph if all selected actions are valid actions
+    const validatedTypes =
+      selectedFilterTypes?.filter((key) => KNOWN_HISTORY_EVENTS_MAP[key] !== undefined) ?? []
+
+    // for simplicity, UI only shows Opened, Closed, Deposited, Withdrawn in the filter modal
+    // Closed type includes Redeemed transactions
+    if (validatedTypes.includes(FilterType.Closed)) validatedTypes.push(FilterType.Redeemed)
```
MeLlamoPablo, line 48: I feel like this is UI logic that should live as close as possible to the component rather than here. For example, on the component's `onChange` it could pretend that the user has selected "redeemed + closed" when the user actually just selected "close"

Not a blocker, maybe on another PR

apps/frontend/core-dapp/src/stores/graphs/CoreGraphStoreV2.ts
```diff
@@ -0,0 +1,85 @@
...
+      const longTokenAddress = getContractAddress(
+        selectedMarket.long.tokenAddress,
+        network.name,
+        supportedContracts
+      )?.toLowerCase()
+      const shortTokenAddress = getContractAddress(
+        selectedMarket.short.tokenAddress,
+        network.name,
+        supportedContracts
+      )?.toLowerCase()
```
MeLlamoPablo, line 68: What's the point of this if we already know `selectedMarket.long.tokenAddress`? Shouldn't it have the correct address depending on the network?

MeLlamoPablo (changes requested):
Some nits marked with "not a blocker", requesting changes for the rest of the comments

apps/frontend/core-dapp/src/features/portfolio/PortfolioStore.ts
chrisling-dev: It's because of the way TransactionModelType is generated by mst-gql, if you only use Required, the optional values would still be interpreted as "possible to be undefined", wrapping it with Partial is the only workaround I found

apps/frontend/core-dapp/src/stores/graphs/CoreGraphStoreV2.ts
chrisling-dev: ops I meant to merge your PR for "no-return", added this comment only to bypass build for now, removing now

apps/frontend/core-dapp/src/stores/graphs/CoreGraphStoreV2.ts
chrisling-dev: So there's no separate Redeem or Close event in the Filter UI. It's just Open, Close, Deposit, Withdraw. When you select close, we want to also select all Redeem events since they're technically closing a position, kind of.

apps/frontend/core-dapp/src/features/portfolio/PortfolioStore.ts
MeLlamoPablo: > the optional values would still be interpreted as "possible to be undefined"

I'm not sure I understand this. Do you mean that it switches `{ foo: string | undefined }` to `{ foo?: string }`?

apps/frontend/core-dapp/src/stores/graphs/CoreGraphStoreV2.ts
chrisling-dev: Perhaps this code isn't as readable but the `tokenAdress` isn't actually the address, it's the address name which is like "ZKSYNC_LONG_TOKEN" instead of "0x123421"

Screenshot 2023-04-13 at 1 01 23 PM

apps/frontend/core-dapp/src/features/portfolio/PortfolioStore.ts
chrisling-dev: So here's an example, Required doesn't convert `foo: string | undefined` to `foo: string`, it only converts `foo?: string` to `foo: string`.  We need to wrap it with Partial to have Required convert everything to required,

Screenshot 2023-04-13 at 1 03 31 PM

apps/frontend/core-dapp/src/stores/graphs/CoreGraphStoreV2.ts
MeLlamoPablo: I understand. My point is that the UI should handle that instead of the store. For example:

```tsx
<Button
  label="Close"
  enabled={}
  onClick={() => {
    setFilters(prevFilters => {
      if (prevFilters.includes(Filters.Close)) {
       return [...prevFilters, Filters.Close, Filters.Redeem]
      } else {
        return prevFilters.filter(it => it !== Filters.Close || it !== Filters.Redeem)
      }
    })
  }}
/>
```

This way, only correct data gets to the store and we can be sure that we're not accidentally using unvalidated types somewhere else on the store because unvalidated data doesn't even exit the filter buttons component.

apps/frontend/core-dapp/src/stores/graphs/CoreGraphStoreV2.ts
MeLlamoPablo: Oh interesting. Maybe we should consider renaming those to `addressAlias` or something.

Also, why do we need `toLowerCase`? Does the checksum'd address not work for us?

apps/frontend/core-dapp/src/stores/graphs/CoreGraphStoreV2.ts
MeLlamoPablo: I don't think my PR applies here though

apps/frontend/core-dapp/src/stores/graphs/CoreGraphStoreV2.ts
chrisling-dev: The Graph requires all addresses to be lower case, otherwise they won't be able to filter correctly

apps/frontend/core-dapp/src/features/portfolio/PortfolioStore.ts
MeLlamoPablo: Ohh thanks for the explanation. Could we add a helper like:

```ts
export type TrueRequired<T> = Required<Partial<T>>
```

With a comment detailing why is it necessary? Or alternatively try [`Required` from utility-types](https://github.com/piotrwitek/utility-types#requiredt-k) and see if it doesn't have this issue?

chrisling-dev: I'd like to avoid creating another helper for a rare use case like this, we've had a lot of helper functions that we barely use or only use once or twice

as for second suggestion, not sure if it's a good idea to add a package just for this.

I think should just keep it as is unless we find ourselves using this pattern a lot

apps/frontend/core-dapp/src/stores/graphs/CoreGraphStoreV2.ts
chrisling-dev: yea you're right, thought that rule you removed would apply here, added return type here with the actual `Query<{transactions: TransactionModelType}>` since `ReturnType<typeof theFunction>` defeats the purpose of `explicit function return type`, which is to restrict what we return in a function

apps/frontend/core-dapp/src/features/portfolio/PortfolioStore.ts
MeLlamoPablo: We already have this package but fair, I'll resolve this and we can always revisit

MeLlamoPablo (approved):
Approving since blockers got resolved

apps/frontend/core-dapp/src/stores/graphs/CoreGraphStoreV2.ts
chrisling-dev: ahh good point, I'll refactor this separately in another PR https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=6b4403e1ecc748a48a63e32a1f35e012&p=1024e0de3f284a0f9337fd1395cfae68&pm=s

apps/frontend/core-dapp/src/stores/graphs/CoreGraphStoreV2.ts
MeLlamoPablo: > `ReturnType<typeof theFunction>` defeats the purpose of `explicit function return type`, which is to restrict what we return in a function

True that! Just was suggesting a dirty way of getting rid of the warning

PR #1220, opened 2023-04-13
feat: fetch cost basis with mst gql

chrisling-dev:
This PR fetches user's position and costBasis from the subgraph using mst-gql. It performs the fetch on once per address to optimize the number of query calls we send to subgraph.


apps/frontend/core-dapp/src/features/portfolio/PortfolioStore.ts
```diff
@@ -1,18 +1,46 @@
...
+  private syncPositions(): void {
+    // using this reaction guaratantees we only fetch positions once per address
+    reaction(
+      () => this.root.web3Store.address,
+      (address) => {
+        // clean up cost basis data if user changes address to fetch new data
+        this.signerCostBasis = undefined
+        if (!address) return
+
+        this.root.coreGraphStore.positionsQuery(address).then(({ positions }) => {
```
MeLlamoPablo, line 29: Can we use async/await?

apps/frontend/core-dapp/src/features/portfolio/PortfolioStore.ts
```diff
@@ -1,18 +1,46 @@
...
+          if (!positions) return
+
+          runInAction(() => {
+            positions?.forEach((position) => {
+              const { id } = position.longShortToken
+              const { costBasis } = position
+
+              if (!this.signerCostBasis) this.signerCostBasis = {}
+              this.signerCostBasis[id] = costBasis
+            })
```
MeLlamoPablo, line 39: What happens if we open a position after this has fetched? Do we get its cost basis too?

MeLlamoPablo (approved):

apps/frontend/core-dapp/src/features/portfolio/PortfolioStore.ts
chrisling-dev: Yes we can, updated

PR #1221, opened 2023-04-14
feat: migrate close trade input values to CloseTradeStore

chrisling-dev:
Created a new CloseTradeStore and moved all close trade's input related stuff to the new store, moving more in the next few PRs


apps/frontend/core-dapp/src/features/trade/close-trade/CloseTradeStore.ts
```diff
@@ -0,0 +1,56 @@
...
+  constructor(private root: RootStore) {
+    makeAutoObservable(this, {}, { autoBind: true })
+  }
+
+  // selected market and positions, for ease of accessing
+  // so we don't have to keep doing this.root.tradeStore...
+  // and can just use this.selectedMarket
+  get selectedMarket(): MarketEntity | undefined {
+    return this.root.tradeStore.selectedMarket
+  }
```
MeLlamoPablo, line 22:
```suggestion
  private get selectedMarket(): MarketEntity | undefined {
    return this.root.tradeStore.selectedMarket
  }
```

I would remove the comment, I think it's self-explanatory. And maybe private?

apps/frontend/core-dapp/src/features/trade/close-trade/CloseTradeStore.ts
```diff
@@ -0,0 +1,56 @@
...
+    if (this.selectedMarket?.resolved || this.input.type === 'max') {
+      return this.selectedPosition?.totalValueInEth ?? ''
+    }
+
+    return this.input.value
+  }
+
+  get valueBN(): BigNumber | undefined {
+    return this.selectedPosition?.token.parseUnits(this.value)
+  }
```
MeLlamoPablo, line 50:
```suggestion
  get valueInEth(): string {
    if (this.selectedMarket?.resolved || this.input.type === 'max') {
      return this.selectedPosition?.totalValueInEth ?? ''
    }

    return this.input.value
  }

  get valueInEthBN(): BigNumber | undefined {
    return this.selectedPosition?.token.parseUnits(this.value)
  }
```

I think we should try to avoid having amount variables without specifying ETH/wstETH (I know we want to move to something like `value.inWstEth.bn` but still for the meantime)

apps/frontend/core-dapp/src/features/trade/close-trade/CloseTrade.tsx
```diff
@@ -15,24 +15,25 @@ const Wrapper = styled.div`
...
   return (
     <Wrapper>
       <PositionsSlideUp />
       <CurrencyInput
         balance={selectedPosition?.totalValueInEth}
         currency={{
-          icon: tradeStore.paymentToken.iconName,
-          text: tradeStore.paymentToken.shortName ?? tradeStore.paymentToken.name,
+          icon: tokensStore.COLLATERAL.iconName,
+          text: tokensStore.COLLATERAL.shortName,
```
MeLlamoPablo, line 30: 👌

MeLlamoPablo (approved):
LGTM after nits

apps/frontend/core-dapp/src/features/trade/close-trade/CloseTradeStore.ts
chrisling-dev: I would think this is pretty self explanatory tho since the input value is in ETH, but sure if it helps with readability

chrisling-dev: Addressed this in this PR instead https://github.com/prepo-io/prepo-monorepo/pull/1225

apps/frontend/core-dapp/src/features/trade/close-trade/CloseTradeStore.ts
chrisling-dev: addressed this here https://github.com/prepo-io/prepo-monorepo/pull/1224/files#diff-3b910fb66dd8dc572d9efb9c7fb45569eb22c304ccbbe919076a13b4600e8b72L66-L68

PR #1222, opened 2023-04-14
feat: move amount stuff to close trade store

chrisling-dev:
This PR moves some of the amount stuff from trade store to close trade store. I'm also rearranging the order of TradeStore to before of CloseTradeStore (in RootStore) because CloseTradeStore depends on TradeStore so TradeStore has to be initialized first


apps/frontend/core-dapp/src/features/trade/close-trade/CloseTradeStore.ts
```diff
@@ -1,35 +1,81 @@
...
+          this.amountOutBN = output.output
+        })
+      }
+    )
+  }
+
+  // clean up when the tab is left
+  cleanUp(): void {
+    this.amountOutBN = undefined
+    this.setInput('')
```
MeLlamoPablo, line 60:
```suggestion
  // clear state when the tab is left
  clear(): void {
    this.amountOutBN = undefined
    this.setInput('')
```

Super small nit, but `cleanUp` makes me think that the `CloseTradeStore` is being destroyed and not used anymore. `clear` makes me think that its state is being reset. Also `reset` could be a good name.

MeLlamoPablo (approved):

PR #1232, opened 2023-04-25
feat: open trade slippage and fee

chrisling-dev:
https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=1160865fbe4c4639af9d63918c7b7d17&p=6ca14fc5dce84020aeab013c65b467f9&pm=s


apps/frontend/core-dapp/src/features/trade/TradeStore.ts
```diff
@@ -395,27 +382,31 @@ export class TradeStore {
...
+    if (+this.openTradeAmountOut === 0) return this.selectedPosition?.priceInWstEth
 
-    const longTokenPriceAfterSlippage =
-      this.direction === 'long' ? priceAfterSlippage : 1 - priceAfterSlippage
-    return longTokenPriceAfterSlippage
+    // since we must include fees when quoting to get an amount that is tradable, we must
+    // subtract the fees when calculating price to simulate price before fees
+    const price =
+      (+this.openTradeAmountInWstETH - +this.openTradeFeesString) / +this.openTradeAmountOut
+    return this.direction === 'long' ? price : 1 - price
```
MeLlamoPablo, line 400: This is the valuation we show to the user (`openTradePrice` - >`tradingValuation` -> `OpenTradeSummary`)

I don't think we want to include fees here? Otherwise the user will think that slippage is too high

apps/frontend/core-dapp/src/features/trade/TradeStore.ts
```diff
@@ -479,14 +466,22 @@ export class TradeStore {
 
     let error: string | undefined
 
+    const positionTokenAmountOut = this.root.advancedSettingsStore.getAmountAfterSlippageForTrades(
+      this.openTradeAmountOutBN
+    )
```
MeLlamoPablo, line 471: Both `wrapAndDepositAndTrade` and `tradeForPosition` already apply the slippage internally, so this would be duplicating the default slippage.

I also think we want to apply slippage inside the trade functions, not outside, to remove complexity from `TradeStore`.

MeLlamoPablo (changes requested):

apps/frontend/core-dapp/src/features/trade/TradeStore.ts
chrisling-dev: oh yea missed that, thanks

apps/frontend/core-dapp/src/features/trade/open-trade/OpenTradeSummary.tsx
```diff
@@ -50,7 +57,16 @@ const OpenTradeSummary: React.FC = () => {
...
+      <SummaryRecord label="Max Slippage" tooltip={<SlippageProtection />}>
+        -{(openTradeSlippage * 100).toFixed(2)}%
+      </SummaryRecord>
+      <SummaryRecord
+        label="Estimated Fee"
+        tooltip={<EstimatedTradeFee />}
+        loading={openTradeFeesString === undefined}
+      >
+        {displayEth(+(openTradeFeesString ?? '0'))}
+      </SummaryRecord>
```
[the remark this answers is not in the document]
chrisling-dev: Let's keep this for now since we wanna get this out ASAP

apps/frontend/core-dapp/src/features/trade/TradeStore.ts
MeLlamoPablo: I guess what's confusing me is why the fee is being subtracted instead of added? Shouldn't it be added then?

chrisling-dev: no it should be subtracted because the actual amountIn sent to uniswap is inputAmount - fee, not amountIn + fee

The formula is that: price = inputAmount (collateral token) / outputAmount (long short token)

Now assuming we're opening with 1 eth, and the fee is 0.1 eth, the actual amount sent to uniswap pool is 1 subtract 0.1, hence 0.9 eth

If we're adding fee to it, our formula is buying the same amount of LSToken with more collateral and hence incorrect and higher price impact

@MeLlamoPablo

PR #1233, opened 2023-04-25
feat: close trade summary fee and slippage

chrisling-dev:
https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=1160865fbe4c4639af9d63918c7b7d17&p=c8e5591b857348ad97f9050bfaef120d&pm=s


apps/frontend/core-dapp/src/features/trade/close-trade/CloseTradeStore.ts
```diff
@@ -235,17 +236,11 @@ export class CloseTradeStore {
       this.amountOutInWstEthBN === undefined ||
       this.root.closeTradeStore.amountBN === undefined ||
       this.root.closeTradeStore.amountBN.eq(0) ||
-      this.fees === undefined
+      this.feesInWstEth === undefined
```
MeLlamoPablo, line 239: We can remove this, right? Since we don't use the fee anymore.

apps/frontend/core-dapp/src/features/trade/close-trade/CloseTradeStore.ts
```diff
@@ -368,8 +378,11 @@ export class CloseTradeStore {
 
     this.root.closeTradeStore.closing = true
 
+    const collateralAmountOut = this.root.advancedSettingsStore.getAmountAfterSlippageForTrades(
+      this.amountOutInWstEthBN
+    )
```
MeLlamoPablo, line 383: Same as in #1232, we shouldn't apply the slippage here since `tradeForCollateral` already handles that.

MeLlamoPablo (changes requested):
Only the last comment is a blocker, otherwise LGTM

apps/frontend/core-dapp/src/features/trade/close-trade/CloseTradeStore.ts
chrisling-dev: nice catch

MeLlamoPablo (approved):

PR #1235, opened 2023-04-25
feat: update the design of OpenTradeSummary

MeLlamoPablo:
Screenshot 2023-04-25 at 17 42 10

Note: this doesn't add new rows from the design, just groups existing rows.

Some considerations for after merging this:

* On the M1 MacBook Pro 14″ the `Connect Wallet` button is very close to the bottom when the summary is expanded. We may want to consider reducing the top padding to fit it.
* Expanding also triggers a scrollbar. We may want to add a scrollbar gutter to avoid layout jumps. (macOS scrollbars set to always visible)


apps/frontend/core-dapp/src/components/PpoRewardSummaryRecord.tsx
```diff
@@ -17,46 +18,43 @@ const Wrapper = styled.div`
...
+}> = observer(({ displayValueInEth = true, reward }) => {
+  const { ppoTokenStore } = useRootStore()
+
+  const ppoReward = reward === undefined ? undefined : ppoTokenStore.formatUnits(reward.inPpo)
+  const ppoValue = reward === undefined ? undefined : +formatEther(reward.inEth)
+
+  if (ppoReward === undefined) {
+    return <Skeleton height="22px" width="64px" />
+  }
+
```
chrisling-dev, line 33: Not a blocker but I think if you rearrange these, ppoValue won't need the undefined check and in the `displayEth` code below you'll have something cleaner: `displayEth(ppoValue)`

```suggestion
export const PPOReward: FC<{
  displayValueInEth?: boolean
  reward: PpoReward | undefined
}> = observer(({ displayValueInEth = true, reward }) => {
  const { ppoTokenStore } = useRootStore()

  const ppoReward = reward === undefined ? undefined : ppoTokenStore.formatUnits(reward.inPpo)
    if (ppoReward === undefined || reward === undefined) {
    return <Skeleton height="22px" width="64px" />
  }
  
  const ppoValue = +formatEther(reward.inEth)

```

chrisling-dev (commented):
LGTM but will approve after checking whether we're okay with the arrow's direction, I personally think it looks good and make sense

chrisling-dev (approved):
LGTM

apps/frontend/core-dapp/src/components/PpoRewardSummaryRecord.tsx
MeLlamoPablo: Good catch! Definitely cleaner

PR #1237, opened 2023-04-26
fix: infinite loading values on expired markets

MeLlamoPablo:
https://www.notion.so/prepo/hide-estimated-price-and-market-expiry-for-ended-markets-d74a2432e8684d39ba16f36775967b81

This PR fixes some values being stuck in loading for ended markets. In order to be defensive, this is fixed in two ways:

* Previously, you could only show the open trade summary by typing a value on the zkSync market and switching to the Arbitrum market. Now, if the market is resolved, whatever you typed previously will be ignored.
* If the market is already expired, don't show the "Market expiry" row.


apps/frontend/core-dapp/src/features/trade/open-trade/OpenTradeSummary.tsx
```diff
@@ -69,25 +69,27 @@ const OpenTradeSummary: React.FC = () => {
...
-              })} FDV`}
-            />
-          )
-        }
-      >
-        {selectedMarket.expiryTime !== undefined && (
-          <>{getFullDateShortenMonthFromMs(selectedMarket.expiryTime)}</>
-        )}
-      </SummaryRecord>
+      {!selectedMarket.resolved && (
```
chrisling-dev, line 72: maybe better to just add 1 line above the return of summary? since we dont want to show anything in the summary when market has expired, it seems strange to me that we're only doing this check for Market Expiry

`if(selectedMarket.resolved) return null`

chrisling-dev (approved):
LGTM after comment

apps/frontend/core-dapp/src/features/trade/open-trade/OpenTradeSummary.tsx
MeLlamoPablo: Good point! Changed

PR #1240, opened 2023-04-26
feat: display `<0.01` for low PPO rewards

MeLlamoPablo:
https://www.notion.so/prepo/ppo-reward-should-show-0-when-tiny-but-0-46408a7a960e46dc8ee15fc017048ced

Screenshot 2023-04-26 at 13 48 50


apps/frontend/core-dapp/src/components/PpoAmount.tsx
```diff
@@ -24,18 +24,26 @@ const PpoAmount: React.FC<{
...
-    signDisplay,
-  }).formatToParts(+(amount ?? '0'))
-
-  const signCharacter = parts.find(isSign)?.value
-  const amountCharacters = parts
-    .filter(isNotSign)
-    .map(({ value }) => value)
-    .join('')
+  let signCharacter
+  let amountCharacters
```
chrisling-dev, line 28: could we type these or assign a default value?
```suggestion
  let signCharacter = ''
  let amountCharacters = ''
```

chrisling-dev (approved):
LGTM after comment

apps/frontend/core-dapp/src/components/PpoAmount.tsx
MeLlamoPablo: I don't see a reason to do that.

* Typescript is smart and detects that in all `if/else` cases the value is being initialized, so it doesn't add `undefined` to the type.
* Likewise, Typescript can infer that the type of those values will be `string` based on what's being assigned to them.

It would be equivalent to doing:

```ts
const signCharacter = typeof amount === 'string' && +amount < 0.01 && +amount > 0
  ? (signDisplay === 'exceptZero' ? '+' : '')
  : parts.find(isSign)?.value
```

PR #1247, opened 2023-04-27
refactor: remove `wstETH` deposit related code

MeLlamoPablo:
We have a significant amount of dead code that supports depositing `wstETH` directly. This code was never reachable, because even though we have the `DepositCurrencyInput` UI configured, the `currency.onClick` of `CurrencyInput` was `undefined`, so the user could never select `wstETH`.

This PR removes all `wstETH` deposit related code, simplifying our `DepositStore` a ton.

I don't think we will ever want to support depositing `wstETH` directly, but if we do, we can always revert this PR. Meanwhile, we shouldn't have unused code lying around.


apps/frontend/core-dapp/src/stores/CollateralStore.ts
```diff
@@ -32,7 +31,6 @@ export class CollateralStore extends Erc20PermitStore {
       CollateralStore,
       'getDepositFee' | 'getPercentDenominator' | 'getWithdrawFee' | 'percentDenominator'
     >(this, {
-      deposit: action.bound,
       depositFee: computed,
       depositHash: observable,
       depositing: observable,
```
chrisling-dev, line 36: Nice! Could you also delete these things that were used by `deposit` but not anywhere else please?

apps/frontend/core-dapp/src/stores/CollateralStore.ts
```diff
@@ -63,30 +61,6 @@ export class CollateralStore extends Erc20PermitStore {
...
-        error: (error as Error).message,
-      }
-    } finally {
-      runInAction(() => {
-        this.depositing = false
-      })
-    }
-  }
-
   async withdraw(...params: Parameters<Withdraw>): Promise<{ success: boolean; error?: string }> {
```
chrisling-dev, line 64: wanna delete withdraw as well? I don't think it's used anywhere

chrisling-dev (changes requested):
Nice work! Some of the tests are breaking, could you remove any irrelevant tests please

apps/frontend/core-dapp/src/stores/CollateralStore.ts
MeLlamoPablo: Good catch! The IDE didn't warn me they were unused because they were used in `makeObservable`.

chrisling-dev (approved):
LGTM!

PR #1250, opened 2023-05-01
feat: compute wstEth/Eth execution price locally using balancer’s SOR

chrisling-dev:
This is the actual ticket, but just to demo that it works, I also included changes for withdraw ticket here, which only includes very minimal changes: https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=1160865fbe4c4639af9d63918c7b7d17&p=3268844e2be44d8f857b737655cd6031&pm=s

https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=1160865fbe4c4639af9d63918c7b7d17&p=3f30ead890184b279b76097a4b1fef3b&pm=s


apps/frontend/core-dapp/src/stores/BalancerStore.ts
```diff
@@ -21,95 +20,130 @@ const wstEthWethBalancerPoolIdByNetwork: Partial<Record<SupportedNetworks, strin
...
+      if (this.wstEthWethPoolData === undefined || tokens === undefined) return undefined
+
+      const wstEthToken = tokens[0]
+      if (wstEthToken.priceRate === undefined) return undefined
+
+      return exactTokenInForTokenOut({
+        ...this.wstEthWethPoolData,
+        tokenIndexIn: 0,
+        tokenIndexOut: 1,
+        amountIn: wstEthAmount.mul(wstEthToken.priceRate).div(parseEther('1')),
```
MeLlamoPablo, line 82: What is this accomplishing? I don't understand this line, it seems that we're converting before calling the conversion function?

apps/frontend/core-dapp/src/utils/balancer-math.ts
```diff
@@ -0,0 +1,48 @@
...
+export const exactTokenInForTokenOut = ({
+  allBalances,
+  amp,
+  swapFee,
+  tokenIndexIn,
+  tokenIndexOut,
+  amountIn,
+  tokenInDecimals,
+}: ExactTokenInForTokenOutProps): BigNumber => {
+  const outputBN = BalancerSDK.StableMath._calcOutGivenIn(
```
MeLlamoPablo, line 35: The function name is `exactTokenInForTokenOut` but we call `_calcOutGivenIn`. It seems wrong, should we name it `exactTokenOutForTokenIn`? Or simply `exactOutForGivenIn`?

apps/frontend/core-dapp/src/stores/BalancerStore.ts
```diff
@@ -258,6 +292,21 @@ export class BalancerStore {
...
+    if (wethOut === undefined) return undefined
+    return +formatEther(tradeAmount.mul(parseEther('1')).div(wethOut))
+  }
+
+  get wethToWstEthSpotPrice(): number | undefined {
+    const tradeAmount = parseEther('0.001')
+    const wstEthOut = this.getEthAmountInWstEth(tradeAmount)
+    if (wstEthOut === undefined) return undefined
+    return +formatEther(tradeAmount.mul(parseEther('1')).div(wstEthOut))
+  }
```
MeLlamoPablo, line 308:
```suggestion
  private get wstEthToWethSpotPrice(): number | undefined {
    const tradeAmount = parseEther('0.001')
    const wethOut = this.getWstEthAmountInEth(tradeAmount)
    if (wethOut === undefined) return undefined
    return +formatEther(tradeAmount.mul(parseEther('1')).div(wethOut))
  }

  private get wethToWstEthSpotPrice(): number | undefined {
    const tradeAmount = parseEther('0.001')
    const wstEthOut = this.getEthAmountInWstEth(tradeAmount)
    if (wstEthOut === undefined) return undefined
    return +formatEther(tradeAmount.mul(parseEther('1')).div(wstEthOut))
  }
```

apps/frontend/core-dapp/src/stores/BalancerVaultStore.ts
```diff
@@ -0,0 +1,16 @@
...
+import { RootStore } from './RootStore'
+import { SupportedContracts } from '../lib/contract.types'
+import { BalancerVaultAbi, BalancerVaultAbi__factory } from '../../generated/typechain'
+
+export type GetPoolTokens = BalancerVaultAbi['functions']['getPoolTokens']
+
+export class BalancerVaultStore extends ContractStore<RootStore, SupportedContracts> {
+  constructor(root: RootStore) {
+    super(root, 'BALANCER_VAULT', BalancerVaultAbi__factory)
+  }
```
MeLlamoPablo, line 11: Should we make this observable?

apps/frontend/core-dapp/src/stores/entities/MetaStablePool.entity.ts
```diff
@@ -0,0 +1,123 @@
...
+    return this.metaStablePool.getPriceRateCache(this.tokenAddress)?.[0]
+  }
+
+  get balance(): BigNumber | undefined {
+    if (this.metaStablePool.poolTokens === undefined) return undefined
+    return this.metaStablePool.poolTokens.balances[this.index]
+  }
+}
+
+export class MetaStablePool extends ContractStore<RootStore, SupportedContracts> {
```
MeLlamoPablo, line 32: We should name this `MetaStablePoolStore`, as it's a store (it's present on `RootStore`). I think the file being named `MetaStablePool.entity` leads to confusion

apps/frontend/core-dapp/src/stores/entities/MetaStablePool.entity.ts
```diff
@@ -0,0 +1,123 @@
...
+      (poolTokens) => {
+        if (poolTokens === undefined) return
+        const tokens = poolTokens.tokens.map(
+          (address, index) => new MetaStablePoolToken(this, address, index)
+        )
+        runInAction(() => {
+          this.tokens = tokens
+        })
+      }
+    )
```
MeLlamoPablo, line 52: Should we cleanup this reaction like we do on other places? We shouldn't expect the tokens to change, right?

apps/frontend/core-dapp/src/stores/entities/MetaStablePool.entity.ts
```diff
@@ -0,0 +1,123 @@
...
+    if (this.tokens === undefined) return undefined
+
+    const balances = this.tokens.map((token) => {
+      if (token.balance === undefined || token.priceRate === undefined) return undefined
+      const priceRate = token.priceRate.eq(0) ? parseEther('1') : token.priceRate
+      return token.balance.mul(priceRate).div(parseEther('1'))
+    })
+
+    if (balances.includes(undefined)) return undefined
+    return balances as BigNumber[]
```
MeLlamoPablo, line 121:
```suggestion
    const balances: BigNumber[] = []

    for (const token of this.tokens) {
      if (token.balance === undefined || token.priceRate === undefined) return undefined
      const priceRate = token.priceRate.eq(0) ? parseEther('1') : token.priceRate
      balances.push(token.balance.mul(priceRate).div(parseEther('1')))
    }

    return balances
```

Bu using `for..of` we can `return undefined` as soon as we encounter loading values, which is equivalent to what we're doing here. No need to cast `Array<BigNumbrer | undefined>` to `BigNumber[]` and more readable IMO.

apps/frontend/core-dapp/src/stores/entities/MetaStablePool.entity.ts
```diff
@@ -0,0 +1,123 @@
...
+    return {
+      tokens: output[0],
+      balances: output[1],
+      lastChangeBlock: output[2],
+    }
+  }
+
+  // Balancer's calculation is based on data from their subgraph, which converts token with 0 price rate to 1
+  // and then apply the price rate to token balance in the reference below
+  // ref: https://github.com/balancer/balancer-sor/blob/john/v2-package-linear/src/pools/metaStablePool/metaStablePool.ts#L120
```
MeLlamoPablo, line 110:
```suggestion
  // ref: https://github.com/balancer/balancer-sor/blob/9059705624d9be5b20ed00e9cee4cffceca43044/src/pools/metaStablePool/metaStablePool.ts#L120
```

Use the github permalink (the branch could be updated)

MeLlamoPablo (changes requested):
Here's an initial round of questions/feedback. I'll add more comments if I come up with anything else.

apps/frontend/core-dapp/src/stores/BalancerStore.ts
chrisling-dev: this is just part of the formula to get the right output. Unfortunately this wasn't documented anywhere in balancer's docs or repo, so I had to come up with this after tons of trial and error

apps/frontend/core-dapp/package.json
```diff
@@ -22,13 +22,15 @@
...
   "dependencies": {
+    "@georgeroman/balancer-v2-pools": "0.0.14",
     "@lingui/core": "3.14.0",
     "@lingui/macro": "3.14.0",
     "@lingui/react": "3.14.0",
     "@sentry/nextjs": "6.17.6",
     "@uniswap/sdk": "3.0.3",
     "@uniswap/v3-core": "1.0.1",
     "antd": "4.21.6",
+    "bignumber.js": "^9.0.1",
```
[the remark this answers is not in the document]
chrisling-dev: We need this for the types to convert from ethers' BigNumber to OldBigNumber for the balancer-v2-pools to work, see the file `balancer-math.ts`

apps/frontend/core-dapp/src/utils/balancer-math.ts
chrisling-dev: I don't think anything seems wrong here, calcOutGivenIn basically means "compute output token, given an input amount", it's just that they reversed the order

both exactTokenOutForTokenIn and exactOutForGivenIn are wrong

apps/frontend/core-dapp/src/stores/BalancerVaultStore.ts
chrisling-dev: ahh good catch

apps/frontend/core-dapp/src/utils/balancer-math.ts
MeLlamoPablo: The current naming `exactTokenInForTokenOut` suggests to me that what the function does is "compute input token, given output token" which is the opposite of what it does. Maybe we could change it to `exactTokenInToTokenOut`? 

I still think the library name is clearer (`calcOutGivenIn`) see no reason to change it.

apps/frontend/core-dapp/src/stores/entities/MetaStablePool.entity.ts
chrisling-dev: yeap!

apps/frontend/core-dapp/package.json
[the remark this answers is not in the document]
MeLlamoPablo: My point is that you can remove this line from `package.json`. `yarn.lock` doesn't change and you still get the `bignumber.js` dependency on `node_modules` because `@georgeroman/balancer-v2-pools` depends on it.

apps/frontend/core-dapp/src/stores/BalancerStore.ts
MeLlamoPablo: Can we then move these calculations inside `balancer-math.ts`? It feels like all of the complex calculations should be contained inside those functions and not leak to `BalancerStore`.

chrisling-dev: actually I take that back, this code is documented and I just forgot or was a little lost when implementing it, added reference

MeLlamoPablo (approved):
Awesome! Thanks for your hard work on this 💪

PR #1251, opened 2023-05-01
fix: use correct slippage on deposit

MeLlamoPablo:
The `wrapAndDeposit` code is wrong, because it subtracts the fee when calculating `amountOutMinimum`. However the `wrapAndDeposit` SC code does the following:

* Swap all the ETH the user sent to wstETH on Balancer
* Check that the wstETH Balancer gave us is `>= amountOutMinimum`
* Call `Collateral.deposit`
  * This subtracts the deposit fee

`amountOutMinimum` is checked before the fee is applied, however our FE is currently subtracting the fee to the user's deposit to calculate it. **This makes deposits have an effective slippage of ~2.05%**.

This PR fixes that by not subtracting the fee when calculating `amountOutMinimum`. This reduces the slippage tolerance from ~2.05% to 0.05%.

⚠️ Since this decreases the slippage tolerance, **we should monitor deposits after merging to ensure that they don't fail** (which would suggest that the slippage tolerance is too low). I tested this twice and both times it worked.


chrisling-dev (approved):

XavierEkkel (approved):
LGTM

PR #1253, opened 2023-05-01
feat: update the design of WithdrawSummary

MeLlamoPablo:
This PR reworks the design of WithdrawSummary:

* It now uses `OperationSummary`
* It adds a `Fee` row
* It adds a `Slippage Protection` row. This makes it consistent with open/close trade.

  (this wasn't in the design)
* It stops considering slippage from the estimates shown, both in the fee and the received amount.
* The "Gas Cost" from the design is omitted, it will be implemented separately.

Also, this PR removes all overwrites of `SummaryRow`'s color. Since all four summaries are updated to the new design, the `neutral14` color is defined directly on `SummaryRow`.

@chrisling-dev note that this PR doesn't add the "Price impact warning" component from the design for two reasons:

* It can easily be added on another PR to keep this one smaller
* I'm not sure if it's worth to put effort into redesigning this? With the current withdraw cap of ~5 ETH, users will never see this. The price impact warning starts triggering after 135 ETH.


apps/frontend/core-dapp/src/features/withdraw/WithdrawPage.tsx
```diff
@@ -23,9 +23,9 @@ const WithdrawPage: React.FC = () => {
         value={withdrawalAmountInEth}
         showBalance
       />
-      <WithdrawButton />
       <WithdrawWarning />
       <WithdrawSummary />
+      <WithdrawButton />
```
chrisling-dev, line 28: Could we rearrange warning to below of button please

```suggestion
      <WithdrawSummary />
      <WithdrawButton />
      <WithdrawWarning />
```

chrisling-dev (approved):
Nice work! LGTM after comment

apps/frontend/core-dapp/src/features/withdraw/WithdrawPage.tsx
MeLlamoPablo: That wouldn't be consistent with open trade and deposit:

Screenshot 2023-05-02 at 09 19 33

Should we do the same on those?

chrisling-dev: sry missed those, but all deposit limit should now be below buttons, but feel free to change this along with other features in separate PRs

chrisling-dev:
> @chrisling-dev note that this PR doesn't add the "Price impact warning" component from the design for two reasons:
> 
> * It can easily be added on another PR to keep this one smaller
> * I'm not sure if it's worth to put effort into redesigning this? With the current withdraw cap of ~5 ETH, users will never see this. The price impact warning starts triggering after 135 ETH.

1. yeap that's fine
2. we should still add the warning in case of unforeseen circumstances, but don't worry about that for now until I have the new balancer off chain calculation PR because that changes the price impact's calculation

PR #1255, opened 2023-05-01
chore: delete more dead code

MeLlamoPablo:
This PR deletes more code that's not used anywhere


chrisling-dev (approved):
Nice! Did you use any tool to find unused code?

MeLlamoPablo:
> Nice! Did you use any tool to find unused code?

Yeah, I used the "Unused symbol" WebStorm inspection. It's not perfect because you apparently can't filter by directory, and it sometimes reports false positives. For example, this property:

Screenshot 2023-05-02 at 09 17 09

is actually used [here](https://github.com/prepo-io/prepo-monorepo/blob/16dbcbba1e38908b4c3bfd6c788a189707453bba/apps/frontend/core-dapp/src/features/trade/TradeStore.ts#L457).

But for the most part it works great.

PR #1256, opened 2023-05-02
feat: make amount multipliers a mapping

ramenforbreakfast:
All contracts using `TokenSenderCaller` will use the `amountMultiplier` mapped to `msg.sender` (`DepositHook` and `WithdrawHook` will be called by `Collateral` and `MarketHook` will be called by various markets), except for `DepositTradeHelper`, which uses the zero address mapping since users will be interacting with it, but we want the same multiplier to be applied to all users.


apps/smart-contracts/core/test/unit/MarketHook.test.ts
```diff
@@ -105,15 +105,15 @@ describe('=> MarketHook', () => {
...
+      await expect(
+        marketHook.connect(user).setAmountMultiplier(deployer.address, TEST_MULTIPLIER)
+      ).revertedWith('Ownable: caller is not the owner')
     })
 
     it('succeeds if owner', async () => {
       expect(await marketHook.owner()).eq(deployer.address)
 
-      await marketHook.connect(deployer).setAmountMultiplier(TEST_MULTIPLIER)
+      await marketHook.connect(deployer).setAmountMultiplier(deployer.address, TEST_MULTIPLIER)
```
XavierEkkel, line 116:
```suggestion
      await marketHook.connect(deployer).setAmountMultiplier(marketSigner.address, TEST_MULTIPLIER)
```

XavierEkkel (approved):
LGTM after comment

apps/smart-contracts/core/test/unit/MarketHook.test.ts
ramenforbreakfast: In this test, we aren't doing it because this is simply a setter testcase and there is no market to extract a signer from. I think it'd be a little overkill to instantiate a market just so we could set the multiplier to an entity that reflects what'd be done in production.

What i will do is add a comment clarifying that the hook would be setting multipliers for a market.

PR #1257, opened 2023-05-02
feat: add events to AccountList

ramenforbreakfast:
This PR adds events to the AccountList contract to make it easier to see write history to an AccountList.
- I decided to add an `AccountListReset` event, and have it contain the timestamp when it happened. I did not store the accounts added to the reset list in `AccountListReset` since I felt that information should remain in the `AccountListChange` event. Since it is assumed all accounts passed into `reset` are being included in the new list, the array of boolean values indicating reset status is not passed in and needs to be constructed inside `reset`.
- There is a testcase showing the situation where there is a duplicate in `accounts` passed to `setAccountList`, but any previous values will be overwritten by the latter most corresponding boolean in `included`. In this case, the event will contain the passed in arrays with the repeated account as well as the multiple boolean values. I felt that it was unnecessary, to perform extra processing to only show the final value, and just emitted the unaltered arrays passed in (same thing we do when actually setting the list).


apps/smart-contracts/core/contracts/AccountList.sol
```diff
@@ -24,6 +24,7 @@ contract AccountList is IAccountList, SafeOwnable {
         ++i;
       }
     }
+    emit AccountListChange(accounts, included);
   }
 
   function reset(address[] calldata includedAccounts)
```
XavierEkkel, line 30: lets change this to only be reset (no set logic)

ramenforbreakfast:
This has been updated with recently merged PRs and with atomic setting of the list after `reset` removed.

packages/prepo-shared-contracts/contracts/interfaces/IAccountList.sol
```diff
@@ -5,6 +5,10 @@ pragma solidity =0.8.7;
  * @notice Stores whether an address is included in a set.
  */
 interface IAccountList {
+  event AccountListChange(address[] accounts, bool[] included);
+
+  event AccountListReset(uint256 timestamp);
```
XavierEkkel, line 10: remove timestamp from event

packages/prepo-shared-contracts/test/AccountList.test.ts
```diff
@@ -65,103 +56,127 @@ describe('=> AccountList', () => {
...
+    it('includes single account', async () => {
+      expect(await accountList.isIncluded(includedUser1.address)).eq(false)
 
-      await accountList.connect(owner).set([includedUser1.address], [true])
+      const tx = await accountList.connect(owner).set([includedUser1.address], [true])
 
-      expect(await accountList.isIncluded(includedUser1.address)).to.eq(true)
+      expect(await accountList.isIncluded(includedUser1.address)).eq(true)
+      await expect(tx)
+        .to.emit(accountList, 'AccountListChange')
```
XavierEkkel, line 81: nit:
```suggestion
        .emit(accountList, 'AccountListChange')
```

XavierEkkel (approved):
LGTM after comments

XavierEkkel (approved):
LGTM after comments

PR #1258, opened 2023-05-02
feat: trade with instant wstETH conversion

chrisling-dev:
UX wise, no changes expected (except there's now less chances of getting rate limited). Whatever trade size you get on app.prepo.io, you should get the exact same price with these changes

https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=1160865fbe4c4639af9d63918c7b7d17&p=2f57b89913cc4c48a6f6b74e8034a160&pm=s


MeLlamoPablo (approved):
Awesome! I love how easy it is to query a price with the new system :)

PR #1262, opened 2023-05-02
feat: make entry fixed for open/close trade on simulator

MeLlamoPablo:
Screenshot 2023-05-02 at 18 43 02

This PR updates the simulator on the trade page to have a fixed entry handle. The way this works is:

* The `<Simulator />` component renders a fixed handle if the corresponding `onChange` is `undefined`
* `<SimulatorCard />` now accepts either a `FormValue<number | undefined>` or a `number | undefined` for both of the entry and exit distances.

  If the caller sends a raw value rather than a `FormValue`, `<SimulatorCard />` doesn't send `onChange` to `<Simulator />` for that value (because it doesn't have an `onChange`)
* `<TradeSimulator />` now doesn't need to call an useless `useFormValue` and passes `entryDistance` directly instead.


packages/prepo-ui/src/components/Simulator/Handle.tsx
```diff
@@ -98,30 +98,46 @@ const TextWrapper = styled.div<{ $color: keyof Color; $isEntry?: boolean }>`
...
+const Handle: React.FC<Props> = ({
+  color,
+  disabled: forceDisable,
+  onChange,
+  value,
+  isDirty,
+  isEntry,
+  x,
+}) => {
+  const disabled = forceDisable || !onChange
```
chrisling-dev, line 110: I've noticed this pattern multiple times where we rename the prop, instead of that, maybe this looks cleaner, but don't bother changing here cus it's not a big deal:

```suggestion
  disabled,
  onChange,
  value,
  isDirty,
  isEntry,
  x,
}) => {
  const _disabled = disabled || !onChange
```

packages/prepo-ui/src/components/Simulator/Handle.tsx
```diff
@@ -29,11 +29,11 @@ const Value = styled.p`
   font-weight: ${({ theme }): number => theme.fontWeight.semiBold};
 `
 
-const Pin = styled.div<{ $color: keyof Color }>`
+const Pin = styled.div<{ $color: keyof Color; $tall: boolean }>`
   background-color: ${({ $color, theme }): string => theme.color[$color]};
   border: solid 2px ${({ theme }): string => theme.color.purpleStroke};
   border-radius: ${spacingIncrement(10)};
-  height: ${spacingIncrement(12)};
+  height: ${({ $tall }) => spacingIncrement($tall ? 20 : 12)};
```
chrisling-dev, line 36: I think you can keep this code simpler by just assigning height: ${spacingIncrement(20)} since the doughnut will be on top of the pin anyway (for draggable handles)

packages/prepo-ui/src/components/Simulator/Handle.tsx
```diff
@@ -98,30 +98,46 @@ const TextWrapper = styled.div<{ $color: keyof Color; $isEntry?: boolean }>`
...
+      axis="x"
+      bounds="parent"
+      disabled={disabled}
+      position={{ x, y: 0 }}
+      onDrag={(_, data): void => onChange?.(data.x)}
+    >
+      {/* Wrap with an extra absolute div so the direct child to track has 0 width, and hence exactly precise calculation */}
+      <div style={{ position: 'absolute', height: '100%' }}>
+        <Wrapper $isEntry={isEntry} $disabled={disabled}>
+          <Pin $color={color} $tall={!onChange} />
```
chrisling-dev, line 123: Then can remove this:
```suggestion
          <Pin $color={color} />
```

chrisling-dev (approved):
Only minor comments

packages/prepo-ui/src/components/Simulator/Handle.tsx
MeLlamoPablo: Good idea!

chrisling-dev (approved):
LGTM!

PR #1264, opened 2023-05-02
feat: remove unnecessary approvals made by market

ramenforbreakfast:
I tried to perform this change in the most concise way possible that preserves/adds the following behavior:
- Revert when the fee percentage is > 0 but the calculated fee amount is 0
- Revert if the redeemed collateral amount is 0

If neither are true then:

- If the fee percentage is 0 _and_ the hook is set, call the hook but don't make approvals or perform calculations around the fee taken. This is because we shouldn't assume hooks are only called if a fee is taken.
- If the fee is > 0 _and_ the hook is set, call the hook with the approval logic
- Else (if the hook is not configured, but neither of the two revert cases apply), just report a zero fee and don't call the hook

Naming of variables has been changed to be more consistent. There are places where, to make it clear we are not referring to a fee percentage, but an amount, we use the word `amount`, but there were other instances where we forewent using the `amount` keyword. I have changed multiple variable names to ensure we use `amount` when we are referring to an amount rather than a percentage.

Edit: fee processing has been moved to an internal function to share with minting.


apps/smart-contracts/core/test/unit/PrePOMarket.test.ts
```diff
@@ -864,12 +864,23 @@ describe('=> prePOMarket', () => {
       expect(redeemHook.hook).not.called
     })
 
+    it('ignores hook if not set and fee = 0', async () => {
```
XavierEkkel, line 867: dont need this test as its adequately tested by the test above

XavierEkkel (approved):
LGTM after nit

PR #1265, opened 2023-05-03
feat: allow collateral permits for minting

ramenforbreakfast:
To allow for collateral permits, I decided to split minting into two separate functions. The rationale for this is that some contracts might want to mint positions (e.g. ArbitrageBroker) and contracts wouldn't process permits on their own behalf. Thus the simplest solution was to override the function and create a separate `mint` function with an additional permit parameter. This also means existing contracts and tests do not have to modified further to accomodate this change.


ramenforbreakfast:
Closing, resubmitted in PR #1301 due to `mint` having diverged significantly

PR #1267, opened 2023-05-03
feat: remove roles and validity lists from factory

ramenforbreakfast:
This removes the validity lists for both valid collateral and markets. Since these lists no longer exist or will be externalized in the market validity case, there is no need for access controls on the contract anymore.


XavierEkkel (approved):
LGTM

PR #1269, opened 2023-05-03
fix: add missing uniswap fee for trade and depositAndTrade

MeLlamoPablo:


apps/frontend/core-dapp/src/features/trade/TradeStore.ts
```diff
@@ -21,6 +22,8 @@ type SlideUpContent = 'OpenMarket' | 'OpenCurrency' | 'ClosePosition' | 'CloseCu
 const DEFAULT_DIRECTION = 'long'
 
 export class TradeStore {
+  private static readonly UNISWAP_FEE = fraction(BigNumber.from(1), BigNumber.from(100))
```
chrisling-dev, line 25: Could we read this from the pool's contract instead? The pool has a `fee` value we can use

chrisling-dev (approved):
LGTM

https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=1160865fbe4c4639af9d63918c7b7d17&p=84e01d30b0814c8289307422c20f003a&pm=s

PR #1271, opened 2023-05-03
fix: add missing uniswap fee for close trade

MeLlamoPablo:
This PR properly considers the Uniswap LP fee in close trade.

It is slightly different to #1269, because:

* For open trade, the prePO fee is deducted before the Uniswap fee. So, in order to calculate the Uniswap fee we simply apply it to the amount after the prePO fee.
* However, for close trade the Uniswap fee is deducted first, and then the prePO fee is applied to the collateral returned by Uniswap.

  The Uniswap quote already comes with its own fee applied. In order to create the `FeeStore`, we obtain the quoted amount before the fee by dividing it by `1 - fee`.


apps/frontend/core-dapp/src/features/trade/close-trade/CloseTradeStore.ts
```diff
@@ -278,33 +280,69 @@ export class CloseTradeStore {
...
+      )
     }
 
+    return new FeeStore({
+      // When quoting, Uniswap gives us the amount after fee. To calculate the
+      // correct fee we have to reverse it and then re-apply it.
+      applicableToWstEth: () => reverseFee(this.amountOutInWstEthBN),
+      fee,
+      tokenSender: undefined,
+    })
```
chrisling-dev, line 306: I think writing the function in applicableToWstEth is more readable and cleaner since we don't have to refer back to `reverseFee` to check again the applicableToWstEth usage? wdyt?

```suggestion
    const fee = selectedMarket.resolved
      ? fraction(BigNumber.from(0), BigNumber.from(1))
      : TradeStore.UNISWAP_FEE

    return new FeeStore({
      applicableToWstEth: () => {
        if (this.amountOutInWstEthBN === undefined) return undefined

        // Get amount before fee from output of Uniswap's quote, which is after fee.
        return div(
          this.amountOutInWstEthBN,
          sub(fraction(BigNumber.from(1), BigNumber.from(1)), TradeStore.UNISWAP_FEE)
        )
      },
      fee,
      tokenSender: undefined,
    })
```

apps/frontend/core-dapp/src/features/trade/close-trade/CloseTradeStore.ts
```diff
@@ -278,33 +280,69 @@ export class CloseTradeStore {
...
+  }
+
+  private get tradeFeeStore(): FeeStore | undefined {
+    const { selectedMarket } = this
+    if (selectedMarket === undefined) return undefined
```
chrisling-dev, line 311: It's not obvious why selectedMarket needs to be checked here since it's not used in this function, better to just check for uniswapFeeStore? Also if we're going to check selectedMarket, we dont have to destructure it because the function is not using many things from `this`

```suggestion
    if(this.uniswapFeeStore === undefined) return undefined
```

OR this if we're keeping the check for selectedMarket

```suggestion
    if(this.selectedMarket === undefined) return undefined
```

chrisling-dev (changes requested):
Minor comments

apps/frontend/core-dapp/src/features/trade/close-trade/CloseTradeStore.ts
MeLlamoPablo: Sure, I actually did it this way first but then switched back because I thought it'd improve readability haha but I have no strong opinion

chrisling-dev: thanks!

apps/frontend/core-dapp/src/features/trade/close-trade/CloseTradeStore.ts
MeLlamoPablo: This was just an oversight, there's no need to check for `selectedMarket` (there's also no need to check for `uniswapFeeStore` since we're using optional chaining. Thanks for catching!

chrisling-dev (approved):
LGTM

PR #1272, opened 2023-05-03
refactor: dynamically retrieve the Uniswap LP fee

MeLlamoPablo:
This is a follow up to #1269 and #1271 which retrieves the fee dynamically from the Uniswap pool instead of hardcoding it.


chrisling-dev (approved):
LGTM

PR #1277, opened 2023-05-03
feat: remove unnecessary burning when redeeming only one token

ramenforbreakfast:
Edit: updated with changes from #1264


apps/smart-contracts/core/test/unit/PrePOMarket.test.ts
```diff
@@ -739,6 +739,26 @@ describe('=> prePOMarket', () => {
       expect(await collateralToken.balanceOf(user.address)).to.eq(totalOwed)
     })
 
+    it('does not burn short tokens if only long tokens are redeemed', async () => {
```
XavierEkkel, line 742: nit:
```suggestion
    it('doesn't burn short tokens if only long tokens redeemed', async () => {
```
same for below

XavierEkkel (approved):
LGTM after nit

PR #1278, opened 2023-05-03
feat: separate deployer and governance roles and assign them at the start

ramenforbreakfast:
For permissionless markets, we need to ensure all roles besides settlement should still remain with governance, and assign the ability to settle a market to the deployer, resulting in the following changes:
- `DEFAULT_ADMIN_ROLE` is no longer assigned to anyone, as that would allow an account effective control over all roles. Since it is assigned by default to the original deployer, we just revoke it from the factory (since the factory is the actual deployer) and never reassign it.
- Since there is no convenient `DEFAULT_ADMIN_ROLE` that has control over all roles, we must now assign a `roleAdmin` to each role. For convenience, I have assigned the role itself as its own admin. For context, a `roleAdmin` is a `roleY` that has the ability to assign people to `roleX`. We want to assign a `roleAdmin` so that in the future, we can assign a different account to the roles we have control over.
- Identified a cosmetic bug while modifying the parameters to accomodate a separate `deployer` and `governance`. The `IPrePOMarketFactory` `governance` and `collateral` parameters were not in the correct order, but since they are both `address` types, this did not actually effect anything.


XavierEkkel (approved):
LGTM

XavierEkkel (approved):
LGTM

PR #1285, opened 2023-05-04
feat: hide reset button on open trade simulator

MeLlamoPablo:
https://www.notion.so/prepo/Simulator-Reset-Button-32ffb72e1bf44c1197ee6099d7ba0c6f

This PR hides the "Reset" button on the open trade simulator and the corresponding yellow dot on the exit handle if the user modifies its value, since the default value is the max profit scenario and that is not meaningful.

The way it is implemented is by pretending that the form value is never dirty. This way, the Simulator won't know that the user modified the value and won't offer the possibility to reset it.


chrisling-dev (approved):
Awesome! Nice and clean solution

PR #1288, opened 2023-05-04
chore: delete unused `TradeSettings`

MeLlamoPablo:
This PR deletes the unused `TradeSettings` component and all related code. I'm not sure if we want to restore it at some point, but it's been hidden a while and we can always revert this PR if necessary.


chrisling-dev (approved):
LGTM

PR #1289, opened 2023-05-05
feat: add recipient to minting function

ramenforbreakfast:
Some changes that might need context:
- Moved `longToken` and `shortToken` initialization up to the before block since there is no need to do it multiple times.
- The `user` and `recipient` naming here for testcases is confusing in this testcase. I think this was done a long time ago or at some point during refactoring and I would like to change it, but doing so in this PR will cause headaches due to it affecting all testcases in this file, and it is already being modified in a parallel train.


XavierEkkel (approved):
LGTM after nit

PR #1290, opened 2023-05-05
feat: add fees to minting function

ramenforbreakfast:
Adding fees to minting required the addition of many testcases and some refactoring around the existing mint tests.
**Note:** Fee bypasses will be introduced in a later PR.
Major changes include:
- In addition to using the `_processFee` function to handle minting fees, some variables were renamed to maintain consistency and to identify whether an amount is referring to Collateral or LongShortTokens.
- Using a mock actual `MarketHook` for minting since we need a functional hook to take away the approved fee
- Using the special `TestMarketHook` that partially takes a fee to test this situation, similar to `redeem` tests.
- Added revert cases around fees and several approval testcases similar to `redeem` tests.
- Approval between the user and market is set to infinity at the beginning of the test. This is because if an approval between two entities is not infinite, the `Approval` event is emitted during any transfer, and thus it is hard to detect whether an approval was actually made or if a transfer simply happened. Because of this, a testcase for insufficient approval was also added as a precaution.


apps/smart-contracts/core/contracts/PrePOMarket.sol
```diff
@@ -89,25 +89,28 @@ contract PrePOMarket is
...
-    emit Mint(msg.sender, recipient, amount);
-    return amount;
+    _collateral.transferFrom(msg.sender, address(this), collateralAmount);
+    uint256 actualCollateralFeeAmount = _processFee(
+      collateralAmount,
+      _mintingFeePercent,
+      _mintHook,
+      recipient
+    );
+    longShortAmountMinted = collateralAmount - actualCollateralFeeAmount;
```
XavierEkkel, line 110:
```suggestion
    longShortAmount = collateralAmount - actualCollateralFeeAmount;
```

apps/smart-contracts/core/test/unit/PrePOMarket.test.ts
```diff
@@ -460,122 +462,286 @@ describe('=> prePOMarket', () => {
...
+        prePOMarket.connect(user).mint(userCollateralBefore, recipient.address)
+      ).revertedWith('ERC20: insufficient allowance')
+    })
+
+    it('transfers collateral from funder if funder = recipient', async () => {
+      const funderCollateralBalanceBefore = await collateralToken.balanceOf(user.address)
+
+      await prePOMarket.connect(user).mint(TEST_MINT_AMOUNT, user.address)
+
+      expect(await collateralToken.balanceOf(prePOMarket.address)).eq(expectedLongShortAmount)
```
XavierEkkel, line 558: also check market balance before (same in other tests)

apps/smart-contracts/core/test/unit/PrePOMarket.test.ts
```diff
@@ -460,122 +462,286 @@ describe('=> prePOMarket', () => {
...
+      await expect(tx)
+        .emit(collateralToken, 'Approval')
+        .withArgs(prePOMarket.address, mintHook.address, expectedMintingFeeAmount)
+      await expect(tx)
+        .emit(collateralToken, 'Approval')
+        .withArgs(prePOMarket.address, mintHook.address, 0)
+      expect(await collateralToken.allowance(prePOMarket.address, mintHook.address)).eq(0)
+    })
+
+    it("doesn't approve fee if hook not set", async () => {
```
XavierEkkel, line 599: this + the one below should go above the 'approves fee for hook to use' test

apps/smart-contracts/core/test/unit/PrePOMarket.test.ts
```diff
@@ -460,122 +462,286 @@ describe('=> prePOMarket', () => {
...
       expect(mintHook.hook).calledWith(
         user.address,
         recipient.address,
         TEST_MINT_AMOUNT,
-        TEST_MINT_AMOUNT
+        expectedLongShortAmount
       )
     })
 
     it('ignores hook if not set', async () => {
```
XavierEkkel, line 727:
```suggestion
    it('ignores hook if not set and fee percent > 0%', async () => {
```
(and add check that fee percent > 0%)

so that we cover both the =0% and >0% cases

XavierEkkel (approved):
LGTM after nits

PR #1296, opened 2023-05-05
chore: remove unused field from `MarketEntity`

MeLlamoPablo:


chrisling-dev (commented):
Let's keep these for now, we're very likely to be using this again this week or next week

PR #1297, opened 2023-05-05
fix: cost basis calculation to be before fee

MeLlamoPablo:
The current `costBasis` from subgraph includes the cost basis before prePO trade fee but after Uniswap fee. For consistency, we can remove the Uniswap fee so that it is before all kind of fees.


MeLlamoPablo:
Superseded by #1315

PR #1302, opened 2023-05-06
feat: generalized value beacon

ramenforbreakfast:
This is a generalized beacon for storing integer values that entities can refer to. This will be initially be used for storing market fees (e.g. a key could be `MarketMintingFee`), but can be redeployed or reused for other purposes.


packages/prepo-shared-contracts/contracts/interfaces/IValueBeacon.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: AGPL-3.0
+pragma solidity =0.8.7;
+
+interface IValueBeacon {
+  event ValueChange(bytes32 key, uint256 value);
+
+  function set(bytes32 key, uint256 value) external;
+
+  function get(bytes32 key) external view returns (uint256);
```
XavierEkkel, line 9:
```suggestion
  function get(bytes32 key) external view returns (uint256 value);
```
(my new proposed convention in interface - but not needed in implementation unless actually used)

XavierEkkel (approved):
LGTM

PR #1303, opened 2023-05-06
feat: generalized address beacon

ramenforbreakfast:
Same as `ValueBeacon`, but for addresses. Will initially be used for storing reference hooks that markets will use by default. Could also be used as a centralized reference for governance.


XavierEkkel (approved):
LGTM

PR #1304, opened 2023-05-08
feat: balancer calculation return fallback value with spot price

chrisling-dev:
https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=1160865fbe4c4639af9d63918c7b7d17&p=0e7fa096fd984f018336d990984dd9d8&pm=s


apps/frontend/core-dapp/src/stores/BalancerStore.ts
```diff
@@ -53,6 +52,25 @@ export class BalancerStore {
...
+      tokenIndexIn: 0,
+      tokenIndexOut: 1,
+      amount,
+    })
+
+    // calculate rate for 1 wstETH
+    return outputOut.mul(parseEther('1')).div(amount)
+  }
+
+  get defaultWethToWstEthRate(): BigNumber | undefined {
```
MeLlamoPablo, line 69:
```suggestion
  get defaultEthToWstEthRate(): BigNumber | undefined {
```

I don't think we refer to ETH as WETH anywhere else, right?

MeLlamoPablo (approved):

apps/frontend/core-dapp/src/stores/BalancerStore.ts
chrisling-dev: ahh probably have to fix this up separately, I did find a few things that were using Weth, but will have a separate PR to refactor after we fix this issue

PR #1308, opened 2023-05-08
feat: get Lido APR from chain

MeLlamoPablo:
In order to move towards decentralized front-ends, we can't have any server side logic. Unfortunately, we can't use the `https://stake.lido.fi/api/sma-steth-apr` API endpoint on the client because it doesn't have an `Access-Control-Allow-Origin` header.

Instead, this PR pulls the data from the Lido Oracle on Ethereum mainnet. The data is not available on Arbitrum, but thankfully it only updates once per day, so we don't have to subscribe to on-chain changes. This simplifies the process since we only request the data at startup and that's it.

Note that the APR displayed is different from the one on production (which is the one on [Lido's app](https://stake.lido.fi/)) because Lido's API returns the 7-day moving average whereas the APR returned from the Oracle corresponds to the last 24 hours. If we wanted a moving average ourselves we'd have to create a subgraph that indexes this information.


chrisling-dev (approved):
LGTM! only one minor comment

PR #1309, opened 2023-05-08
chore: replace server-side redirects with client-side redirects

MeLlamoPablo:
In order to move towards decentralized front-ends, we can't have any server side logic.

This PR moves all of our redirects from the server to the client.

We can also investigate [IPFS Redirects](https://docs.ipfs.tech/how-to/websites-on-ipfs/redirects-and-custom-404s), but in order to transition incrementally to decentralized front-ends we can start by using Next's client-side redirects which will work with both centralized and decentralized front-ends.


apps/frontend/core-dapp/src/pages/index.tsx
```diff
@@ -1,13 +1,13 @@
-import { GetServerSideProps } from 'next'
 import { Routes } from '../lib/routes'
+import { Redirect } from '../components/Redirect'
+import TradePage from '../features/trade/TradePage'
 
-export default (): null => null
+const WithdrawPage: React.FC = () => (
```
chrisling-dev, line 5: Did you mean 
```suggestion
const TradePage: React.FC = () => (
```

chrisling-dev (approved):
is it just me or the new `/` route is much faster haha, nice work! LGTM after fixing component name in /pages/index.tsx

MeLlamoPablo:
> is it just me or the new `/` route is much faster

Theoretically server-side redirects should be faster, so maybe Next's server wasn't as fast as it should've been. I'm also cheating by rendering `<TradePage />` both in `/` and `/trade`, so you actually see the trade page before the redirect.

PR #1310, opened 2023-05-08
feat: add `EtherAmount`

MeLlamoPablo:
This PR introduces an abstract class `BigAmount` and one children: `EtherAmount`. In the future, `Erc20Amount`, another children, can be implemented. They are intended to become the primary way we operate with amount values to improve code consistency, safety and maintainability.

## Definition

An amount is a numeric value representing the quantity of a given token.

* `2 ETH` is an amount, represented by `EtherAmount`
* `0.5 zkSync Long Token` is an amount, represented by `Erc20Amount`
* `2% deposit fee` is NOT an amount because it does not refer to a token quantity.
* `1.1 ETH per wstETH` is NOT an amount either.

## Motivation and design goals

* Unify operating with ETH and wstETH.

  We operate with `wstETH` conversion frequently within our codebase, resulting in varaibles named `amountInEth` and `amountInWstEth`, increasing code verbosity and having to handle many loading states. We convert back and forth too many times.

  `EtherAmount` is designed to abstract this and work with both units. You can instantiate it with `.fromEth` and `.fromWstEth`. You can operate with them seamlessly, even if they are different in denomination:

  ```ts
  const a = EtherAmount.fromEth(depositAmount, this.root)
  const b = EtherAmount.fromWstEth(depositFee, this.root)

  const balanceIncrease = depositAmount.sub(depositFee)
  ```

* Unify operating with `BigNumber` and `string`

  Across the codebase we use `BigNumber` and the "formatted" `string` representation. Operating with `BigNumber` is precise, but the formatted string represents "token units" which is more convenient.

  This leads to confusion, and variables named `amountBN` and `amount/amountString/amountFormat`.

  A design goal is to combine both of these properties. `BigAmount`s have information about the decimals of the token they represent, and so we can operate with them using `BigNumber` precision but still be able to format as a string whenever we need to show the value in the UI.

  We should only need to convert a `BigAmount` to a string when showing information in the UI.

* Improve type safety.

  Since `BigAmount`s refer to specific tokens, we can increase the type safety of our operations:

  * Adding `EtherAmount + EtherAmount` makes sense and is allowed
  * Adding `EtherAmount + Erc20Amount` doesn't makes sense and is not allowed
  * Multiplying `EtherAmount * BigNumber` makes sense and is allowed
  * Multiplying `EtherAmount * EtherAmount` doesn't make sense and isn't allowed

* Abstract away `BigNumber`

  The goal is to completely wrap `BigNumber` so that we don't have to interact with it directly. Theoretically, we could even switch the underlying `BigNumber` library if we wanted and it would still work.

  This allows us to easily extend its functionality. For example, instead of throwing errors on division-by-zero, we safely default to 0 and log an error to the console.

  The only two cases where we should need to unwrap to `BigNumber` are:

  * Legacy code compatibility (we can't migrate all the codebase at once)
  * Sending `BigNumber` to the blockchain
    * In the future, we could add a ESLint rule to ensure that `inEthBN`/`inWstEthBN` only get called inside `sendTransaction`

* Batteries-included operations

  Currently, whenever we want to operate with amounts, we sometimes need to get additional information from the stores: call `BalancerStore` for conversion and call the token store for the amount of decimals.

  `EhterAmount`/`Erc20Amount` are designed to self-contain all of their dependencies. After you instantiate them, you don't need to use any external store to operate or display data:

  ```ts
  // In the store
  const amount = EtherAmount.fromWstEth(BigNumber.from(100_000), this.root)
  const half = amount.div(BigNumber.from(2))
  
  // In the UI
  half.toFormattedEthString() // Converts wstETH -> ETH and formats at the same time.
  ```


chrisling-dev (commented):
Had a re-review and am not convinced that this PR will save us time:
- The BigAmount only has one special function, the div function, compared to BigNumber, which imo can be done much more easily with a `safeDiv` util function, (because changing every BigNumber to use BigAmount is a lot of work, but applying a `safeDiv` function to division isn't)
- The EtherAmount class seems a little overcomplicated, and passing this.root everytime isn't very intuitive, since the conversion is only directly relevant to BalancerStore and not root store nor any other stores, I think it's better to just create a function from Balancer that simply returns the right value

I've suggested an alternative to achieve the same goal in a separate PR https://github.com/prepo-io/prepo-monorepo/pull/1335, note that I'm not addressing the safe div operation in my PR because that should be a separate PR to address

PR #1311, opened 2023-05-08
feat: copy over beacon contracts and add setters

ramenforbreakfast:
To instantiate the `AddressBeacon` and `ValueBeacon` contracts for testing, we need to copy them over to the core folder (the interfaces can still remain in `prepo-shared-contracts` and be referenced). This PR only includes the copying over of these files and adding setters to reduce the size of later PRs which will actually replace existing fee and hook settings which is a much more invasive change.


ramenforbreakfast:
Updated with new convention of immediately assigning roles. Also added in a forgotten test for SET_MINTING_FEE_PERCENT_ROLE.

XavierEkkel (approved):
LGTM

PR #1314, opened 2023-05-09
chore: remove i18n

MeLlamoPablo:
In order to move towards decentralized front-ends, we must use Next.js Static Exports, and there's [a number of unsupported features we can't use](https://nextjs.org/docs/pages/building-your-application/deploying/static-exports#unsupported-features), among which there's Internationalized Routing.

We don't take advantage of our current internationalization anyway, so this PR removes it. In the future, when we actually implement i18n we'll have to use a solution compatible with Next.js Static Exports.

Note that this PR introduces an `useClient` barrier to `_app.tsx` in order to prevent our app from being rendered during Next Static Site Generaton (build time). **This simply maintains the old behavior because apparently `I18nProvider` was preventing SSG too**. In order to check this, simply inspect the source of `https://app.prepo.io/trade`. You'll see that there's nothing rendered inside `div#__next`.

Screenshot 2023-05-09 at 13 32 35

We never took advantage of Static Site Generation to begin with, and this PR simply prevents SSG from activating with the removal of `I18nProvider`. There are light/dark mode sync issues that need to be resolved before activating it, but we should do that in the future for a faster perceived load time.


chrisling-dev (approved):
LGTM

PR #1315, opened 2023-05-09
feat: exclude fee from costBasis for better rate

chrisling-dev:
https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=1160865fbe4c4639af9d63918c7b7d17&p=444f0c3fb6a54d8687e85887a901dbe8&pm=s


apps/subgraphs/prepo-arbitrum/src/utils/positions.ts
```diff
@@ -74,9 +74,14 @@ export function updatePosition(event: Swap): void {
     ? pool.token0Price
     : pool.token1Price
 
+  // get price excluding fee
+  const feeDenominator = BigDecimal.fromString('1000000')
+  const percentageAfterFee = feeDenominator.plus(pool.fee.toBigDecimal())
+  const priceWithoutFee = priceInWstEth.times(feeDenominator).div(percentageAfterFee)
```
MeLlamoPablo, line 80: Should we maybe store `pool.fee` directly as:

```ts
BigDecimal.fromString(event.params.fee.toString()).div(BigDecimal.fromString('1000000'))
```

?

This way we wouldn't have to apply the `feeDenominator` every time we want to use the fee value. Note that I haven't tested that `event.params.fee.toString()` works properly.

MeLlamoPablo (approved):

apps/subgraphs/prepo-arbitrum/src/utils/positions.ts
chrisling-dev: I think it's better to store values in the Pool entity the same they are stored on the blockchain, that way it's more consistent we no one would ever forget/get confused about whether the value is converted or not. If we use fee a lot we can write util function to convert them

PR #1316, opened 2023-05-09
chore: import onboard dynamically

MeLlamoPablo:
With the move to a decentralized front-end, now it's a good time to trim down the size of our app bundle (shared across all pages) and our page-specific bundles. This PR isolates Onboard, which is our largest dependency, from our app bundle into its own JS file.

image

Isolating dependencies is always good because they can be reused between front-end versions in the user's cache. But especially in decentralized front-ends there's also a big advantage because the same file can be referenced from multiple directories (aka front-end versions) so it only needs to be pinned once.

This means that we only need to re-upload and re-pin dependencies whenever we update them, not every time we deploy a version. These savings should compound greatly over time, considering that we need to pin all versions forever.

For this PR to work, Onboard is imported asynchronously, so the code that interacts with it has been adapted. If the user tries to connect their wallet before onboard is loaded, the app will simply wait until it is.


chrisling-dev (approved):
Fantastic! LGTM

PR #1317, opened 2023-05-09
chore: remove unused icons

MeLlamoPablo:
This PR removes the unused `PreFakeStock` and `PreFakeToken` icons. It also removes the unused currency icons, and all unused currency-related code that was importing those icons.


chrisling-dev (approved):
LGTM

PR #1318, opened 2023-05-09
refactor: dynamically import icons

MeLlamoPablo:
Blocked by #1317

This PR imports all icons dynamically so that:

* We de-duplicate them and can reuse them between versions (See #1316 for rationale)
* We don't bring unrelated icons into all apps.

    For example, the Tesla and Stripe icons which are used in the simulator app were also included in the bundle of `core-dapp` due to how our `Icon` component is constructed.


chrisling-dev (approved):
LGTM

PR #1319, opened 2023-05-09
feat: repurpose MarketHook accountlist for fee bypass

ramenforbreakfast:
Can simply repurpose `_accountList` in `MarketHook` now that it is no longer being used for blocking access.


apps/smart-contracts/core/contracts/MarketHook.sol
```diff
@@ -24,7 +24,7 @@ contract MarketHook is
     uint256 amountAfterFee
   ) external virtual override onlyAllowedMsgSenders {
     if (address(_accountList) != address(0)) {
-      require(_accountList.isIncluded(funder), "Redeemer not allowed");
+      if (_accountList.isIncluded(funder)) return;
     }
```
XavierEkkel, line 28: nit: `if (address(_accountList) != address(0) && _accountList.isIncluded(funder)) return;`

apps/smart-contracts/core/test/unit/MarketHook.test.ts
```diff
@@ -277,8 +258,49 @@ describe('=> MarketHook', () => {
         expect(tokenSender.send).not.called
       })
 
+      it("doesn't transfer fee to treasury if funder in allowlist and funder = recipient", async () => {
```
XavierEkkel, line 261: add test:
`transfers fee to treasury if fee > 0, recipient in allowlist, funder != recipient`

apps/smart-contracts/core/test/unit/MarketHook.test.ts
```diff
@@ -277,8 +258,49 @@ describe('=> MarketHook', () => {
         expect(tokenSender.send).not.called
       })
```
XavierEkkel, line 260: above there are some tests using 'redeemer' which should be changed to 'funder'

XavierEkkel (approved):
LGTM after comments

apps/smart-contracts/core/contracts/MarketHook.sol
ramenforbreakfast: This will not work,  if it is an AND case, it will always try to assess both conditions, which will fail if `_accountList` is not set, since it will attempt to call `isIncluded` on 0x000... which will cause a revert

PR #1320, opened 2023-05-10
feat: guardrails for setting multipliers

ramenforbreakfast:
Some context around design choices:
- I chose to use a universal `CannotSetMultiplierForAccount` `error` declared in the `TokenSenderCaller` module all of these contracts inherit so that they all have access to it. Since the restrictions on multipliers is different and could possibly change in the future, I felt this was a generic enough error. This does mean that some contracts inheriting `TokenSenderCaller` that do not have restrictions will have this `error` but never use it. I personally do not think this is a big deal and `TokenSenderCaller` is the module that makes the most sense to make this error available.

- For `DepositHook` and `WithdrawHook`, the logic is slightly different because we want to restrict multiplier setting to only `collateral`. But there is the possibility that `collateral` is not set at the beginning. Hence, there is an extra safeguard against setting a multiplier for the zero address.


apps/smart-contracts/core/contracts/DepositTradeHelper.sol
```diff
@@ -218,6 +218,7 @@ contract DepositTradeHelper is
     override
     onlyOwner
   {
+    if (account != address(0)) revert CannotSetMultiplierForAccount(account);
```
XavierEkkel, line 221: add comment above clarifying why we use address zero for the trade rebate multiplier

XavierEkkel (approved):
LGTM after comments

PR #1324, opened 2023-05-10
refactor: import `antd` dynamically

MeLlamoPablo:
This PR imports `antd` dynamically, which brings the bundle from 634kb down to 562kb.


chrisling-dev (approved):
LGTM

PR #1325, opened 2023-05-10
chore: temporarily remove Sentry

MeLlamoPablo:
This PR temporarily removes Sentry because:

* It uses server-side features which we can't have on decentralized front-ends.
* Right now, it is ~5% of our main bundle and loading it dynamically isn' straightforward because of the Next.js integration.

After redesigning the deploy pipeline for decentralized front-end, we can properly re-integrate in a way that doesn't bloat our main bundle size.


chrisling-dev (approved):
LGTM

PR #1328, opened 2023-05-10
refactor: import contract factories and ABIs dynamically

MeLlamoPablo:
ABIs are a good candidate to split out too because theyt were a significant portion of our app bundle and they are generally immutable. Everywhere we interact with contract factories is async or can be made async, so it's easy to add dynamic imports.


MeLlamoPablo:
Closing as this breaks tests and the fix is not as straightforward as I imagined. It's probably not worth to put too much effort into this as Crust storage is very cheap, but I may revisit it in the future.

PR #1329, opened 2023-05-11
feat: refactor collateral deposit approvals to hooks

ramenforbreakfast:
This PR contains some medium-level refactoring of the `deposit` tests for `Collateral`, since cohesive tests around hooks and whether or not a hook would take the whole fee were not present. These tests have been added since this refactoring is entirely focused on reducing unnecessary approvals made in certain situations, similar to what we did with Markets.

Some important changes to note below:
- Introduction of a `TestCollateralHook` contract similar to the `TestMarketHook` contract introduced to have a situation where the fee is partially taken. This is an improved version where the treasury address can be specified as well as the amount to take (so that we can actually have a situation where the hook takes 0% as well, instead of just setting the hook to 0).
- Use of a `MockContract` for `DepositHook`, because we want the hook to actually take a fee so that we can validate cases where the fee is not taken or only partially taken.
- Use of an unlimited approval between the `sender` and `Collateral`. This is so that we can more accurately verify that approvals are not given to hooks in certain cases. Without this, the initial transfer between the user and the `Collateral` contract would emit an `Approval` event, making it very difficult to verify that a BaseToken approval did not happen to the hook.
- `amountToDeposit` and `fee` were moved to the beginning of the setup, since there were many very unnecessary consistent assertions that the deposit amount was > 0 and recalculations of the fee when it was not necessary.
- Removed many unnecessary allowance assertions since we now give unlimited approval at the beginning, and allowances should only need to be re-verified if the allowance is changing (for example, when we are testing when there is insufficient allowance).

I am aware this test uses `sender` nomenclature which is inconsistent with our later agreed upon `funder` and `recipient` naming, I avoided renaming it in this PR and should be done in a later PR.


XavierEkkel (approved):
LGTM after nit

PR #1335, opened 2023-05-11
feat: make eth convertible object from balancer store

chrisling-dev:
This PR uses CloseTradeStore's value as example, other example of useful use case:

Modify `openTradeAmount` which now has multiple versions for string, in ETH, in WstETH, and etc, and simply do
```
get openTradeAmount(): EthConvertible | undefined {
    if(this.openTradeInput === '') return undefined
    const parseInput = parseEther(this.openTradeInput)
    return this.root.balancerStore.convert.fromEth(parsedInput)
}
```

Then we can remove openTradeAmountBN, openTradeIntermediateWstEthOut, openTradeAmountInWstEthBN, openTradeAmountInWstETH, priceImpact and along with a few more values, and simple access them like:

`openTradeAmount?.wstEthBN` // for wst eth in BN
`openTradeAmount.ethBN` // for eth in BN
`openTradeAmount.wstEth` // for wst eth in number
`openTradeAmount.eth` // for eth in number
`openTradeAmount.priceImpact`// for price impact  in number


MeLlamoPablo (changes requested):

MeLlamoPablo (approved):
Nice! I think it's cleaner this way

PR #1338, opened 2023-05-11
feat: refactor collateral withdraw approvals to hooks

ramenforbreakfast:
Similar to the deposit tests, refactoring seemed necessary due to the need to check approvals and hook behavior.

Some important changes to note below:
- Switching the `WithdrawHook` from a fake to a mock so that the hook actually takes a fee.
- Precalculating commonly used values at the beginning, since for 70% of the testcases, they are essentially the same and there was a lot of unnecessary verbosity with repeat assertions and re-calculations of these values. (This is probably the most intrusive change)
- Unlimited base token approval similar to the prior PR to reduce `Approval` event emission noise.

I am aware this test is also confusing in that these tests don't use `funder`, but just use `user` (even more confusing given the `deposit` tests use a different nomenclature). This change can happen in a later PR.


apps/smart-contracts/core/contracts/Collateral.sol
```diff
@@ -81,27 +81,21 @@ contract Collateral is
...
-      _baseToken.approve(address(_withdrawHook), fee);
-      _withdrawHook.hook(
-        msg.sender,
-        recipient,
-        baseTokenAmount,
-        baseTokenAmountAfterFee
-      );
-      _baseToken.approve(address(_withdrawHook), 0);
-    }
+    uint256 actualBaseTokenFeeAmount = _processFee(
```
XavierEkkel, line 85:
```suggestion
    uint256 baseTokenFeeAmount = _processFee(
```

XavierEkkel (approved):
LGTM after nit

PR #1339, opened 2023-05-12
feat: added a bypass list for collateral deposits

ramenforbreakfast:
So for this addition, we need to have some nesting logic because we do not want to call `isIncluded` if the bypass list is not set. But we cannot have direct return cases like previous additions because we do not want to return immediately if there is no bypass list.

For deposits, we also need to call `recordDeposit` without the fee, since even for a bypassed user, we want their deposits to be recorded and accounted for in the total.


XavierEkkel (approved):
LGTM after nit

PR #1340, opened 2023-05-12
feat: added bypass list for collateral withdrawals

ramenforbreakfast:
Some context around changes in this PR:
- `recordWithdrawal` was moved up to the very top so that we can `return` immediately if we detect that the recipient is in the bypass list. This is because withdrawals are always recorded with the full amount without the fee subtracted.
- We do not want to include a bypassed recipient in withdraw limit controls since these are mainly protections against users withdrawing, not for bookkeeping the total amount of assets backing `Collateral`.


apps/smart-contracts/core/contracts/WithdrawHook.sol
```diff
@@ -56,11 +60,15 @@ contract WithdrawHook is
    * accurate value.
    */
   function hook(
+    // TODO can leave funder as an unnamed parameter since it is not used.
     address funder,
```
XavierEkkel, line 64: why not just change now?

XavierEkkel (approved):
LGTM after nits

PR #1344, opened 2023-05-12
feat: setup e2e and add basic test

MeLlamoPablo:
This PR sets up the end-to-end test suite using [Playwright](https://playwright.dev/) and adds a basic test that connects a random wallet to the app.

### How does the wallet connection work

The two options I considered to inject a mock wallet into the test suite were mocking MetaMask via `window.ethereum` or creating a mock WalletConnect-supported wallet. I opted for the latter because it required the least amount of trickery (injecting variables in `window`, mocking the MetaMask API...) and we could benefit from the WalletConnect SDK.

The way it works is:

1. The test suite clicks the "WalletConnect" button and scans the WalletConnect URI
2. `connectMockWallet` is called with the URI, which registers a mock wallet with the WalletConnect server.
3. The prePO app receives connection confirmation and displays the "connected wallet" UI.

For now, the mock wallet is dumb and can only connect. It can't handle any other operation. In the next PR, it will have logic to automatically approve transactions and signature requests, making it useful to test any transaction within the app.

### How to setup

1. Obtain a WalletConnect Project ID from [WalletConnect Cloud](https://cloud.walletconnect.com/)

    Or DM me on Telegram for mine! 😃 
2. Add to your `apps/frontend/core-dapp/.env`:

    ```
    NEXT_PUBLIC_WALLETCONNECT_PROJECT_ID=<project id>
    ```
3. Install dependencies:

    ```sh
    yarn install
    yarn playwright install # on the core-dapp directory
    ```

4. Ensure that the local server is running:

    ```sh
    yarn dev # on the core-dapp directory
    ```

4. Run the tests:

    ```sh
    yarn playwright test
    ```

## e2e roadmap

1. **Set up playwright with a basic test _(you are here)_**
6. Set up a transaction test from an EOA account. You will need to provide a seed via env vars. (#1346)
7. Set up a Tenderly DevNet

    * If that isn't possible, use the real Arbitrum network but simulate transactions instead of submitting them
8. Configure the GitHub actions pipeline

PR #1346, opened 2023-05-12
feat: add support for sending transactions on `e2e` and add deposit&trade test

MeLlamoPablo:
This PR:

* Adds to `connectMockWallet` functionality to respond to the dApp's signature requests
* Adds a test case to test a "Deposit And Trade" transaction and checks that the advertised PPO rebate is actually received, not less.

## How to run

Refer to #1344 for how to run Playwright.

Additionally, you will need a development wallet with some ETH to cover gas, and the `0.0001 ETH` which will be deposited and traded for a long zkSync position.

1. Gather the mnemonic seed of your wallet. The wallet must be the first wallet for that seed.
    
    **⚠️ Use a testing wallet. Do not use a wallet with assets you can't afford to lose.**

2. Expose the seed as an environment variable.

    ```shell
      export E2E_TEST_MNEMONIC=<your mnemonic>
    ```

    **⚠️ Notice the two spaces before `export`. Ensure that you write two spaces, like `  export`. Otherwise, the plaintext seed will be saved to your shell history.**

3. Run tests as usual with `yarn playwright test`. A link to the sent transaction will be logged in the console.
4. Exit your shell session so that the environment variable is deleted from memory.

## e2e roadmap

1. Set up playwright with a basic test (#1344)
5. **Set up a transaction test from an EOA account. You will need to provide a seed via env vars. __(you are here)__**
6. Set up a Tenderly DevNet

    * If that isn't possible, use the real Arbitrum network but simulate transactions instead of submitting them
7. Configure the GitHub actions pipeline


apps/frontend/core-dapp/src/features/trade/open-trade/OpenTradeSummary.tsx
```diff
@@ -86,10 +87,10 @@ const OpenTradeSummary: React.FC = () => {
       <OperationSummary
         key="summary"
         header={
-          <>
+          <TestId id="open-trade-summary-header">
```
chrisling-dev, line 90: I wonder if it'd be better if we keep track of a list of ids in a `test-ids` file, this way we don't accidentally assign duplicated test id to different components?

apps/frontend/core-dapp/src/components/SummaryRecord.tsx
```diff
@@ -36,7 +36,9 @@ const SummaryRecord: React.FC<Props> = ({ children, className, label, loading, t
       <Flex gap={4}>
         <LabelText>{label}</LabelText>
       </Flex>
-      <ValueCol>{loading ? <Skeleton height="22px" width="64px" /> : children}</ValueCol>
+      <ValueCol className="summary-record-value">
```
chrisling-dev, line 39: did you mean data-testid or is this intended?

```suggestion
      <ValueCol data-testid="summary-record-value">
```

apps/frontend/core-dapp/src/components/PpoAmount.tsx
```diff
@@ -55,7 +55,7 @@ const PpoAmount: React.FC<{
   }
 
   return (
-    <Wrapper $gap={iconSpacing}>
+    <Wrapper className="ppo-amount" $gap={iconSpacing}>
```
chrisling-dev, line 58: ditto

```suggestion
    <Wrapper data-testid="ppo-amount" $gap={iconSpacing}>
```

apps/frontend/core-dapp/e2e/utils/wallet-connect.ts
```diff
@@ -55,9 +56,61 @@ export async function connectMockWallet({
...
+        data: txParams.data,
+        from: txParams.from,
+        to: txParams.to,
+        value: txParams.value,
+      }
+
+      const [gas, nonce] = await Promise.all([
+        provider.estimateGas(txBody),
+        provider.getTransactionCount(txBody.from),
+      ])
```
chrisling-dev, line 94: We should use `getFeeData` here to get maxFeePerGas and maxPriorityFeePerGas so it doesn't keep failing because of gas. I've tried the test command few times and only had 1 success, the rest were failures because of gas price 

```suggestion
      const [gas, nonce, feeData] = await Promise.all([
        provider.estimateGas(txBody),
        provider.getTransactionCount(txBody.from),
        provider.getFeeData()
      ])
```

chrisling-dev (changes requested):
Great work! minor comments

apps/frontend/core-dapp/src/components/SummaryRecord.tsx
MeLlamoPablo: This is intended, because there can be multiple summary records in the same screen, which would result in multiple elements with the same test ID. The intention is to find the parent element by test ID and then find the value by class name.

apps/frontend/core-dapp/src/components/PpoAmount.tsx
MeLlamoPablo: Same, intended

apps/frontend/core-dapp/e2e/utils/wallet-connect.ts
MeLlamoPablo: Let me re-evaluate this. I started by pulling the `gasFeeData` but it resulted in overpaying for transactions by a huge margin, so I don't know what went wrong

apps/frontend/core-dapp/src/features/trade/open-trade/OpenTradeSummary.tsx
MeLlamoPablo: My initial reaction is that it would be overkill, but I guess it can't hurt to maintain an enum, and we can always revert the change if it becomes too tedious.

chrisling-dev: Yea I think having this enum is very useful in this case
- prevents unintentional duplicate ids
- allows us to easily search for element using a test id (e.g. can browse the enum for test ids rather than searching through the entire app)

apps/frontend/core-dapp/e2e/utils/wallet-connect.ts
chrisling-dev: how much did it over pay for? If it's not that much I don't think it'll be a big deal since we'll probably be using some kind of devnets soon and will have unlimited ETH

MeLlamoPablo: Added the fee data query and it now works more reliably, thanks! I think in my last attempt I was just unlucky and sent the transaction in a time where arbitrum was very expensive.

chrisling-dev (approved):
LGTM!

PR #1349, opened 2023-05-16
feat: add basic "bring your own RPC support"

MeLlamoPablo:
This PR adds a basic "bring your own RPC system". This will be useful for:

* Injecting the Tenderly RPC on end-to-end tests
* Making decentralized front-ends resilient to RPCs shutting down

It works setting a `?rpc` query parameter the first time the app is visited:

```
https://app.prepo.io?rpc=42161:https://arbitrum-mainnet.infura.io/v3/<infura_id>
```

☝️ The above url will load the app with the RPC `https://arbitrum-mainnet.infura.io/v3/<infura_id>` for Chain ID `42161` (Arbitrum).

The system supports multichain out of the box so it will be compatible when we deploy on other chains:

```
https://app.prepo.io?rpc=1:https://ethereum-mainnet.infura.io/v3/<infura_id>&rpc=42161:https://arbitrum-mainnet.infura.io/v3/<infura_id>
```

Note that the RPC can only be set before the app loads. This means that if/when we have an UI to modify this setting, the app will have to be reloaded.

Theoretically, we could reinitialize our `Web3Store.coreProvider` with the new RPC and it will work, but Onboard doesn't support changing RPCs on the fly and re-initializing Onboard means that the user gets disconnected. Therefore, I think it's simpler to just set the RPC override on startup.

### Security considerations

Malicious RPCs can't steal private keys, but they can attempt to trick the user into signing a transaction unfavorable to them. To prevent this, for now we filter out untrusted RPCs (so far the only RPC allowed will be Tenderly).

If in the future we want to allow the user to customize their RPC, we should design a flow warning them of the dangers of using untrusted RPCs.


packages/prepo-utils/src/getRpcOverrides.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  let rawRpcs: string[]
+
+  if (typeof query.rpc === 'string') {
+    rawRpcs = [query.rpc]
+  } else if (Array.isArray(query.rpc)) {
+    rawRpcs = query.rpc.filter((str): str is string => typeof str === 'string')
+  } else {
+    rawRpcs = []
+  }
```
chrisling-dev, line 23: I don't think we need these lines?
```suggestion
  }
```

chrisling-dev (approved):
Minor comment, LGTM

packages/prepo-utils/src/getRpcOverrides.ts
MeLlamoPablo: We don't if we initialize `rawRpcs = []` which I didn't do before 👍

PR #1351, opened 2023-05-16
feat: integrate e2e tests with Tenderly

MeLlamoPablo:
This PR integrates the e2e tests with Tenderly so that transactions run on their DevNets rather than on the public Abritrum network. This has two advantages:

* We can send tenderly-ETH to any wallet, which removes the need for key management.
* We can impersonate a wallet to execute any arbitrary transaction. For example, we could impersonate the prePO Multisig to increase the deposit limit.

    (not implemented on this PR)

## How to run

Add the following variables to your `apps/frontend/core-dapp/.env`:

```
TENDERLY_ACCOUNT_ID=...
TENDERLY_PROJECT_ID=...
TENDERLY_DEVNET_TEMPLATE_SLUG=...
TENDERLY_ACCESS_TOKEN=...
```

And that's it! Run Playwright as usual - refer to #1344 for how to do it.

### e2e roadmap

1. #1344
5. #1346
6. **Set up a Tenderly DevNet _(you are here)_**
7. Configure the GitHub actions pipeline


chrisling-dev (approved):
LGTM, really awesome work, watching it connect to an account with 10k eth and perform tests on it's on is just like magic

PR #1352, opened 2023-05-17
feat: add chart under maintenance message

chrisling-dev:
https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=1160865fbe4c4639af9d63918c7b7d17&p=c99116c5f99a401e87c1c9eeacdde751&pm=s


apps/frontend/core-dapp/src/stores/graphs/UniswapV3GraphStore.ts
```diff
@@ -78,7 +83,12 @@ export class UniswapV3GraphStore extends GraphStore<RootStore, SupportedContract
         maxDataPerFetch: Math.min(days, UNISWAP_MAX_FIRST),
         onNewData: handleNewHistoricalPoolData(days),
       }
-    )?.data
+    )
+
+    runInAction(() => {
+      if (query?.error) this.outOfService = true
```
MeLlamoPablo, line 89:
```suggestion
      this.outOfService = !!query?.error
```

Otherwise, our chart will be disabled until the user refreshes the app if a query fails for any reason (i.e. no internet), right?

MeLlamoPablo (changes requested):

apps/frontend/core-dapp/src/stores/graphs/UniswapV3GraphStore.ts
chrisling-dev: ahh yes good point

PR #1353, opened 2023-05-17
feat: pull Lido APR client-side

MeLlamoPablo:
This PR moves the Lido API calls to the client side, now that [their API](https://docs.lido.fi/integrations/api/#last-lido-apr-for-steth) supports CORS. Thus, this PR unblocks decentralized FEs again.

Ideally we would fetch the APR from chain to avoid a dependency on an additional external API which could potentially track users. However, the method described on their docs doesn't seem to be available yet, since the contract doesn't seem to be updated. We may be able to switch to on-chain in the future.


apps/frontend/core-dapp/src/stores/LidoStore.ts
```diff
@@ -0,0 +1,37 @@
...
+      throw new Error('Unexpected response from Lido')
+    }
+
+    return (body as LidoAprResponse).data.smaApr / 100
+  }
+}
+
+type LidoAprResponse = {
+  data: { smaApr: number }
+}
```
chrisling-dev, line 37: Could we move this to top of file so it follows the pattern of every other files - types are defined above

chrisling-dev (approved):
LGTM after nit

chrisling-dev (changes requested):
oh and of course after test is fixed

chrisling-dev (approved):
LGTM

PR #1354, opened 2023-05-18
Fix/chart datapoints

chrisling-dev:
https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=1160865fbe4c4639af9d63918c7b7d17


MeLlamoPablo (approved):
We should probably [stop handling undefined here](https://github.com/prepo-io/prepo-monorepo/blob/3ca4a9408b9ef6e34b3b0c110f617a47f9545f7e/apps/frontend/core-dapp/src/features/trade/market-chart/Chart.tsx#L108C1-L109) too right?

PR #1357, opened 2023-05-18
feat: implement state overriding and add deposit tests

MeLlamoPablo:
This PR adds an test case to the end-to-end test suite that deposits 100 ETH to prePO and ensures that the user gets at least the minimum amount advertised in the UI.

In order for the test to succeed, both the global and user deposit limits are raised by 100 wstETH.

I thought that Tenderly supported impersonating wallets (i.e. pretend to be the prePO Multisig and call `setGlobalNetDepositCap`), but DevNets don't support this. Tenderly Forks do support this, but they are not appropriate for our use case.

Instead, we can use `tenderly_setStorageAt` to manually set the state of any contract to be whatever we want. This is a bit more cumbersome, as one has to manually find the storage slot of each variable we want to modify, but it works for our use case too. [This is how the deposit limits are overridden](https://github.com/prepo-io/prepo-monorepo/blob/025e8e1c16b90a5c50d5d3b88142dbffeb793d68/apps/frontend/core-dapp/e2e/utils/limits.ts#L28-L40).

By the way, since our number of tests is growing, you can increase the default limit of 5 Playwright workers for faster results:

```
yarn playwright test --workers=10
```


apps/frontend/core-dapp/e2e/utils/tenderly.ts
```diff
@@ -45,3 +45,25 @@ export async function addBalanceToWallet({
...
+  if (value.byteLength !== 32) throw new Error('Invalid state value length')
+  await provider.send('tenderly_setStorageAt', [contract, slot, `0x${value.toString('hex')}`])
+}
+
+export function bnToBytesArray32(bn: ethers.BigNumber): Buffer {
+  const bnBytes = Buffer.from(bn.toHexString().replace('0x', ''), 'hex')
+  const target = Buffer.alloc(32, 0)
+  bnBytes.copy(target, target.length - bnBytes.length)
+  return target
+}
```
chrisling-dev, line 69: I'd recommend using AbiCoder instead of writing our own code to encode value

With AbiCoder:

```suggestion
const abiCoder = new AbiCoder()

export async function overrideState({
  contract,
  slot,
  value,
  provider,
}: {
  contract: string
  slot: string
  value: BigNumber
  provider: ethers.providers.JsonRpcProvider
}): Promise<void> {
  const encodedValue = abiCoder.encode(['uint'], [value])
  await provider.send('tenderly_setStorageAt', [contract, slot, encodedValue])
}
```

apps/frontend/core-dapp/e2e/utils/limits.ts
```diff
@@ -0,0 +1,41 @@
...
+    slot: DepositRecordSlots._globalNetDepositCap,
+    value: bnToBytesArray32(globalCapBefore.add(parseEther('100'))),
+  })
+
+  await overrideState({
+    contract: depositRecordAddress ?? '',
+    provider,
+    slot: DepositRecordSlots._userNetDepositCap,
+    value: bnToBytesArray32(userCapBefore.add(parseEther('100'))),
+  })
```
chrisling-dev, line 40: and here you'd just do:
```suggestion
  await overrideState({
    contract: depositRecordAddress ?? '',
    provider,
    slot: DepositRecordSlots._globalNetDepositCap,
    value: globalCapBefore.add(parseEther('100')),
  })

  await overrideState({
    contract: depositRecordAddress ?? '',
    provider,
    slot: DepositRecordSlots._userNetDepositCap,
    value: userCapBefore.add(parseEther('100')),
  })
```

chrisling-dev (changes requested):
The suggestion assumes that you'll only ever overrideStates with number values, which is not scalable. Alternatively, value can be a string and accepts an encoded value instead of encoding inside overrideState function.

Nevertheless, I think we should get rid of bnToBytesArray32 and just use AbiCoder, which I have test will return the same string value with less code

apps/frontend/core-dapp/e2e/utils/playwright.ts
```diff
@@ -54,6 +53,21 @@ export async function acceptTerms({ page }: { page: Page }): Promise<void> {
...
+  // Wait to ensure balance is refreshed
+  await new Promise((resolve) => {
+    setTimeout(resolve, 10_000)
+  })
+
+  const ppoText = await page.getByTestId(TestIds.PortfolioTotalBalance).textContent()
+  const ppo = +(ppoText?.replace(' ETH', '') ?? 0)
+  await page.goBack()
+  return ppo
+}
```
chrisling-dev, line 69: since the UI relies on the underlying stores data anyways (which is fetched onchain), is there a way we can directly access `rootStore.portfolioStore` here to retrieve the portfolio value?

apps/frontend/core-dapp/e2e/utils/playwright.ts
```diff
@@ -2,6 +2,7 @@
 import { Page } from '@playwright/test'
 import { ethers } from 'ethers'
 import { connectMockWallet } from './wallet-connect'
+import { TestId, TestIds } from '../../src/components/TestId'
```
chrisling-dev, line 5: Appears to be unused

```suggestion
import {TestIds } from '../../src/components/TestId'
```

chrisling-dev (changes requested):

apps/frontend/core-dapp/e2e/utils/contract-slots.ts
```diff
@@ -0,0 +1,11 @@
...
+// Go to Tenderly explorer and look for a transaction that modified the slot
+// you're looking for. If you can't find any, import the contract into Tenderly
+// and simulate one. Go to the "State" tab and tenderly will show the slot
+// number.
+// Learn more: https://docs.soliditylang.org/en/v0.8.20/internals/layout_in_storage.html
+
+export enum DepositRecordSlots {
+  _globalNetDepositCap = '0x0000000000000000000000000000000000000000000000000000000000000006',
+  _userNetDepositCap = '0x0000000000000000000000000000000000000000000000000000000000000008',
+}
```
chrisling-dev, line 11: don't have to do in this PR but what about using ethers's getStorage so it's dynamic and doesn't break in the event of contract upgrade/mechanism change? :

https://ethereum.stackexchange.com/questions/84176/how-can-i-make-an-off-chain-call-to-get-the-data-stored-at-a-specific-storage-sl

MeLlamoPablo: I haven't found a way to query the storage slot given the variable name, but that would be pretty useful. Do you know one? That link shows how to query a storage value given its slot, not a storage slot given its variable name.

MeLlamoPablo:
> The suggestion assumes that you'll only ever overrideStates with number values, which is not scalable.

I added `Buffer` so that you could pass anything, not just BigNumbers. But I like your `AbiCoder` suggestion, it's much simpler. Note that I didn't implement exactly as you; you need to pass the type to `overrideState` too.

chrisling-dev:
> > The suggestion assumes that you'll only ever overrideStates with number values, which is not scalable.
> 
> I added `Buffer` so that you could pass anything, not just BigNumbers. But I like your `AbiCoder` suggestion, it's much simpler. Note that I didn't implement exactly as you; you need to pass the type to `overrideState` too.

LGTM, another approach would be to just accept value as { types: string[], values: unknown[] } so it can decode many things at the same time, but yea this approach looks good enough for this PR, will approve after testing the tests locally

chrisling-dev (approved):
LGTM! Tested that tests passed locally

PR #1360, opened 2023-05-19
chore: update to yarn v3

MeLlamoPablo:
Blocked by #1379

This PR upgrades to Yarn v3, given that development of Yarn v1 has stopped. The [`nodeLinker: node-modules` option](https://github.com/prepo-io/prepo-monorepo/blob/96bd421967c9475ada577b964099f7d6da1b6e3e/.yarnrc.yml#L1) is enabled, which maintains the classic `node_modules` structure for backwards compatibility.

@ramenforbreakfast - pinging you here as this affects you too. Could you try your regular workflow in this PR?

### What do I need to do to migrate?

Nothing. Starting from Yarn v2, yarn releases are committed to the repo (in our case, on `.yarn/releases/yarn-3.5.1.cjs`) to ensure that everyone is on the same version. Running `yarn` will still work as usual.

### What benefits do we gain?

* Consistency, as everyone will be on the same version.
* Out of the box, installations should be faster.
* Automatic install of typescript types of packages that don't have it (via `plugin typescript`).
* The `yarn upgrade-interactive` command via `plugin interactive-tools`.

See [here](https://yarnpkg.com/getting-started/qa#why-should-you-upgrade-to-yarn-modern) for a more detailed reasoning.

### Can we still improve our setup after this PR?

Yes, there are two main improvements we can make:

1. Committing the cache to the repo.

    Yarn now stores every package compressed on `.yarn/cache`. Then, since we have defined `nodeLinker: node-modules`, running `yarn install` creates the `node_modules` structure.

    `.yarn/cache` uses an efficient format that can be committed, thus moving the package download step to `git clone`. This is beneficial because:

      * Cloning is faster than downloading dependencies. Cloning get slower, but installing dependencies gets much more faster.
      * It reduces the supply chain attack surface, as the CI will not connect to NPM anymore.
      * It makes it very easy to see which dependencies are changing on git.

2. Using `nodeLinker: pnpm`.

    This creates a `node_modules` structure similar to [that of pnpm](https://pnpm.io/motivation), which requires less space and is more efficient. This should reduce install times.

    However, it doesn't work out of the box as there are some errors that need to be fixed.

### Should we migrate to Yarn PnP?

Since Yarn v2, the default install strategy is [Plug'n'Play](https://yarnpkg.com/features/pnp). This strategy aims to eliminate `node_modules` altogether, so that you only clone the repo and don't need to run `yarn` at all - instead, packages are read at runtime from `.yarn/cache`.

This sounds very nice in practice, but in my experience many dependencies don't support PnP out of the box. Migrating to PnP probably will involve a lot work and likely won't be worth it.


ramenforbreakfast:
@MeLlamoPablo I've setup a test branch off this one incorporating the most recent changes to SC GitHub Actions and it seems to work fine!

packages/prepo-constants/package.json
```diff
@@ -3,13 +3,17 @@
...
+    "src/types.ts"
+  ],
   "scripts": {
     "build": "tsc",
     "build:watch": "yarn build --watch"
   },
   "dependencies": {
-    "ethers": "5.6.9"
+    "ethers": "5.6.9",
+    "tslib": "^2.5.2"
```
chrisling-dev, line 16: reading the github repo of tslib, it seems like we need to use `importHelpers` to use tslib, does it do anything if we're simply installing and not specifying the usage anywhere?

package.json
```diff
@@ -68,13 +64,13 @@
         "typescript": "4.6.3"
     },
     "dependencies": {
-        "ethers": "5.6.9",
         "@nomicfoundation/hardhat-chai-matchers": "1.0.2",
-        "@nomiclabs/hardhat-ethers": "yarn:hardhat-deploy-ethers@0.3.0-beta.13",
+        "@nomiclabs/hardhat-ethers": "npm:hardhat-deploy-ethers@0.3.0-beta.13",
```
chrisling-dev, line 68: why is this using npm instead of yarn?

chrisling-dev (commented):
LGTM, have some questions however. Will approve after testing a few PRs that branch off from this PR

packages/prepo-constants/package.json
MeLlamoPablo: Good catch, I'll remove it. I didn't introduce this intentionally, it may have been removed on `main` but reintroduced here due to conflict resolution? I don't know

package.json
MeLlamoPablo: I guess `npm:` is the standard protocol for aliasing packages, and yarn v3 simply auto converted this. See ["Protocols" in the Yarn docs](https://yarnpkg.com/features/protocols).

PR #1362, opened 2023-05-22
feat: make collateral balance convertible and remove overriding balan…

chrisling-dev:
…ceOfSigner

https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=1160865fbe4c4639af9d63918c7b7d17&p=446f8ffc4db14f71baeac0ee43194a71&pm=s


apps/frontend/core-dapp/src/features/withdraw/WithdrawStore.ts
```diff
@@ -44,35 +42,27 @@ export class WithdrawStore {
...
-    if (amount === tokenBalanceFormat) {
-      // If the user clicks "MAX", all the wstETH balance (priced in ETH) will
-      // be selected. However, since the wstETH price is updated in real time,
-      // it is possible that it could grow before the user submits the tx.
-      // Storing 'max-balance' here avoids leaving the user wiht dust.
-      this.userWithdrawalAmountInEth = {
-        type: 'max-balance',
-      }
+    if (value === balance?.inEthString) {
+      // when type is max, use underlying wstETH balance for any logic
```
MeLlamoPablo, line 53:
```suggestion
```

If we're gonna simplify the comment I think we may as well remove it altogether, because the current comment simply describes what the line below does, which I think is redundant.

apps/frontend/core-dapp/src/features/withdraw/WithdrawStore.ts
```diff
@@ -151,7 +137,7 @@ export class WithdrawStore {
 
     return [
       new FeeStore({
-        applicableToWstEth: () => this.withdrawalAmountInWstEthBN,
+        applicableToWstEth: () => this.withdrawalAmount?.inWstEthBN,
```
MeLlamoPablo, line 140: Of note for the future, I think it will be more ergonomic on `FeeStore` to have an `applicableTo` prop that accepts `EthConvertible` rather than `applicableToWstEth`

MeLlamoPablo (approved):
Nice!

apps/frontend/core-dapp/src/features/withdraw/WithdrawStore.ts
chrisling-dev: yeap that could work!

apps/frontend/core-dapp/src/features/withdraw/WithdrawStore.ts
chrisling-dev: cool, addressed this in a separate PR since not blocker

PR #1364, opened 2023-05-22
fix: avoid calculating impact when value is too small

chrisling-dev:
https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=1160865fbe4c4639af9d63918c7b7d17&p=319b7e310be645da90f994f4fb5e8890&pm=s


MeLlamoPablo (approved):
Good catch!

PR #1366, opened 2023-05-22
feat: add e2e test case for withdrawals

MeLlamoPablo:
This PR adds an end-to-end test case that deposits ETH and then withdraws part of the deposited ETH, to ensure that the minimum received amounts we promise to the user are correct.


apps/frontend/core-dapp/e2e/withdraw.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+    rpc: {
+      chainId: NETWORKS.arbitrumOne.chainId,
+      url: rpcUrl,
+    },
+    wallet,
+  })
+
+  await addBalanceToWallet({
+    address: wallet.address,
+    provider: await walletConnect.provider,
```
chrisling-dev, line 35: why not just do `const provider = await walletConnect.provider` at top of function?

apps/frontend/core-dapp/e2e/withdraw.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  await acceptTerms({ page })
+
+  await page.getByRole('link', { name: 'Portfolio' }).click()
+  await page.getByRole('link', { name: 'Deposit →' }).click()
+  await page.getByPlaceholder('0').fill('100')
+  await page.getByRole('button', { name: 'Deposit', exact: true }).click()
+  await walletConnect.waitForTxHash()
+
+  const balanceBefore = await (await walletConnect.provider).getBalance(wallet.address)
```
chrisling-dev, line 46: ditto, code would be more readable if we have provider on top?

```suggestion
  const balanceBefore = await provider.getBalance(wallet.address)
```

chrisling-dev (changes requested):
LGTM with minor comment

MeLlamoPablo:
> why not just do `const provider = await walletConnect.provider` at top of function?

Instead of doing that, I simply awaited the provider in `connectMockWallet` which is much easier. (699ac18)

chrisling-dev (approved):
LGTM

PR #1368, opened 2023-05-22
feat: add `GasStore`

MeLlamoPablo:
https://www.notion.so/prepo/Gas-estimation-ffd6fbf093dd462cbc6275ef6b26e5fc

This PR adds a `GasStore` that exposes a `maxBaseFeePerGas` that can be used to estimate transactions.

Instead of polling `provider.getFeeData` (which by the way hardcodes an 1.5 gwei priority fee, a fee too high for Arbitrum) we call a `BaseFeeGetter` contract. This way, we can integrate it seamlessly into our Multicall flow. [The contract](https://arbiscan.io/address/0xacb0b03e821dc2abeb145341ff507cf0b843864a#code) is just this:

```solidity
contract BaseFeeGetter {
  function getBaseFee() view external returns (uint256) {
    return block.basefee;
  }
}
```

In a future PR we'll be able to use this store to estimate gas.


chrisling-dev (approved):
LGTM, ideally a store like this should sit in prepo-stores, but this PR is good for now

PR #1369, opened 2023-05-22
feat: add Gas Cost to Deposit Summary

MeLlamoPablo:
Blocked by #1368 

Screenshot 2023-05-22 at 19 31 25

https://www.notion.so/prepo/Gas-estimation-ffd6fbf093dd462cbc6275ef6b26e5fc

This PR adds the gas cost to the deposit summary. It does so by hardcoding a 2.5M gas limit and multiplying it by the base fee. This hardcoded limit is what an average deposit transaction may consume, **without including a big 2x buffer like the Notion ticket suggests**. I went for this solution because:

* The gas cost we show is an estimate (as mentioned in the tooltip). The users will need to check their wallets for a more precise value.
* We don't want a 2x buffer because that would result in very high estimations. We want to estimate for the average case, not the worst.
  * Similarly, we use the exact base fee, without adding a buffer.
* For some transactions, we can't estimate gas without connecting a wallet.
* Setting up an `estimateGas` flow would be rather complex as we'd have to modify our "send transaction" methods to be able to output the parameters without submitting the whole transaction. Since the gas is an estimate, we should go for the simplest solution.


apps/frontend/core-dapp/src/stores/GasStore.ts
```diff
@@ -4,17 +4,15 @@ import { computed, makeObservable, observable } from 'mobx'
 import { RootStore } from './RootStore'
 import { SupportedContracts } from '../lib/contract.types'
 import { BaseFeeGetterAbi, BaseFeeGetterAbi__factory } from '../../generated/typechain'
-import { fraction, mul } from '../utils/fraction-utils'
 
 type GetBaseFee = BaseFeeGetterAbi['getBaseFee']
```
chrisling-dev, line 8:
```suggestion
type GetBaseFee = BaseFeeGetterAbi['functions']['getBaseFee']
```

apps/frontend/core-dapp/src/stores/GasStore.ts
```diff
@@ -26,14 +24,18 @@ export class GasStore extends ContractStore<RootStore, SupportedContracts> {
...
-    const baseFee = this.getBaseFee()
+    const { baseFee, maxPriorityFeePerGas } = this
     if (baseFee === undefined) return undefined
-    return mul(baseFee, GasStore.BASE_FEE_BUFFER)
+    return baseFee.add(maxPriorityFeePerGas)
+  }
+
+  private get baseFee(): BigNumber | undefined {
+    return (this.getBaseFee() as [BigNumber] | undefined)?.[0]
   }
```
chrisling-dev, line 34:
```suggestion
  private get baseFee(): BigNumber | undefined {
    return this.getBaseFee()?.[0]
  }
```

chrisling-dev (approved):
LGTM, with suggestion on getBaseFee type

apps/frontend/core-dapp/src/stores/GasStore.ts
MeLlamoPablo: Thanks!

PR #1374, opened 2023-05-23
feat: assign default beacons upon market deployment

ramenforbreakfast:
Since we are moving to permissionless market deployment, we need to be able to enforce fee and hook beacons when markets are created in `createMarket`. This PR adds setters for beacons that will be automatically assigned to any deployed market.


XavierEkkel (approved):
LGTM

PR #1375, opened 2023-05-24
feat: remove scroll in terms checklist

chrisling-dev:
https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=1160865fbe4c4639af9d63918c7b7d17&p=afc7144e5fbb4c6b83b217bfdb87d39f&pm=s


MeLlamoPablo (changes requested):
I'm not necessarily against this solution, but we should note that this doesn't get rid of the scrollbar if the window height shrinks (in that case we have two scrollbars, the terms one and the body one)

Maybe we could not have `max-height` on the terms box and let the body scrollbar be the only one?

chrisling-dev:
> I'm not necessarily against this solution, but we should note that this doesn't get rid of the scrollbar if the window height shrinks (in that case we have two scrollbars, the terms one and the body one)
> 
> Maybe we could not have `max-height` on the terms box and let the body scrollbar be the only one?

good point, addressed

MeLlamoPablo (changes requested):
This change affects the terms scroll too (which also makes impossible to agree to them since we rely on the scrollbar).

We should only do this for the "rules" screen? (The notion ticket is misleading)

MeLlamoPablo (approved):
LGTM

PR #1379, opened 2023-05-24
refactor: manage `codegen` with `turbo` instead of `postinstall`

MeLlamoPablo:
This PR stops using `postinstall` to run `codegen` and manages `codegen` script with `turbo` instead. This is because previously all `codegen` scripts would run when installing dependencies, even those unrelated to the package being built. Using `turbo` we benefit from caching and dependency resolution so the only executed `codegen` scripts are relevant.

For convenience, all codegen scripts are run on the workspace root `prepare` script, unless we are on CI.


chrisling-dev (approved):
LGTM!

PR #1383, opened 2023-05-26
feat: add missing reentrancy guards

ramenforbreakfast:


apps/smart-contracts/core/contracts/DepositRecord.sol
```diff
@@ -29,6 +31,7 @@ contract DepositRecord is
   function recordDeposit(address user, uint256 amount)
     external
     override
+    nonReentrant
```
XavierEkkel, line 34: not needed because the only external calls are `view` only

apps/smart-contracts/core/contracts/DepositRecord.sol
```diff
@@ -48,6 +51,7 @@ contract DepositRecord is
   function recordWithdrawal(uint256 amount)
     external
     override
+    nonReentrant
```
XavierEkkel, line 54: not needed because the only external call (in `onlyAllowedMsgSenders`) is `view` only

XavierEkkel (approved):
LGTM after comments

PR #1386, opened 2023-05-26
refactor: transition remaining relevant shared contracts to use errors

ramenforbreakfast:
I only refactored relevant shared contracts, with the exception of access control related ones because they inherit from OpenZeppelin contracts and thus only a small portion of their error cases would be changed to use custom errors.


XavierEkkel (approved):
LGTM

PR #1387, opened 2023-05-26
fix: decimals typed at the end of the string

MeLlamoPablo:
This PR fix a bug in CurrencyInput where the decimal point would always be typed at the end no matter where the cursor was located. This is because in #1018 we assumed that the user would always be typing the decimal point at the end of the input, but that isn't necessarily the case. Here's the bug:

https://github.com/prepo-io/prepo-monorepo/assets/11708035/3fb14871-832c-4329-9782-9719bb3c167c

And here's the fix:

https://github.com/prepo-io/prepo-monorepo/assets/11708035/e2f25611-5d52-4772-a876-46633d8490f6


chrisling-dev (changes requested):
Build seems to be failing?

chrisling-dev (changes requested):
There's a bug in this PR, added suggestion and description of bug here: https://github.com/prepo-io/prepo-monorepo/pull/1399

PR #1395, opened 2023-05-30
feat: adding arbitrary data field for market contracts

ramenforbreakfast:


apps/smart-contracts/core/test/fixtures/PrePOMarketFixture.ts
```diff
@@ -43,8 +44,8 @@ export async function smockPrePOMarketFixture(
     marketParams.floorValuation,
     marketParams.ceilingValuation,
     marketParams.expiryTime,
-    marketParams.hookBeacon,
-    marketParams.feeBeacon
+    marketParams.hookBeacon ? marketParams.hookBeacon : ZERO_ADDRESS,
+    marketParams.feeBeacon ? marketParams.feeBeacon : ZERO_ADDRESS
```
XavierEkkel, line 48: this is unrelated to the new data field right?

XavierEkkel (approved):
LGTM

apps/smart-contracts/core/test/fixtures/PrePOMarketFixture.ts
ramenforbreakfast: Yes sorry I should have included in this PR notes regarding that some of these changes are related to the fact that integration tests were not being run on accident and thus I did not catch this breaking change in our fixtures related to the introduction of fee beacons.

PR #1397, opened 2023-05-30
feat: add support for data fields for broker and helper

ramenforbreakfast:
One thing that probably needs context for this PR, is that in the past week, we have submitted some changes to our core stack that make the setup for complex integration tests like `DepositTradeHelper` more involved.

Specifically, before, since the `Collateral` contract took the fee even if no hook was configured, setting it aside as additional reserve, we could have the `Collateral` contract take the fee without setting up any hook. Now, the hook needs to be setup properly in order for the fee to be taken.

While I was working on these changes last week, I seem to have forgotten that `yarn t` only runs the `unit` tests, and thus did not catch these test changes that needed to be done to account for the recent core changes.

I have included the additions to the test setup here since they are needed for these tests to run.


apps/smart-contracts/core/helpers/index.ts
```diff
@@ -151,14 +151,15 @@ export async function mintCollateralFromBaseToken(
   funder: SignerWithAddress,
   recipient: string,
   collateralAmount: BigNumber,
-  collateral: Collateral | MockContract<Collateral>
+  collateral: Collateral | MockContract<Collateral>,
+  depositData = '0x'
```
XavierEkkel, line 155: use a `JUNK_PAYLOAD` constant like the other test files

and same for other files using '0x'

XavierEkkel (approved):
LGTM after nit

PR #1398, opened 2023-05-30
feat: modify ArbitrageBroker to use an external list

ramenforbreakfast:


XavierEkkel (commented):
almost LGTM, just a comment

PR #1399, opened 2023-05-31
fix: currency input decimals

chrisling-dev:
Suggested fix for: https://github.com/prepo-io/prepo-monorepo/pull/1387


MeLlamoPablo (approved):
Good catch! Thanks

PR #1402, opened 2023-06-01
feat: update IPNS after deploying to IPFS

MeLlamoPablo:


.github/workflows/preload-fe.yml
```diff
@@ -0,0 +1,62 @@
...
+          NEXT_PUBLIC_PREPO_BUILD_ID: ${{ github.sha }}
+          NEXT_PUBLIC_VERCEL_ENV: production
+
+      - name: Preload
+        working-directory: apps/frontend/core-dapp
+        env:
+          PLAYWRIGHT_BASE_URL: ${{ inputs.url }}
+          PLAYWRIGHT_EXPECTED_BUILD_ID: ${{ inputs.expected_build_id }}
+          PLAYWRIGHT_SITE_PATH: ${{ inputs.site_path }}
+        run: yarn playwright test -g "ensure correct Build ID" --retries=200
```
chrisling-dev, line 62: Do we not have to include a build step after installation and codegen step and before preloading?

chrisling-dev (changes requested):
LGTM after comment

.github/workflows/preload-fe.yml
MeLlamoPablo: * Why do you think we need to `build` the front-end here? We test directly against the gateway/ENS resolver.
* `codegen` is required because the end-to-end test suite depends on generated code (i.e. smart contract interfaces).

    (Technically the test suite we run here doesn't depend on it, but Playwright evaluates all files so the build fails without `codegen`)

chrisling-dev: ahh gotcha, it was a bit confusing to me in that almost all steps in our deployment pipeline is executed here (install -> codegen -> test) except for build, and wasn't clear that the install and codegen is for testing only

I was wondering either we don't need install/codegen at all since the build is already done in another action, or that build should be triggered here lol

chrisling-dev (approved):
LGTM

.github/workflows/preload-fe.yml
MeLlamoPablo: We do need install again because this is a different job, so nothing that happens on the deploy jobs has any effect here.

PR #1405, opened 2023-06-02
fix: `turbo` caching

MeLlamoPablo:
This PR fixes caching inconsistencies in `turbo`. Sometimes the build was incorrectly hitting the cache. The changes are:

* Turbo is upgraded.
* The environment dependencies are moved to `env`, as having them on `dependsOn` is now deprecated.
* The `core-dapp` tasks properly specify their inputs
* `next-sitemap` moved to a separate `next export` task as it was polluting the `public` directory which is an input of `core-dapp`, forcing the cache to always be invalidated.


chrisling-dev (approved):
LGTM

PR #1407, opened 2023-06-02
chore: resolve solhint warnings, mostly new import requirements

ramenforbreakfast:
So this proved to be much more involved than originally anticipated, due to the following reasons:
- We were using an old solhint from March 2021, which was released prior to Solidity custom errors being a feature. Because we now have introduced Solidity custom errors, I was getting parsing errors when running `solhint` since it didn't know how to parse custom errors. Therefore, I decided to update our `solhint` to the latest version, `3.4.1`.
- `3.4.1`'s new parsing engine, detects a solhint warning that was never previously detected, highlighting the way imports happen. Previously, we were just doing global imports, without specifying _what_ we were importing from each file. This is deemed not good practice by solhint recommended rules and I think it makes sense given we often globally import entire files but only need portions of them. 99% of the changes in this PR are just altering these imports to specify exactly what is needed from the file.
- Also noticed that we had left `Ownable` as an imported file in `PrePOMarket`, this is now removed.
- Note, there are still 51 warnings in our `solhint` report. These I have left unresolved because they are complaints that we have non mixed-case function declarations in our interfaces, because we chose to expose constants which are all caps in our interfaces as view functions. There are also complaints around line lengths specifically for our imports, which is not really practical given import paths themselves can exceed our line length. However, it didn't make sense to altogether list these as ignored rules because in most cases, we would want to follow these rules.

- A follow up PR will be submitted focusing on shared contracts, since this PR is already massive.


XavierEkkel (approved):
LGTM

PR #1408, opened 2023-06-02
chore: resolve solhint warnings for shared contracts

ramenforbreakfast:
Same story with the previous PR, 99% of changes are related to new import guidelines. I have also swapped the order of some imports since they weren't following the ordering the rest of our contracts use. I have also ignored the remaining linting errors that are either complaining about non-mixed-case declarations in our interfaces or line lengths for our imports.

In addition, I decided to just go ahead and get rid of the `Pausable` and `ProtectedHook` contracts that we don't use since I saw no point in resolving linting errors for them.


XavierEkkel (approved):
LGTM

PR #1409, opened 2023-06-03
refactor: immediately assign roles to deployer

ramenforbreakfast:
For context, in some tests the role holder reversion test had to be modified since before the deployer did not automatically get assigned to each role. We often used the deployer as a user that didn't have a role which now doesn't work since these contracts now automatically assign them. I have instead modified these to use our more elegant `revertIfNotRoleHolder` utility that was recently written, which will use the last account generated from a key since we most likely wouldn't use it in a test.


XavierEkkel (approved):
LGTM

PR #1410, opened 2023-06-03
feat: make AccountList enumerable

ramenforbreakfast:
Decided to also change the naming for the arrays used in the `AccountList` test because when they were written, I believe it was written thinking we were using them as blacklists, hence why the array to set all values to true was called `blockedArray`


packages/prepo-shared-contracts/contracts/AccountList.sol
```diff
@@ -33,6 +40,24 @@ contract AccountList is IAccountList, SafeOwnable {
   }
 
   function isIncluded(address account) external view override returns (bool) {
-    return _resetIndexToAccountToIncluded[_resetIndex][account];
+    (, uint256 inclusionUint) = _resetIndexToAccountToIncluded[_resetIndex]
+      .tryGet(account);
+    return inclusionUint != 0 ? true : false;
```
XavierEkkel, line 45:
```suggestion
    return inclusionUint != 0;
```

same for duplicate file

XavierEkkel (approved):
LGTM after nits

PR #1412, opened 2023-06-06
feat: custom rpc settings

chrisling-dev:
https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=00b907a28d144d1d9dadd16c96b536ea&p=7cf7d063c90741bfb6a2e8f656859e5b&pm=s


apps/frontend/core-dapp/src/components/SettingsMenu/RpcSettings.tsx
```diff
@@ -0,0 +1,88 @@
...
+        <Flex gap={12} flexDirection="column">
+          <Description>Entry custom RPC for {network.displayName ?? network.chainName}</Description>
+          <Input
+            size="small"
+            placeholder="e.g. https://rpc.custom.com"
+            value={rpcUrl}
+            onChange={(e) => {
+              setRpcUrl(e.target.value)
+            }}
+          />
```
MeLlamoPablo, line 78: Some suggestions:

* Shorten the copy:

    ```suggestion
              <Description>Custom RPC for {network.displayName ?? network.chainName}</Description>
              <Input
                size="small"
                placeholder="e.g. https://rpc.custom.com"
                value={rpcUrl}
                onChange={(e) => {
                  setRpcUrl(e.target.value)
                }}
              />
    ```

* Make `Description` a label and attach it to the input for better accessibility.

packages/prepo-stores/src/Web3Store.ts
```diff
@@ -71,6 +78,44 @@ export class Web3Store {
...
+          provider,
+          priority: 0,
+          stallTimeout: FALLBACK_PROVIDER_CONFIG.STALL_TIMEOUT,
+          weight: 1,
+        },
+      ])
+
+      this.initProvider()
+    }
+  }
```
MeLlamoPablo, line 117: Why not reuse `createFallbackProvider` and modify/remove the `getRpcOverrides` logic to support passing providers from `Web3Store`?

apps/frontend/core-dapp/src/components/SettingsMenu/RpcSettings.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+  const [showForm, setShowForm] = useState(false)
+  const [rpcUrl, setRpcUrl] = useState(customRpc[network.name] ?? '')
+  const disabled = customRpc[network.name] === rpcUrl || rpcUrl === ''
+
+  if (isProduction) return null
+
+  const handleSave = (): void => {
+    web3Store.setRpc(network.name, rpcUrl)
+  }
```
MeLlamoPablo, line 50: IMO we should close on save

MeLlamoPablo (approved):
Looking good! None of the comments are blockers so feel free to disregard, address them here or on another PR

packages/prepo-stores/src/Web3Store.ts
chrisling-dev: Most of getRpcOverrides focus on qetting query params and I don't want to do a big refactor on that in this PR as this PR only focuses on the ability to add a custom RPC without having to refresh the page. There will be a separate PR to remove the logic from getRpcOverrides + make sure test isn't broken by this

apps/frontend/core-dapp/src/components/SettingsMenu/RpcSettings.tsx
```diff
@@ -0,0 +1,88 @@
...
+          setShowForm(true)
+        }}
+      >
+        RPC Settings
+      </SettingsMenuItem>
+      <FormWrapper show={showForm}>
+        <Flex alignItems="center" justifyContent="space-between" mb={12}>
+          <Title>RPC Settings</Title>
+          <CloseIconWrapper onClick={() => setShowForm(false)}>
+            <Icon name="cross" height="16" width="16" />
```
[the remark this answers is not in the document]
chrisling-dev: hmm maybe, but the UI in this PR is only for internal testing and may look completely different (might not even be in this settings dropdown), so I'd suggest we don't spend any more time on UI

apps/frontend/core-dapp/src/components/SettingsMenu/RpcSettings.tsx
chrisling-dev: The label in Input is too big and I don't want to go through the hassle of changing the Input component since this UI may eventually be completely different anyway, shortened the label

PR #1413, opened 2023-06-06
feat: add a `rollback-ipns` GitHub Workflow

MeLlamoPablo:
This PR

* Extracts the `update_ipns` and `preload` jobs from `deploy-fe` into a reusable `publish-ipns` workflow.
* Creates a `rollback-ipns` `workflow_dispatch`-activated workflow. We will use this workflow to manually initiate rollbacks via the GitHub UI.

Note: I haven't tested the new `rollback-ipns` workflow because it needs to be on `main` first, but it should work as it reuses most of the code.


chrisling-dev (approved):
LGTM

PR #1414, opened 2023-06-06
feat: log IPFS deploys to Notion

MeLlamoPablo:
This PR expands the `deploy-fe` workflow with Notion logging. After each successful deployment, the IPFS CID and Git commit hash will be saved to a Notion database. This will be helpful for keeping track of deployed front-ends in order to easily do rollbacks.

This implementation saves the information [on this database controlled by me](https://animated-polyester-aad.notion.site/c22372351a6a455c82d19b28ddb584ae?v=1af9a0b2f2cd4143926c6e5618508304&pvs=4). When we've properly tested and configured the integration in our Notion workspace, we will migrate it there.


.github/workflows/deploy-fe.yml
```diff
@@ -94,3 +94,67 @@ jobs:
...
+              "Git Commit": {
+                "rich_text": [
+                  {
+                    "text": { "content": "$GIT_SHA" }
+                  }
+                ]
+              },
+              "Date": {
+                "date": { "start": "$now" }
+              }
```
chrisling-dev, line 151: I think you can just create a "Created Time" field in notion database, so this field can be removed here

chrisling-dev (approved):
LGTM after comment

.github/workflows/deploy-fe.yml
MeLlamoPablo: Good catch, much easier this way!

PR #1415, opened 2023-06-07
feat: add a fallback settlement price after expiry

ramenforbreakfast:
So this proved to be much more involved than originally anticipated. Originally, the scope of this feature was simply to add an additional fallback final settlement price that can be assigned by _anyone_ after the expiry time has passed for a market.
However, there was additional refactoring involved because the addition of another constructor argument brought on STACK_TOO_DEEP issues. To address this, I decided to perform a refactoring I've been wanting to do for awhile, here is a summary of the changes:
- `expiryLongPayout` is a new parameter passed in upon market creation. `finalLongPayout` can be set to this price by anybody after `expiryTime` has passed. The `finalLongPayout` controller can still set the price after `expiryTime` has reached, but once it is set by either method, it cannot be set again.
- `PrePOMarket` parameters now uses a `MarketParameters` struct, which is used to store parameters that are shared by both the `PrePOMarket` constructor and the `createMarket` method in `PrePOMarketFactory`. Many tests and utilities have been modified/streamlined to use the new parameter struct.

Some caveats about design decisions when writing the switch to `MarketParameters`:
- The `MarketCreation` event now uses the `MarketParameters` directly, instead of having each element being individually passed in. This is both to avoid STACK_TOO_DEEP and for simplicity. This does mean that more information is passed to the event, but I think this is harmless.
- For some reason, I could not figure out why only importing the `MarketParameter` from `IPrePOMarket.sol` did not work. The only way I could get the contract to compile was to import the whole interface and reference it like `IPrePOMarket.MarketParameters`.  Sources indicated to me that this should be possible, but I spent quite a bit of time on this and could not figure why it didn't work here. It is unfortunate given the whole point of using this nomenclature is to extract single items outside of a larger file import.


apps/smart-contracts/core/contracts/PrePOMarketFactory.sol
```diff
@@ -18,18 +18,11 @@ contract PrePOMarketFactory is
   IValueBeacon private _feeBeacon;
 
   function createMarket(
-    string memory tokenNameSuffix,
-    string memory tokenSymbolSuffix,
+    string calldata tokenNameSuffix,
+    string calldata tokenSymbolSuffix,
```
XavierEkkel, line 22: why?

apps/smart-contracts/core/contracts/interfaces/IPrePOMarket.sol
```diff
@@ -7,6 +7,18 @@ import {IAddressBeacon} from "prepo-shared-contracts/contracts/interfaces/IAddre
 import {IValueBeacon} from "prepo-shared-contracts/contracts/interfaces/IValueBeacon.sol";
 
 interface IPrePOMarket {
+  struct MarketParameters {
+    address deployer;
```
XavierEkkel, line 11: shouldn't be in parameters as it shouldn't be settable by the caller of `PrePOMarketFactory#createMarket`

apps/smart-contracts/core/contracts/PrePOMarketFactory.sol
ramenforbreakfast: I made the string arguments calldata because we've been doing that consistently as a legacy ETH mainnet optimization, although of course these optimizations vary in worth between L2s. Casting them as memory is a leftover decision from the initial PrePOMarket implementation last year.

So mainly this was a decision for consistencies sake.

apps/smart-contracts/core/contracts/PrePOMarket.sol
```diff
@@ -11,22 +11,24 @@ contract PrePOMarket is
   ReentrancyGuard,
   SafeAccessControlEnumerable
 {
-  ICollateral private immutable _collateral;
   ILongShortToken private immutable _longToken;
   ILongShortToken private immutable _shortToken;
```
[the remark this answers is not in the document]
ramenforbreakfast: I decided grouping these based on the order they are passed in, but really we haven't established core guidelines around how we want to organize state variables.

This was also done prior to any sort of standard around how we wrote state variables, this pertains to the comment above as well regarding newlines.

In this file, I seem to have opted for grouping state based on relevancy, so KEY constants together, ROLE constants together, LIMIT/UNIT constants together, payouts together, etc.

But in general, this might be a good opportunity to decide on general guidelines for how we group state variables. (This should be done in a separate PR).

ramenforbreakfast:
This PR has been updated to enforce that the `deployer` of a `PrePOMarket` created via `createMarket` be the `msg.sender` of `createMarket`. `deployer` is no longer specifiable as a parameter.

Of course, `deployer` must remain as a constructor argument so the factory can relay the `msg.sender` when creating the market.

apps/smart-contracts/core/test/unit/PrePOMarketFactory.test.ts
```diff
@@ -299,15 +306,17 @@ describe('=> PrePOMarketFactory', () => {
...
       const longToken = await LongShortTokenAttachFixture(await prePOMarket.getLongToken())
       const shortToken = await LongShortTokenAttachFixture(await prePOMarket.getShortToken())
 
       expect(await prePOMarket.getCollateral()).to.eq(collateralToken.address)
       expect(await longToken.owner()).to.eq(prePOMarket.address)
       expect(await shortToken.owner()).to.eq(prePOMarket.address)
-      expect(await prePOMarket.getFloorLongPayout()).to.eq(TEST_FLOOR_PRICE)
-      expect(await prePOMarket.getCeilingLongPayout()).to.eq(TEST_CEILING_PRICE)
+      expect(await prePOMarket.getFloorLongPayout()).to.eq(TEST_FLOOR_PAYOUT)
+      expect(await prePOMarket.getCeilingLongPayout()).to.eq(TEST_CEILING_PAYOUT)
```
XavierEkkel, line 319: should also check the expiry time and expiry payout values

XavierEkkel (approved):
LGTM after comments

apps/smart-contracts/core/test/fixtures/PrePOMarketFixture.ts
```diff
@@ -1,52 +1,39 @@
...
-    marketParams.floorValuation,
-    marketParams.ceilingValuation,
-    marketParams.expiryTime,
-    marketParams.hookBeacon,
-    marketParams.feeBeacon
-  )) as PrePOMarket
+  return (await prePOMarket
+    .connect(marketParams.deployer)
+    .deploy(
+      marketParams.deployer.address,
```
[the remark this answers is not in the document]
ramenforbreakfast: ah yeah this fixture is no longer used so wasn't affecting any tests. ill fix this anyway.

PR #1416, opened 2023-06-07
chore: remove newlines between events and alphabetize

ramenforbreakfast:


ramenforbreakfast:
closing since we need to perform more markets refactoring and makes sense to do this after all of that is completed.

PR #1417, opened 2023-06-07
fix: make market beacons immutable and remove governance

ramenforbreakfast:
2nd Update Summary: 
- `PrePOMarket` will now have `immutable` beacons, set upon constructor. Requirement for both beacons is still enforced inside `PrePOMarketFactory` when `createMarket` is called.
- beacon setters and roles have been removed.
- More complex fee logic involving a custom fee key that will be referenced if > 0 instead of the default key. This custom fee key is simply the padded address of the contract.
- This logic is moved to an `internal` method to reduce nesting and improve readability as the existing fee processing function is already quite lengthy.
- Because this fee logic would have been incredibly cumbersome to add to the existing fee logic tests that are in both `mint` and `redeem`, I have created a `TestPrePOMarket` contract that exposes this `internal` method for easier testing.


apps/smart-contracts/core/contracts/test/TestPrePOMarket.sol
```diff
@@ -0,0 +1,28 @@
...
+  )
+    PrePOMarket(
+      deployer,
+      longToken,
+      shortToken,
+      addressBeacon,
+      uintBeacon,
+      parameters
+    )
+  {}
```
XavierEkkel, line 23: can't this be left out?

apps/smart-contracts/core/contracts/test/TestPrePOMarket.sol
```diff
@@ -0,0 +1,28 @@
...
+
+  function getFeePercent(bytes32 feeKey) external view returns (uint256) {
```
XavierEkkel, line 25: i think itd be good to just surface this in the actual interface

apps/smart-contracts/core/contracts/interfaces/IPrePOMarketFactory.sol
```diff
@@ -6,36 +6,36 @@ import {IAddressBeacon} from "prepo-shared-contracts/contracts/interfaces/IAddre
...
+  event AddressBeaconChange(address beacon);
   event MarketCreation(
     address market,
     address deployer,
     address longToken,
     address shortToken,
-    address hookBeacon,
-    address feeBeacon,
+    address addressBeacon,
+    address valueBeacon,
```
XavierEkkel, line 17:
```suggestion
    address uintBeacon,
```

apps/smart-contracts/core/contracts/test/TestPrePOMarket.sol
ramenforbreakfast: No, leaving it out only applies to zero-arg constructors.

apps/smart-contracts/core/contracts/PrePOMarket.sol
```diff
@@ -252,12 +225,10 @@ contract PrePOMarket is
     address recipient,
     bytes calldata data
   ) internal returns (uint256 actualCollateralFeeAmount) {
-    if (address(_hookBeacon) == address(0)) return 0;
-    IHook hook = IHook(_hookBeacon.get(hookKey));
+    if (address(_addressBeacon) == address(0)) return 0;
```
XavierEkkel, line 228: can remove this, since this is no longer possible

apps/smart-contracts/core/contracts/PrePOMarket.sol
```diff
@@ -311,4 +282,21 @@ contract PrePOMarket is
     _shortToken.mint(recipient, longShortAmount);
     emit Mint(msg.sender, recipient, longShortAmount, collateralFeeAmount);
   }
+
+  function _getFeePercent(bytes32 feeKey)
+    internal
+    view
+    returns (uint256 feePercent)
+  {
+    if (address(_uintBeacon) == address(0)) return 0;
```
XavierEkkel, line 291: can remove this since it will always be non-zero

apps/smart-contracts/core/contracts/PrePOMarket.sol
```diff
@@ -311,4 +282,21 @@ contract PrePOMarket is
...
+    uint256 customFeePercent = _uintBeacon.get(
+      bytes32(uint256(uint160(address(this))))
```
XavierEkkel, line 293: worth adding a comment above this explaining why so match typecasting is done

apps/smart-contracts/core/contracts/PrePOMarket.sol
```diff
@@ -311,4 +282,21 @@ contract PrePOMarket is
...
+    );
+    if (customFeePercent != 0) {
+      feePercent = customFeePercent > FEE_LIMIT ? 0 : customFeePercent;
```
XavierEkkel, line 296: use type(uint256).max instead of FEE_LIMIT to represent 0

apps/smart-contracts/core/contracts/PrePOMarket.sol
```diff
@@ -311,4 +282,21 @@ contract PrePOMarket is
...
+    } else {
+      feePercent = _uintBeacon.get(feeKey);
+    }
+    if (feePercent > FEE_LIMIT) revert FeePercentTooHigh();
```
XavierEkkel, line 300: instead of reverting, just set at the FEE_LIMIT in this case
and maybe this means we can remove this error

XavierEkkel (approved):
LGTM after comments

PR #1419, opened 2023-06-07
feat: FallthroughProviders

chrisling-dev:
https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=00b907a28d144d1d9dadd16c96b536ea&p=c901e4d0bab843dc86cbe391cd264e88&pm=s


packages/prepo-stores/src/utils/CustomProviders.ts
```diff
@@ -0,0 +1,74 @@
...
+      // trigger await here to catch errors
+      const res = await Promise.any(performs)
+      return res
+    } catch (e) {
+      // this catch block will only be reached if all RPCs failed
+      // TODO: remove all toasts from individual calls, and toast one site wide error instead
+      // alternatively, consider having an rpcFailed flag which triggers
+      // a different UI when all RPCs are down (e.g. prompt user to add custom RPC)
+      this.options.onError?.(e as Error)
+      return '0x00'
```
MeLlamoPablo, line 62: I'm not sure what the implications of returning `0x00` are. Why not throw the first RPC's error and let the local error handlers do their work?

MeLlamoPablo (commented):
Preliminary review: I like the direction in which this is headed!

We should make sure to thoroughly test `sendTransaction` methods. I'm not sure if calling two write methods at once in multiple providers could have negative implications.

packages/prepo-stores/src/utils/CustomProviders.ts
chrisling-dev: 0x00 is the result we will get if the request fails (can verify this by adding a faulty rpc and checking the res)

we definitely don't want to throw error here because this perform only overrides read functions, I'll elaborate more on this later, most of our read calls don't handle error. So if this starts throwing error, the entire dapp will start throwing a bunch of error (imagine every single read functions throwing errors)

This catch block will only be reached if in 2 situations:
- both endpoints are broken
- there is programming error that shouldn't be released into production

Realistically it would be very hard to distinguish between these 2 situations just from the error object, so I added `onError` which will be passed in from Web3Store, which allows us to handle the error once instead of having every single read functions throwing the same error

On second thought, it should return undefined instead so the UI would show loading screen, then onError can toast an "RPC Unavailable" error once so the UI is more intuitive

chrisling-dev:
> We should make sure to thoroughly test sendTransaction methods. I'm not sure if calling two write methods at once in multiple providers could have negative implications.

i have tested this from our exploration yesterday, when a sendTransaction happens, it happens via the user's wallet, so none of the changes here affects sendTransaction

You can verify this by checking the ContractStore's sendTransaction function, and see that a write contract is initiated for sendTransaction call using the Signer instead of the provider

Also can verify by making a transaction and check that no new requests are sent from the dapp, and that it is sent from the wallet

@MeLlamoPablo

packages/prepo-stores/src/utils/CustomProviders.ts
MeLlamoPablo: What I don't understand though is that if we are in this catch block it's because all the original `.perform` calls errored too. It's not us throwing the error, it's all of the RPCs. But if the original `.perform` successfully return `0x00` instead of throwing, shouldn't we handle that too?

chrisling-dev: > What I don't understand though is that if we are in this catch block it's because all the original `.perform` calls errored too. It's not us throwing the error, it's all of the RPCs. But if the original `.perform` successfully return `0x00` instead of throwing, shouldn't we handle that too?

oh wait sry I misundertood the question and and the behaviour ytd, so when an error occurs the perform will throw, it will not return 0 so returning 0x00 here was wrong, hence updated to throwing an error today

MeLlamoPablo (approved):
Great job! Tested it and works great, also all e2e are passing when you run them with the changes in #1432

I think we may want to tone down the amount of requests we send in the future because I think that sending all requests to all 3 on all times is overkill, but for now it's fine. I have some ideas (e.g. pick "winners" after a couple of requests and only use the "winners"), we can discuss them with more time.

PR #1424, opened 2023-06-08
refactor: move beacon usage to separate shared module

ramenforbreakfast:


ramenforbreakfast:
removing since we do not set beacons inside markets anymore

PR #1428, opened 2023-06-08
feat: add crust pinning

MeLlamoPablo:
This PR adds pinning on Crust to the `deploy-fe` workflow. We are currently using a temporary account controlled by me. We will migrate to our official hot wallet in a future PR.


chrisling-dev (approved):
LGTM

PR #1429, opened 2023-06-09
feat: custom rpc logic improvement

chrisling-dev:
This PR contains the logic of adding custom RPC to FallthroughProviders

https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=00b907a28d144d1d9dadd16c96b536ea&p=8f489d4997ab40dfadc691274c4a8311&pm=s


packages/prepo-stores/src/Web3Store.ts
```diff
@@ -75,6 +74,40 @@ export class Web3Store {
...
 
+  private initCustomRpc(): void {
+    const customRpc = window.localStorage.getItem('rpcs')
+    if (customRpc) this.customRpc = JSON.parse(customRpc)
+    this.syncCustomRpc()
+
+    reaction(
+      () => ({ rpcs: this.customRpc, network: this.network }),
+      () => this.syncCustomRpc()
+    )
```
MeLlamoPablo, line 85: Won't this reaction run the first time? Do we need to call it before too?

MeLlamoPablo (approved):
LGTM after fixing CI

packages/prepo-stores/src/Web3Store.ts
chrisling-dev: Nope, as stated in the documentation: https://mobx.js.org/reactions.html#reaction

> Unlike autorun, the side effect won't run once when initialized, but only after the data expression returns a new value for the first time.

We need to setup a value for customRpc before reaction would react to it

PR #1432, opened 2023-06-09
fix: price impact error on end-to-end tests

MeLlamoPablo:
Some tests were failing because we were hitting the price impact error by attempting to deposit 100 ETH.


chrisling-dev (approved):
LGTM

PR #1433, opened 2023-06-09
feat: add `cleanup-infura` workflow

MeLlamoPablo:
This PR adds a `cleanup-infura` workflow that aims to remove old, outdated pins from Infura. It accepts a `MAX_PINS` variable and removes the oldest pins until only that many pins are kept. Initially, we will start with `MAX_PINS: 5` to ensure that the job works properly. Then, we should increase it to around 50.

The workflow automatically runs once per week, and we can also trigger it manually.

The Infura IPFS API doesn't expose the "pin time" of a given version, so instead a `BUILD_TIME.json` file is added to each deployment, containing the UNIX timestamp of the build time. Then, the workflow uses that information to determine the oldest pins.

[Here's an example of a successful run](https://github.com/prepo-io/prepo-monorepo/actions/runs/5223994025/jobs/9431630076).


chrisling-dev (approved):
LGTM!

PR #1435, opened 2023-06-12
feat: byo rpc UI

chrisling-dev:
https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=00b907a28d144d1d9dadd16c96b536ea&p=7cf7d063c90741bfb6a2e8f656859e5b&pm=s


packages/prepo-stores/src/utils/CustomProviders.ts
```diff
@@ -36,6 +36,14 @@ export class FallthroughProvider extends ethers.providers.BaseProvider {
   }
 
   setCustomProvider(provider?: ethers.providers.JsonRpcProvider): void {
+    if (this.okProviders.length === 0) {
+      // calls would lose observability if all providers were previously failing
```
MeLlamoPablo, line 40: Could you elaborate on this problem? I'm not sure I understand it.

MeLlamoPablo (changes requested):

packages/prepo-stores/src/utils/CustomProviders.ts
chrisling-dev: so basically if all the default list is failing, and you try to add a custom RPC to recover it, nothing will react. You'll see the network requests to failing RPCs are stopped and it's now sending to new RPC, but the read values are no longer updated. I still don't know what the root cause is yet, wasn't able to find out

In the case of Web3Provider it was because a `called` flag is preventing the call from being retriggered but that fix doesn't solve this, so the best solution I have so far is to refresh the page so the FallthroughProviders can be re-initialized. Explained the impact on UX of this refresh

chrisling-dev: nvm I found a solution to this after some further exploration

apps/frontend/core-dapp/src/components/SettingsMenu/RpcSettings.tsx
```diff
@@ -32,24 +33,88 @@ const Description = styled.p`
...
+    if (status.error)
+      return (
+        <Flex color="error">
+          <Tooltip overlay={status.error} placement="topRight">
+            <Icon name="cross" height="16" width="16" />
+          </Tooltip>
+        </Flex>
+      )
+    return null
+  }
```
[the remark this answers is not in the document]
chrisling-dev: I don't agree with creating a component within RpcSettings as that makes the code very confusing IMO

I don't think this necessary has to be a component and this is a pretty common pattern, it's pretty much the same as doing this but returning a JSX instead of string:

```
const buttonText = useMemo(() => {
   if(loading) return 'Loading'
   if(InsufficientLiquidity) return 'Insufficient Liquidity'
   return "Trade"
}, [...])

return <Button>{buttonText}</Button>
```
 We can also do useMemo here like in the example I gave above but the computation here isn't expensive at all

If we want to do <StatusIcon /> we should do it outside the component with proper props passing but that'd be overkill for a component that isn't reused

packages/prepo-ui/src/components/Spinner/index.tsx
```diff
@@ -0,0 +1,16 @@
+import SpinFC from 'antd/lib/spin'
+import { LoadingOutlined } from '@ant-design/icons'
+import { Color, useTheme } from 'styled-components'
+
```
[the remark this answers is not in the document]
chrisling-dev: Good suggestion, the type will be inferred, no? And we can just import `@ant-design/icons/LoadingOutlined` instead

```
const SpinFC = dynamic(() => import('antd').then(({ Spin }) => Spin))
const LoadingOutlined = dynamic(() => import('@ant-design/icons/LoadingOutlined'))

```

MeLlamoPablo (approved):
LGTM!

PR #1436, opened 2023-06-12
feat: add wallet provider to FallthroughProviders

chrisling-dev:
https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=00b907a28d144d1d9dadd16c96b536ea&p=ac7e08fcf5f646bd99d55dd8f24df6d1&pm=s


packages/prepo-stores/src/ContractStore.ts
```diff
@@ -209,7 +209,9 @@ export class ContractStore<RootStoreType, SupportedContracts> {
             this.contract.functions[methodName](...params)
               .then(onFirstSet)
               .catch((error) => {
-                if (isImportantError(error)) throw this.root.captureError(error)
+                this.called[methodName][paramStr] = false
+                if (isImportantError(error) && process.env.NEXT_PUBLIC_VERCEL_ENV !== 'production')
```
MeLlamoPablo, line 213: Should use `isProduction` instead?

packages/prepo-stores/src/utils/CustomProviders.ts
```diff
@@ -44,7 +50,31 @@ export class FallthroughProvider extends ethers.providers.BaseProvider {
...
+        if (walletProviderIndex < 0) this.providers = [provider, ...this.providers]
+      } else {
+        const newProviders = this.providers
+        if (walletProviderIndex > -1) {
+          newProviders.splice(walletProviderIndex, 1)
+          this.providers = newProviders
+        }
+      }
+    })
+  }
```
MeLlamoPablo, line 74: I think a more simple way of doing this would be to have:

```typescript
class FallThroughProviders {
  // ...

  private _walletProvider?: ethers.providers.Web3Provider
  private _defaultProviders: ethers.providers.JsonRpcProvider[] = []

  get providers(): ethers.providers.JsonRpcProvider[] {
    if (this._walletProvider)
      return [this._walletProvider, ...this._defaultProviders]
    return this._defaultProviders
  }

  setWalletProvider(provider?: ethers.providers.Web3Provider): void {
    this._walletProvider = provider
  }

  // ...
}
```

This way we don't have to worry about checking whether the wallet provider is or isn't set, or manipulating indexes.

MeLlamoPablo (changes requested):

packages/prepo-stores/src/ContractStore.ts
chrisling-dev: this is ContractStore and isProduction isn't exposed here

packages/prepo-stores/src/utils/CustomProviders.ts
chrisling-dev: That was my initial approach but there was some observability issue (which I later fixed in contract store), let me try this again with that fix

packages/prepo-stores/src/ContractStore.ts
chrisling-dev: Resolving this because the fix is brought forward to previous PR

MeLlamoPablo: > this is ContractStore and isProduction isn't exposed here

Didn't realize this, sorry!

MeLlamoPablo (approved):

PR #1443, opened 2023-06-13
chore: add error logging when all RPCs error

MeLlamoPablo:
There are two causes for the "No RPCs available" error:

* All RPCs are erroring because none is available
* All RPCs are erroring because we have an error in our requests

The latter isn't likely to happen on produiction, but could happen during local development if we do something wrong (e.g. call the wrong contract), so this logging will be helpful in such cases, instead of just logging an obscure "No RPCs available" error.


packages/prepo-stores/src/utils/CustomProviders.ts
```diff
@@ -84,6 +84,10 @@ export class FallthroughProvider extends ethers.providers.BaseProvider {
       // alternatively, consider having an rpcFailed flag which triggers
       // a different UI when all RPCs are down (e.g. prompt user to add custom RPC)
       this.options.onError?.(e as Error)
+      if (e instanceof AggregateError) {
+        // eslint-disable-next-line no-console
+        e.errors.forEach((error) => console.warn(error))
+      }
```
chrisling-dev, line 90: I think it should just throw the error object if it's not production, otherwise we might still miss an error warning from the console, wdyt?

```suggestion
      if (isProduction) {
        throw e
      } else {
        throw Error('No RPCs available at the moment.')
      }
```

chrisling-dev (changes requested):

packages/prepo-stores/src/utils/CustomProviders.ts
```diff
@@ -84,6 +84,11 @@ export class FallthroughProvider extends ethers.providers.BaseProvider {
       // alternatively, consider having an rpcFailed flag which triggers
       // a different UI when all RPCs are down (e.g. prompt user to add custom RPC)
       this.options.onError?.(e as Error)
+      if (isProduction) {
+        throw e
+      } else {
+        throw Error('No RPCs available at the moment.')
+      }
```
chrisling-dev, line 91: well actually these lines can be ignored because there's already the line to throw no rpcs available error
```suggestion
      }
```

chrisling-dev (approved):
LGTM after fixing build

PR #1444, opened 2023-06-13
refactor: handle v1.1 versions of `TokenSender`, `DepositHook` and `WithdrawHook`

MeLlamoPablo:
https://www.notion.so/prepo/SCv1-1-Add-TokenSenderv1_1-DepositHookv1_1-WithdrawHookv1_1-DepositTradeHelperv1_1-d044b7bfb480482eb51841c57839a235

## Summary

This PR adds support for the v1.1 versions of `TokenSender`, `DepositHook` and `WithdrawHook`, while still maintaining v1.0 compatibility. This PR does not yet add compatibility for the v1.1 versions of `DepositAndTradeHelper` nor `MarketHook` (f.k.a `RedeemHook` in v1.0) - those will be added in a subsequent PR.

## Before and after

The PPO reward calculation changes slightly on the Deposit and Withdraw pages, as you can see below: _(left: v1.0, right: v1.1)_

image

v1.1 seems to be slightly more precise.

## TokenSender v1.1 differences

* `TokenSender` v1.0 worked by scaling the PPO token price (`scaledPrice`) locally and using that price to compute the PPO reward. This required multiple `TokenSender`s
* `TokenSender` v1.1 is only aware of the PPO token price and doesn't scale it locally. This allows us to only have one single `TokenSender`. Thus, callers of this `TokenSender` need to scale the reward amount before passing it on.

    Callers of `TokenSender` (like the deposit and withdraw hooks) are now `TokenSenderCaller` and define an `amountMultiplier` that is used to scale the reward amount. 

  This is represented on the front-end code as a separate `TokenSenderCallerEntity` (even though it's the same contract, but this way we can reuse the code) which exposes the same API as `TokenSender` for backwards compatibility.

## Store versioning

In order to maintain v1.0 compatibility while still seamlessly transitioning to v1.1 without needing to push updates to the front-end, the `DepositHookVersionManager` and `WithdrawHookVersionManager` classes are introduced. These middleman classes take the contract address and return the v1.0 version if the address matches that of the current production contracts, or the v1.1 version if it's a new address.

The version manager classes also provide a helper (the `tokenSender` computed prop) to ensure that the rest of the codebase can interact with the v1.0/v1.1 stores without needing to be aware of their version.

There is no need for a `TokenSenderVersionManager` as it is assumed that the new hooks will only integrate with v1.1 the `TokenSender`.

After the v1.1 migration is done (by updating the deposit/withdraw hooks on `Collateral`) we will be able to remove the version manager and the v1.0 version of the stores.

### Testing the v1.0 contracts

Simply run the local server or go to the preview. Everything should keep working as usual. The end-to-end tests are all passing.

### Testing the v1.1 contracts

This PR also expands the `yarn local-devnet` contract with support for "bootstrap recipes". With this feature, we can pass any arbitrary code snippet to prepare the Tenderly DevNet.

In this case, to test the v1.1 contracts we need to:

* As the prePO Multisig, fund the new `TokenSender` with enoguh PPO
* As the prePO deployer, set the PPO rebate amount multipliers on the hooks:
  * x2 on the deposit hook
  * x1.5 on the withdraw hook
* As the prePO Multisig, update the deposit and withdraw hooks on `Collateral`

Here's the snippet that does this:

```js
const deployer = impersonate(address("PREPO_DEPLOYER"))
const multisig = impersonate(address("PREPO_MULTISIG"))

const collateral = factories.CollateralAbi__factory.connect(address("COLLATERAL"), multisig)
const ppo = factories.Erc20Abi__factory.connect(address("PPO"), multisig)
const depositHook = factories.DepositHookV11Abi__factory.connect("0xCEcEBFe08206AF06cc047003B9bd9176689D0BEC", deployer)
const withdrawHook = factories.WithdrawHookV11Abi__factory.connect("0x564dA7be6702b6667fdf2C4f26E77482d4BCA64A", deployer)

await ppo.transfer(await depositHook.getTokenSender(), parseEther("1000000"))

await depositHook.setAmountMultiplier(address("COLLATERAL"), 2000000) // x2
await collateral.setDepositHook(depositHook.address)

await withdrawHook.setAmountMultiplier(address("COLLATERAL"), 1500000) // x1.5
await collateral.setWithdrawHook(withdrawHook.address)
```

We can use this recipe by feeding it to the script's stdin, like this:

```bash
cat << EOF | yarn local-devnet

const deployer = impersonate(address("PREPO_DEPLOYER"))
const multisig = impersonate(address("PREPO_MULTISIG"))

const collateral = factories.CollateralAbi__factory.connect(address("COLLATERAL"), multisig)
const ppo = factories.Erc20Abi__factory.connect(address("PPO"), multisig)
const depositHook = factories.DepositHookV11Abi__factory.connect("0xCEcEBFe08206AF06cc047003B9bd9176689D0BEC", deployer)
const withdrawHook = factories.WithdrawHookV11Abi__factory.connect("0x564dA7be6702b6667fdf2C4f26E77482d4BCA64A", deployer)

await ppo.transfer(await depositHook.getTokenSender(), parseEther("1000000"))

await depositHook.setAmountMultiplier(address("COLLATERAL"), 2000000) // x2
await collateral.setDepositHook(depositHook.address)

await withdrawHook.setAmountMultiplier(address("COLLATERAL"), 1500000) // x1.5
await collateral.setWithdrawHook(withdrawHook.address)

EOF
```

Simply copy-paste that command into your terminal and it will spawn a Tenderly RPC with the v1.1 contracts, then launch the app. You will be able to test the advertised PPO rebates.

### Testing depositing/withdrawing with the v1.1 contracts

This is beyond the scope of this PR, because it first requires supporting `DepositTradeHelper v1.1`. It would also require to upgrade the code of `Collateral` in the Tenderly DevNet, as `Collateral v1.0` is incompatible with the v1.1 hooks.


apps/frontend/core-dapp/src/stores/entities/FeeEntity.ts
```diff
@@ -42,7 +43,7 @@ export class FeeStore {
   }: {
     applicableToWstEth: () => BigNumber | undefined
     fee: Fraction
-    tokenSender: TokenSenderEntity | undefined
+    tokenSender: TokenSenderEntity | TokenSenderCallerEntity | undefined
```
chrisling-dev, line 46: So based on discussion, an idea is to make this hook instead of tokenSender, which only requires the prop to meet requirement of 1 interface, something like this

```
interface FeeHook { 
  calculateReward: (feeInWstEth: BigNumber) => BigNumber
  calculateRewardValueInEth: (ppo: BigNumber) => BigNumber
}
```

```suggestion
    hook: FeeHook
```

apps/frontend/core-dapp/src/stores/DepositHookStore.ts
```diff
@@ -9,12 +10,14 @@ type DepositsAllowed = DepositHookAbi['functions']['depositsAllowed']
...
+
   constructor(public root: RootStore, addressGetter: () => string | undefined) {
     super(root, addressGetter, DepositHookAbi__factory)
     makeObservable(this, { depositsAllowed: computed })
   }
 
-  get tokenSender(): string | undefined {
+  private getTokenSender(): string | undefined {
     return this.call<GetTokenSender>('getTokenSender', [])?.[0]
   }
```
chrisling-dev, line 22: and the old hook can have extra functions like:
```
  calculateReward(feeInWstEth: BigNumber): BigNumber | undefined {
    return this.tokenSender.calculateReward(feeInWstEth)
  }
```

apps/frontend/core-dapp/src/stores/DepositHookVersionManager.ts
```diff
@@ -0,0 +1,36 @@
...
+    if (this.v1_1.address !== undefined) return this.v1_1
+    return this.v1_0
+  }
+
+  get tokenSender(): TokenSenderEntity | TokenSenderCallerEntity {
+    return this.value instanceof DepositHookStore_v1_1
+      ? this.value.tokenSenderCaller
+      : this.value.tokenSender
+  }
+}
```
chrisling-dev, line 36: then we don't have to worry about this because the FeeEntity will just take the hook, which is the value above

apps/frontend/core-dapp/src/stores/DepositHookStore_v1_1.ts
```diff
@@ -0,0 +1,28 @@
...
+import { makeObservable, computed } from 'mobx'
+import { RootStore } from './RootStore'
+import { TokenSenderCallerEntity } from './entities/TokenSenderCallerEntity'
+import { SupportedContracts } from '../lib/contract.types'
+import { DepositHookV11Abi, DepositHookV11Abi__factory } from '../../generated/typechain'
+
+type GetDepositsAllowed = DepositHookV11Abi['functions']['getDepositsAllowed']
+
+// eslint-disable-next-line @typescript-eslint/naming-convention
+export class DepositHookStore_v1_1 extends DynamicContractStore<RootStore, SupportedContracts> {
```
chrisling-dev, line 11: and hookstore would extend TokenSenderCallerEntity instead of creating a new TokenSenderCaller store within it, so it shares all the functions in TokenSenderCallerEntity including the calculateReward functions

chrisling-dev (changes requested):
would be even more ideal IMO if we don't have to worry about HookVersionManager (now that both hooks share the same interface FeeEntity requires) and just have

```
get depositHook(): DepositHookStore | DepositHookStore_v1_1 {
   if(hookAddress === oldAddress) return this.depositHook_v1_0
   return this.depositHook_v1_1 
}
```

where depositHook_v1_0 and depositHook_v1_0 are created similarly to how they are now in DepositHookVersionManager

apps/frontend/core-dapp/src/stores/CollateralStore.ts
```diff
@@ -18,9 +20,39 @@ type GetWithdrawHook = CollateralAbi['functions']['getWithdrawHook']
...
 
+const DEPOSIT_HOOK_V1_0_ADDRESS = '0x1Da0a54D7130DA06b7C41154A1ccBd98eEe9007E'
+const WITHDRAW_HOOK_V1_0_ADDRESS = '0x48b765f21ce8cE70c6B08E2eF0f4098EBE808C22'
+
 export class CollateralStore extends Erc20PermitStore {
-  readonly depositHook = new DepositHookStore(this.root, () => this.getDepositHook()?.[0])
-  readonly withdrawHook = new WithdrawHookStore(this.root, () => this.getWithdrawHook()?.[0])
+  private readonly depositHook_v1_0: DepositHookStore = new DepositHookStore(this.root, () => {
+    const address = this.getDepositHook()?.[0]
+    if (address !== DEPOSIT_HOOK_V1_0_ADDRESS) return undefined
```
chrisling-dev, line 29: should the address check be converted to .toLowerCase?

chrisling-dev (changes requested):
minor comments

chrisling-dev (approved):
LGTM after comments

PR #1446, opened 2023-06-15
chore: migrate IPNS key to GitHub Secrets

MeLlamoPablo:


apps/frontend/core-dapp/src/pages/_app.tsx
```diff
@@ -15,6 +15,8 @@ import TermsWall from '../features/terms/TermsWall'
 import { TestIds } from '../components/TestId'
 import VersionsWall from '../components/VersionsWall'
 
+// Force front-end redeployment - TODO remove this comment on a future PR
```
chrisling-dev, line 18: hmm does a comment trigger redeployment at all?

chrisling-dev (approved):
LGTM

apps/frontend/core-dapp/src/pages/_app.tsx
MeLlamoPablo: Because otherwise `turbo-ignore` would skip the commit

PR #1448, opened 2023-06-15
feat: enable region blocking

MeLlamoPablo:


apps/frontend/core-dapp/src/utils/compose.tsx
```diff
@@ -0,0 +1,14 @@
+import { ComponentType, ReactNode } from 'react'
+
+type Component = ComponentType<{ children?: ReactNode }>
+
+export function compose(...[Base, ...other]: [Component, Component, ...Component[]]): Component {
```
chrisling-dev, line 5:
```suggestion
export function compose(...[Base, ...other]: Component[]): Component {
```

apps/frontend/core-dapp/src/utils/compose.tsx
```diff
@@ -0,0 +1,14 @@
...
+  const Other = other.length === 1 ? other[0] : compose(...(other as [Component, Component]))
```
chrisling-dev, line 6:
```suggestion
  const Other = other.length === 1 ? other[0] : compose(...other)
```

apps/frontend/core-dapp/src/utils/region.ts
```diff
@@ -0,0 +1,29 @@
...
+  } = await import('moment-timezone/data/meta/latest.json')
+
+  return Object.entries(zones).reduce(
+    (acc, [zoneName, { countries: zoneCountries }]) => ({
+      ...acc,
+      [zoneName]: countries[zoneCountries[0]].name,
+    }),
+    {}
+  )
+}
```
chrisling-dev, line 19: I think we should simply import this rather than lazing loading it:

Complexity: making this a Promise adds a fair bit of complexity to this feature and UX, whether a region is banned should be instantly available the moment the dapp is launched (e.g. what happens if the app is loaded but the region list is still being loaded, do users see the UI first then it jumps to region wall?) 

Performance: the region list is about 18kb, _app.js is sitting at about 450 - 500kb, I don't think simply importing this would increase the loading time significantly

apps/frontend/core-dapp/src/components/RegionWall.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+const Wrapper = styled.div`
+  color: ${({ theme }): string => theme.color.neutral4};
+  text-align: center;
+`
+
+export const useRegionBlocked = (): boolean => {
+  const regionBlock = usePromise(isUserInBannedRegion)
+  return regionBlock.status === 'success' && regionBlock.value
+}
```
chrisling-dev, line 13: then we can greatly simplify this by removing the usePromise hook and use isUserInBannedRegion directly (maybe cache it in a variable like isProduction and reuse that variable)

apps/frontend/core-dapp/src/components/layout/Header.tsx
```diff
@@ -46,10 +47,11 @@ const Banner = styled.div`
 const bannerContent: ReactNode | undefined = undefined
 
 const Header: React.FC = () => {
+  const regionBlocked = useRegionBlocked()
```
chrisling-dev, line 50: and don't need this hook, simply import isRegionBlocked like isOutdated and isProduction

chrisling-dev (changes requested):

apps/frontend/core-dapp/src/utils/compose.tsx
MeLlamoPablo: The reason it's done like that is to enforce a minimum of two elements on compile time, rather than runtime

PR #1450, opened 2023-06-16
feat: isolate historical data logic

chrisling-dev:
https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=1160865fbe4c4639af9d63918c7b7d17&p=acda45d21e0543b3bf3ae902c232ec79&pm=s


apps/frontend/core-dapp/src/stores/entities/UniswapV3PoolHistoricalData.ts
```diff
@@ -0,0 +1,212 @@
...
+ * It is designed this way to allow fetching the latest data every block without having to refetch the entire historical data
+ * Because fetching the entire historical data is both computationally expensive and slow
+ *
+ * This store handles caching historical data, so we don't have to refetch them when timeframe is switched,
+ * and intelligently subscribe to latest data of the selected timeframe
+ */
+export class UniswapV3PoolHistoricalData {
+  cachedTimeframe = ChartTimeframe.DAY
+  cachedHistoricalData?: SeparatedMarketHistoryData
+  constructor(private root: RootStore, public market: MarketEntity) {
```
MeLlamoPablo, line 59:
```suggestion
  constructor(private root: RootStore, public market: BaseMarketEntity) {
```

Should probably use `BaseMarketEntity`? This class doesn't reference any property specific to `MarketEntity`

MeLlamoPablo (approved):

apps/frontend/core-dapp/src/stores/entities/UniswapV3PoolHistoricalData.ts
chrisling-dev: yeap good point

PR #1451, opened 2023-06-16
fix: wallet provider conflicting with Tenderly

MeLlamoPablo:
Having the `walletProvider` in `FallthroughProviders` makes testing with Tenderly harder because sometimes the wallet provider wins the request race and sometimes it doesn't, so whatever changes we're testing in the Tenderly DevNet won't be always persisted.

The solution is simply to stop sending requests to the wallet provider if we're using Tenderly.


packages/prepo-stores/src/utils/CustomProviders.ts
```diff
@@ -18,23 +18,33 @@ export class TenderlyRpcProvider extends ethers.providers.JsonRpcProvider {
 }
 
 type FallthroughProviderOptions = {
+  includeWalletProvider?: boolean
```
chrisling-dev, line 21: I think simpler if this is `excludeWalletProvider`, then in constructor we don't need to re-assign to option parameter and can remove the DEFAULT_OPTIONS because excludingWalletProvider is falsy by default

```suggestion
  excludeWalletProvider?: boolean
```

packages/prepo-stores/src/utils/CustomProviders.ts
```diff
@@ -18,23 +18,33 @@ export class TenderlyRpcProvider extends ethers.providers.JsonRpcProvider {
...
   _providers: ethers.providers.JsonRpcProvider[] = []
   _walletProvider?: ethers.providers.Web3Provider
   customProvider: ethers.providers.JsonRpcProvider | undefined = undefined
+  private readonly options: FallthroughProviderOptions
 
   constructor(
     defaultProviders: ethers.providers.JsonRpcProvider[],
     network: Networkish | Promise<Network>,
-    private options: FallthroughProviderOptions = {}
+    options: FallthroughProviderOptions = {}
```
chrisling-dev, line 40: and can undo all changes here around options

packages/prepo-stores/src/utils/CustomProviders.ts
```diff
@@ -58,7 +68,9 @@ export class FallthroughProvider extends ethers.providers.BaseProvider {
   }
 
   get providers(): ethers.providers.JsonRpcProvider[] {
-    return this._walletProvider ? [this._walletProvider, ...this._providers] : this._providers
+    return this._walletProvider && this.options.includeWalletProvider
```
chrisling-dev, line 71: and simply do this:
```suggestion
    return this._walletProvider && !this.options.excludeWalletProvider
```

packages/prepo-stores/src/utils/createFallthroughProvider.ts
```diff
@@ -27,7 +27,9 @@ export const createFallthroughProvider = (network: Network): FallthroughProvider
    */
   if (rpcOverride?.startsWith('https://rpc.vnet.tenderly.co')) {
     const tenderlyRpcProvider = new TenderlyRpcProvider(rpcOverride, network.chainId)
-    return new FallthroughProvider([tenderlyRpcProvider], network.chainId)
+    return new FallthroughProvider([tenderlyRpcProvider], network.chainId, {
+      includeWalletProvider: false,
+    })
```
chrisling-dev, line 32:
```suggestion
    return new FallthroughProvider([tenderlyRpcProvider], network.chainId, {
      excludeWalletProvider: true,
    })
```

chrisling-dev (approved):
minor comment

PR #1452, opened 2023-06-16
[DON'T MERGE] feat: upgrade `DepositTradeHelper` to v1.1

MeLlamoPablo:
Blocked by #1472 

---

This PR upgrades our `DepositTradeHelper` to v1.1 and switches the contract to use the new v1.1 address.

Unlike #1444, this PR does not maintain the v1.0 and v1.1 code side by side. It simply removes the v1.0 code, as the `DepositTradeHelper v1.1` can be used with `Collateral v1.0`.


packages/prepo-stores/src/ContractStore.ts
```diff
@@ -103,6 +103,9 @@ export class ContractStore<RootStoreType, SupportedContracts> {
     // Craft and send the tx with the signer
     const writeContract = this.getWriteContract()
     try {
+      // TODO undefined porque tiene overloads
```
chrisling-dev, line 106: what's porque tiene haha

apps/frontend/core-dapp/abi/ppo.abi.json
```diff
@@ -0,0 +1,587 @@
+[
```
chrisling-dev, line 1: do we use this anywhere?

apps/frontend/core-dapp/abi/restrictedTransferHook.abi.json
chrisling-dev, line 1: ditto, do we use this anywhere?

chrisling-dev (changes requested):
Agree with the approach explained in PR description, I think we can simplify things by assuming the TokenSender is always whitelisted for PPO transfer, since this PR will only be merged after TokenSender is whitelisted

I assume the flow looks something like:
- whitelist and fund new TokenSender
- test and merge this PR
- remove whitelist for old TokenSender (not sure if this is necessary)

apps/frontend/core-dapp/abi/ppo.abi.json
MeLlamoPablo: Only on the deploy script mentioned in the PR:

```
const ppo = factories.PpoAbi__factory.connect(address("PPO"), multisig)
const transferHook = factories.RestrictedTransferHookAbi__factory.connect(await ppo.getTransferHook(), multisig)
const ppoSourceAllowlist = factories.AccountListAbi__factory.connect(await transferHook.getSourceAllowlist(), multisig)
```

I could add them on another PR, or remove after the changes on the SC side are ready, but I think it's nice to be able to test this PR without needing to do anything extra.

MeLlamoPablo:
> Agree with the approach explained in PR description, I think we can simplify things by assuming the TokenSender is always whitelisted for PPO transfer, since this PR will only be merged after TokenSender is whitelisted
> 
> I assume the flow looks something like?:
> 
>     * whitelist and fund new TokenSender
>     * test and merge this PR
>     * remove whitelist for old TokenSender (not sure if this is necessary)

We are assuming this in the code. The `TokenSender` being whitelisted is one of the three prerequisites for this PR being merged. The whole flow would be:

* Whitelist and fund new `TokenSender`
* Set `amountMultiplier` to `x3` in `DepositTradeHelper`
* Test and merge this PR

PR #1453, opened 2023-06-19
feat: handle redemption fee of MarketEntity v1.1

chrisling-dev:
https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=1160865fbe4c4639af9d63918c7b7d17&p=4e40a878c55845ffb887e9d2bddda10f&pm=s

MarketHook is also integrated here (which basically just extends TokenSenderCallerEntity) because if it's in separate PR the MarketEntityV1_1 wouldn't work
https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=1160865fbe4c4639af9d63918c7b7d17&p=b22ae668440c41e285755925a1906666&pm=s


apps/frontend/core-dapp/src/stores/entities/AddressBeaconEntity.ts
```diff
@@ -0,0 +1,16 @@
...
+type Get = AddressBeaconAbi['functions']['get']
+
+export class AddressBeaconEntity extends DynamicContractStore<RootStore, SupportedContracts> {
+  constructor(public root: RootStore, address: string) {
+    super(root, () => address, AddressBeaconAbi__factory)
+  }
+
+  get(...params: Parameters<Get>): ContractReturn<Get> {
+    return this.call<Get>('get', params)
+  }
```
MeLlamoPablo, line 15: Should `get` be `observable`?

apps/frontend/core-dapp/src/stores/ResourceStore.ts
```diff
@@ -0,0 +1,26 @@
...
+import { RootStore } from './RootStore'
+
+/**
+ * Keeps track of all resources that could be reused by multiple contracts, like
+ * - AddressBeacon
+ * - TokenSender
+ * - ValueBeacon
+ */
+export class ResourceStore {
+  addressBeacons: Record<string, AddressBeaconEntity> = {}
```
MeLlamoPablo, line 12:
```suggestion
  addressBeacons: Record<string, AddressBeaconEntity | undefined> = {}
```

apps/frontend/core-dapp/src/stores/entities/BaseMarketEntity.ts
```diff
@@ -294,4 +297,9 @@ export class BaseMarketEntity
       maxDecimals: 2,
     })
   }
+
+  // eslint-disable-next-line class-methods-use-this
+  get redemptionFee(): Fraction | undefined {
+    throw new Error('redemptionFee not implemented')
+  }
```
MeLlamoPablo, line 304: I think it's better to make `BaseMarketEntity` `abstract` and then do:

```ts
abstract redemptionFee: Fraction | undefined
```

Then we can override this on `MarketEntity` and `MarketEntity_v1_1`

MeLlamoPablo (changes requested):

apps/frontend/core-dapp/src/stores/entities/AddressBeaconEntity.ts
chrisling-dev: yeap!

MeLlamoPablo (approved):
Unblocking this - LGTM after comments

apps/frontend/core-dapp/src/stores/entities/BaseMarketEntity.ts
chrisling-dev: Not sure why but making redemptionFee causes the logic in MarketEntity to be skipped and it keeps returning undefined (even console.log in MarketEntity's redemptionFee is skipped), I'll keep it untouched for now

PR #1455, opened 2023-06-19
chore: increase IPNS lifetime

MeLlamoPablo:
I have been observing that IPNS updates work well, so at this point we're ready to increase the IPNS lifetime. This will make our front-end ready for production use.


.github/workflows/publish-ipns.yml
```diff
@@ -45,7 +44,7 @@ jobs:
           ipfs name publish "$CID" \
             --resolve=false \
             --key prepo \
-            --lifetime 24h \
+            --lifetime 1440h \
```
chrisling-dev, line 47: just making sure, 1440h is 60 days, wasn't the plan to make it 30 days?

chrisling-dev (approved):
LGTM after comment

.github/workflows/publish-ipns.yml
MeLlamoPablo: 30 days is when the front-ends self-disable, but if we don't update the latest front-end, we still want it to be available, hence 60 days.

PR #1460, opened 2023-06-20
feat: add Safe{Wallet} compatibility

MeLlamoPablo:
This PR adds support for prePO to become a [Safe App](https://docs.safe.global/learn/safe-apps), which is a dApp that can be used from within [Safe{Wallet}](https://safe.global/) (f.k.a. Gnosis Safe).

The transition is pretty easy because there's a dedicated Onboard module that does the hard work for us. Here are the other changes that were necessary:

1. Added a `manifest.json`, which is a prerequisite for being listed on the curated Safe App list.
2. Added logic to auto-connect to Safe{Wallet} if we are inside it, which is another requirement.

    The ideal UX is to avoid anything connection-related. The user doesn't need to "connect" because they are already inside their wallet.
    * The Onboard initialization logic will wait until `Web3Store.detectSafeWallet` resolves, but this is okay because if the app is not inside an iframe, `detectSafeWallet` bails early. In practice, there won't be a loading time increase.
3. Hid the disconnect button if inside a Safe. Since we don't want a connection experience, we don't want the user to disconnect either.
4. Hid the connect button on the header, and disabled the connect button taking the place of the "Trade" button.

    These show briefly while the app loads, so we don't want to direct the user's attention to them.
5. This change is entirely optional, but I think it's a nice detail.

    Replaced Jazzicon with Blockies while we're inside a Safe, because this is what Safe uses. Safe's own generated avatar for the connected wallet and ours are very close apart, so I think it's neat that they speak the same visual language.

    Blockies will only load if the user is inside a Safe so the rest of the users won't load unnecessary libraries.

    Screenshot 2023-06-20 at 19 38 02


packages/prepo-stores/src/Web3Store.ts
```diff
@@ -287,4 +306,27 @@ export class Web3Store {
...
+
+      const { default: SafeAppsSDK } = await import('@safe-global/safe-apps-sdk')
+      const sdk = new SafeAppsSDK()
+
+      const safe = await Promise.race([
+        sdk.safe.getInfo(),
+        new Promise<undefined>((resolve) => {
+          setTimeout(resolve, 200)
+        }),
+      ])
```
chrisling-dev, line 323: cant we just wait for getInfo only?

apps/frontend/core-dapp/public/manifest.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "prePO",
+  "description": "Pre‑IPO & Pre‑Token Exchange",
+  "start_url": "/",
```
chrisling-dev, line 4: should the start_url be `/trade`? I was stuck at blank screen inside safe until I manually changed the url to /trade

chrisling-dev (commented):
LGTM after comments

chrisling-dev (approved):

apps/frontend/core-dapp/public/manifest.json
MeLlamoPablo: Interesting, that wasn't the case for me, but sure we can change it.

packages/prepo-stores/src/Web3Store.ts
MeLlamoPablo: The Safe API is pretty awkward in this regard - if the app isn't loaded inside a Safe, `sdk.safe.getInfo` will hang indefinitely, so we need to add a timeout. [This is taken from here](https://github.com/blocknative/web3-onboard/blob/dbddeb855e5ad88cfc3ed6acb864b9133beb80fd/packages/gnosis/src/index.ts#L46-L49) by the way.

PR #1463, opened 2023-06-21
refactor: simplify `makeErrorCapturer`

MeLlamoPablo:
The error capturer code was a bit confusing to me, and I found out that we never pass `errorCapturer` to `makeErrorCapturer` (whaterver that was supposed to be, maybe it opened the door for a sentry-like integration?), so I simplified it. I think our error handling mechanisms could be clearer and more unified but that would be a task for another time.


chrisling-dev (approved):
Nice work, yeap the errorCapturer was only for sentry such that we make the `captureError` function once in RootStore, and everytime we capture an error sentry will also capture it. Good to remove!

PR #1468, opened 2023-06-23
feat: add support for tx bundling in Safe{Wallet}

MeLlamoPablo:
## PR Train Visualization

```mermaid
flowchart RL
  a[<a href='https://github.com/prepo-io/prepo-monorepo/pull/1465'>feat/contract-wallet-support</a>] --> main
  d[<a href='https://github.com/prepo-io/prepo-monorepo/pull/1470'>feat/rest-of-tx-approvals</a>] --> a
  b[<a href='https://github.com/prepo-io/prepo-monorepo/pull/1468'>feat/safe-bundle-transactions</a>] --> a
  c[<a href='https://github.com/prepo-io/prepo-monorepo/pull/1469'>chore/deduplicate-safe-sdk</a>] --> b
```

## Summary

This PR expands the functionality of `TransactionBundleStore` to support bundling transactions in Safe{Wallet}. This enables users to execute Approve + Trade in just one transaction, enjoying a similar experience to EOAs that sign permits.

In order to achieve this, the transaction bundle is sent to Safe{Wallet} via `@safe-global/safe-apps-sdk` rather than via `ethers`. Then, a successful response from Safe{Wallet} containing the tx hash awaited, and finally the transaction flow resumes as usual.

The action button (Trade / Deposit and Trade) switches to "Execute/Confirm in Safe{Wallet}" to provide more context for the user.

## How to test

* Create a Safe{Wallet} or use an existing one
* Go to "Apps > My Custom Apps"
* Add this PR's preview URL as a custom app: https://core-dapp-eafklnp4d-prepo.vercel.app/
* Open it and make a trade (only open trades are supported for now)

Note: it is not possible to bundle transactions if the safe is connected via WalletConnect. In that case, the connected wallet will be treated as a generic smart contract, and Approving + Trading will be a two-step process.


apps/frontend/core-dapp/src/stores/TransactionBundleStore.ts
```diff
@@ -117,6 +132,56 @@ export class TransactionBundleStore {
...
+        const response = await safeAppsSdk.txs.send({
+          txs: [
+            {
+              to: approveTx.to,
+              data: hexlify(approveTx.data),
+              value: hexValue(approveTx.value ?? BigNumber.from(0)),
+            },
+            {
+              to: actionTx.tx.to,
+              data: hexlify(actionTx.tx.data),
```
chrisling-dev, line 161: cmiiw but data should always be in hex format right? hexlify seems redundant here?

apps/frontend/core-dapp/src/stores/TransactionBundleStore.ts
```diff
@@ -199,6 +264,61 @@ export class TransactionBundleStore {
...
+        const receipt = await retry(() => safeAppsSdk.txs.getBySafeTxHash(safeTxHash))
+
+        switch (receipt.txStatus) {
+          case TransactionStatus.AWAITING_CONFIRMATIONS: {
+            this.safeTxStatusMessage = 'Confirm in Safe{Wallet}'
+            break
+          }
+          case TransactionStatus.AWAITING_EXECUTION: {
+            this.safeTxStatusMessage = 'Execute in Safe{Wallet}'
+            break
```
chrisling-dev, line 289: these two should be wrapped with runInAction since they happen after an await

apps/frontend/core-dapp/src/stores/TransactionBundleStore.ts
```diff
@@ -199,6 +264,61 @@ export class TransactionBundleStore {
...
+            }
+          }
+        }
+
+        await new Promise((resolve) => {
+          setTimeout(resolve, 15_000)
+        })
+      }
+    } finally {
+      this.safeTxStatusMessage = undefined
```
chrisling-dev, line 316:
```suggestion
      runInAction(() => {
        this.safeTxStatusMessage = undefined
      })
      
```

apps/frontend/core-dapp/src/stores/TransactionBundleStore.ts
```diff
@@ -199,6 +264,61 @@ export class TransactionBundleStore {
...
+            return {
+              success: false,
+              error: 'Transaction execution failed. Check details in Safe{Wallet}.',
+            }
+          }
+        }
+
+        await new Promise((resolve) => {
+          setTimeout(resolve, 15_000)
+        })
```
chrisling-dev, line 313: Would this ever get executed? The switch already handles all possible cases for `receipt.txStatus` which will either break the loop or return the function. Also led me to thinking whether the `while` loop will ever run a second time since as long as receipt is returned, which doesn't seem like it can be undefined, the loop will break.

chrisling-dev (changes requested):
Haven't test this because the previous PR was breaking at permit signature step for EOA

apps/frontend/core-dapp/src/features/connect/ConnectButton.tsx
```diff
@@ -38,7 +38,7 @@ const ConnectButton: React.FC<Props> = ({ hideWhenConnected, ...buttonProps }) =
     // eslint-disable-next-line react/jsx-props-no-spreading
     <Button
       block
-      disabled={isSafeWallet || connecting}
+      disabled={!!safeAppsSdk || connecting}
```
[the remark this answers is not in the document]
MeLlamoPablo: No, we intentionally disable the button when the app is running in Gnosis Safe. This is because Gnosis Safe automatically connects, so we don't want the user to see an actionable button.

Notice that `isSafeWallet` is equivalent to `!!safeAppsSdk`

apps/frontend/core-dapp/src/stores/TransactionBundleStore.ts
```diff
@@ -73,10 +82,16 @@ export class TransactionBundleStore {
...
     const names = this.actionNames()
-    if (!canSign && hasAllowance === false) return 'Approve'
-    return names.join(' And ')
+    if (!canSign && hasAllowance === false) {
+      if (canBundle) {
+        if (transacting && safeTxStatusMessage) return safeTxStatusMessage
+        return enumerate('Approve', ...names)
+      }
+      return 'Approve'
+    }
```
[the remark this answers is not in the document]
MeLlamoPablo: Great catch, love this change!

apps/frontend/core-dapp/src/stores/TransactionBundleStore.ts
MeLlamoPablo: The `break`s in the first two `switch` cases will break the `switch`, not the loop. Here's a simple snippet demonstrating that:

```ts
for (let i = 0; i < 10; i++) {
  console.log("Loop", i)
  
  switch (i) {
    case 2: {
      console.log("Special case 2!")
      break
    }
    case 3: {
      console.log("Special case 3!")
      break
    }
  }

  if (i === 5) {
    console.log("I'm done with this loop")
    break
  }
}
```

Which prints:

```
Loop 0
Loop 1
Loop 2
Special case 2!
Loop 3
Special case 3!
Loop 4
Loop 5
I'm done with this loop
```

So yes, the 15s delay will run and after it the loop will run again, rinse and repeat until any of the returns are reached.

apps/frontend/core-dapp/src/stores/TransactionBundleStore.ts
MeLlamoPablo: `data` could also be `Bytes` which is `number[]` but the Safe SDK needs `string`

PR #1473, opened 2023-06-26
chore: add retrying to `cleanup-infura-pins`

MeLlamoPablo:
The Infura API seems to be a bit unreliable so added some retrying logic


apps/frontend/core-dapp/internals/cleanup-infura-pins.mjs
```diff
@@ -16,12 +17,14 @@ const Authorization = `Basic ${Buffer.from(
...
-  },
-}).then((response) => response.json())
+const pins = await retry(() =>
+  fetch('https://ipfs.infura.io:5001/api/v0/pin/ls?type=recursive', {
+    method: 'POST',
+    headers: {
+      Authorization,
+    },
+  })
+).then((response) => response.json())
```
chrisling-dev, line 27: Should we configure a max `retries` so it doesn't just keep retrying forever if something is actually wrong? (e.g. infura is down for 5 hours)

apps/frontend/core-dapp/internals/cleanup-infura-pins.mjs
```diff
@@ -31,8 +34,8 @@ const sortedCidList = (
   await Promise.all(
     cidList.map(async (cid) => ({
       cid,
-      time: await fetch(`https://cf-ipfs.com/ipfs/${cid}/BUILD_TIME.json`).then((response) =>
-        response.json()
+      time: await retry(() => fetch(`https://cf-ipfs.com/ipfs/${cid}/BUILD_TIME.json`)).then(
+        (response) => response.json()
```
chrisling-dev, line 38: ditto

apps/frontend/core-dapp/internals/cleanup-infura-pins.mjs
```diff
@@ -44,12 +47,14 @@ const toUnpin = sortedCidList.slice(0, -MAX_PINS)
...
-      },
-    })
+    await retry(() =>
+      fetch(`https://ipfs.infura.io:5001/api/v0/pin/rm?arg=${cid}`, {
+        method: 'POST',
+        headers: {
+          Authorization,
+        },
+      })
+    )
```
chrisling-dev, line 57: ditto

chrisling-dev (changes requested):

apps/frontend/core-dapp/internals/cleanup-infura-pins.mjs
MeLlamoPablo: By default `async-retry` [will only retry 10 times](https://github.com/vercel/async-retry#api).

chrisling-dev (approved):
LGTM

PR #1474, opened 2023-06-26
feat: add Starknet market (unlisted)

MeLlamoPablo:
Blocked by #1471

----

This PR adds the Starknet market to the front-end, unlisted but accessible via URL params.

⚠️ Note that the market pools aren't ready yet so we're using the zkSync pools as a placeholder


apps/frontend/core-dapp/src/features.ts
```diff
@@ -0,0 +1,9 @@
+/*
+ * If this is true, the Starknet market is publicly shown. (Right now it's only
+ * accessible via URL)
+ *
+ * ⚠️ WARNING: the Starknet market is currently using the zkSync Uniswap pools
+ *    because the Starknet pools aren't live yet. We shouldn't set this flag to
+ *    true until the pool contracts are updated.
+ */
+export const STARKNET_MARKET_PUBLICLY_AVAILABLE = false
```
chrisling-dev, line 9: I think let's not have flag like this that is one time use only, we can simply have this in the `markets` file like `unlisted: true` rather than `unlisted: !STARKNET...AVAILABLE`

apps/frontend/core-dapp/src/features/trade/open-trade/MarketSlideUp.tsx
```diff
@@ -19,12 +19,15 @@ const SelectedMarket = styled(SlideUpRow)`
...
+  const { listedMarkets } = marketStore
   const { allPositions } = portfolioStore
   const { slideUpContent, selectedMarket, setSlideUpContent, selectedPosition } = tradeStore
 
   const enabled =
-    Object.keys(unresolvedMarkets).length > 1 || !selectedMarket || selectedMarket.resolved
+    Object.keys(listedMarkets).length > 1 ||
+    !selectedMarket ||
+    selectedMarket.resolved ||
+    selectedMarket.unlisted
```
chrisling-dev, line 30: actually, since we're going to launch the market with this PR merged anyway, I think can just remove the `unlisted` property, and simply merge this PR with the pool addresses when everything is ready

apps/frontend/core-dapp/src/stores/MarketStore.ts
```diff
@@ -35,15 +35,15 @@ export class MarketStore {
...
   }
 
   setSearchQuery(query: string): void {
     this.searchQuery = query
   }
 
-  get unresolvedMarkets(): MarketEntity[] {
-    return Object.values(this.markets).filter((market) => !market.resolved)
+  get listedMarkets(): MarketEntity[] {
+    return Object.values(this.markets).filter((market) => !market.resolved && !market.unlisted)
```
chrisling-dev, line 47:
```suggestion
  get listedMarkets(): MarketEntity[] {
    return Object.values(this.markets).filter((market) => !market.resolved)
```

chrisling-dev (changes requested):
LGTM after removing `unlisted` flag, I think we'll just merge this PR with the correct pool addresses

chrisling-dev (approved):
LGTM

PR #1488, opened 2023-06-27
feat: cap the open and close trade inputs at 5 ETH

MeLlamoPablo:
This PR:

* Switches the cap of the open/close trade inputs from 99,999 to 5
* Enforces the cap on input blur rather than immediately
* Shows a warning if the cap is enforced


chrisling-dev (approved):
LGTM

PR #1493, opened 2023-07-05
fix: return short price for resolved markets

chrisling-dev:


ramenforbreakfast (approved):
LGTM, Im assuming it's OK that if the price is undefined, nothing is shown?

chrisling-dev:
> LGTM, Im assuming it's OK that if the price is undefined, nothing is shown?

Yeap, if price is undefined (market not resolved), it will try to load price from the pool. This is fine for all markets except for arbitrum because we removed pool addresses for arbitrum.

PR #1499, opened 2023-07-07
fix: install command

chrisling-dev:


ramenforbreakfast (approved):
Thanks for addressing this!

Some background. The recent work around establishing a hybrid foundry/hardhat repo requires the usage of direct importing of tarballs using relative paths.

This caused some issues with our vercel build jobs because they were configured to run `yarn install` locally within each workspace, which actually will still try to install the entire monorepo's package, but from the local directory, thus breaking any reference to a relative path.

This resolves the issue by always performing yarn install from the root directory, which should be done anyway.

PR #1500, opened 2023-07-11
fix:core: caching SC Github action runs results in false positives

ramenforbreakfast:
Vercel's built-in caching engine for determining when to skip re-running tests and rely on caching results is not currently working for our SC directories and is prematurely skipping tests even though contract files have changed.

This fix is meant to disable this caching so that tests are run as long as it is specified within the PR name.


XavierEkkel (approved):
LGTM if tested

PR #1502, opened 2023-07-11
feat: add Pausable contract back to prepo-shared-contracts

ramenforbreakfast:
This adds back the Pausable contract that was removed previously from our repo. It has been modified to not impose `SafeOwnable` on an inheriting contract.


packages/prepo-shared-contracts/contracts/interfaces/IPausable.sol
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: AGPL-3.0
+pragma solidity =0.8.7;
+
+/**
+ * @notice Allows a contract owner to pause their contract.
+ * @dev Pausing a contract will only affect functions using the included
+ * `whenNotPaused` modifier.
+ */
```
XavierEkkel, line 8: remove natspec for consistency with other interfaces

XavierEkkel (approved):
LGTM after comments

PR #1503, opened 2023-07-12
feat: update stale token repo to be compilable and removed old tests

ramenforbreakfast:
To start off the series of changes to the `Vesting` contract we need to first get the token project to be in a compilable/testable state with the following changes:
- Update the `AccountList.sol` copied into this repo to use the most recent `AccountList.sol`, since the interface has significantly changed.
- Removed the `AccountList.test.ts`, seems to be that this most likely existed prior to `prepo-shared-contracts` and survived as an independent test suite.
- Removed `PPOStaking.test.ts` and `PPOGamifiedToken.test.ts`, since token tests have not been migrated to use snapshots yet, and this is a low hanging fruit removal to speed up tests.

`Vesting.sol` changes
- Modified to work with the new `Pausable.sol`, since `Pausable.sol` follows our new convention of not imposing access controls, also added an inheritance of `SafeOwnable.sol`


apps/smart-contracts/token/test/Vesting.test.ts
```diff
@@ -1004,6 +1004,16 @@ describe('Vesting', () => {
     })
   })
 
+  describe('# setPaused', () => {
+    beforeEach(async () => {
+      await setupVesting()
+    })
+
+    it("doesn't revert", async () => {
```
XavierEkkel, line 1012: missing `reverts if not owner` test

XavierEkkel (approved):
LGTM after comment

PR #1504, opened 2023-07-13
feat: various minor Vesting contract changes

ramenforbreakfast:
This PR includes various minor Vesting contract changes and accompanying Vesting test changes:
- Remove usage of `SafeERC20`
- Removal of `setToken`, `setVestingStartTime`, and `setVestingEndTime`, moved these to the constructor
- Added error inside of constructor to prevent setting a start time >= the end time
- Modified `totalAllocatedSupply` to use an underscore
- Removed comments since the contract will most likely change and we do not want to deploy a contract with partial comments.

Test changes in addition to testing the above behavior:
- Removed the redundant separate definition of an `owner` to simplify the test setup
- Removed the `setupVesting` function, since everything is set at deployment
- Removed redundant multiple redefining of `testStartTime` and `testEndTime`, moved to just be set once in the `deployVesting` function.
- Renamed `vestingStartTime` and `vestingEndTime` variable to `testStartTime` and `testEndTime`, to prevent any possible namespace conflict with the actual initialization variable in the future if we need to define custom start/end times for a testcase.


apps/smart-contracts/token/test/Vesting.test.ts
```diff
@@ -577,7 +333,7 @@ describe('Vesting', () => {
     })
 
     it('reverts if vesting not started', async () => {
-      expect(currentTime).to.be.lt(vestingStartTime)
+      expect(currentTime).to.be.lt(testStartTime)
```
XavierEkkel, line 336:
```suggestion
      expect(currentTime).lt(testStartTime)
```

XavierEkkel (approved):
LGTM after comments

PR #1506, opened 2023-07-13
feat: add function to read vesting amounts at different times

ramenforbreakfast:
In order to ensure compatibility with governance voting systems, we need to be able to read vested amounts based on timestamps to enable vote weighting at a specific snapshot.

For `getClaimedAmount`, this will also require implementing snapshotting of claimed tokens, but that is more involved and will come in a later PR.

In addition, since I was modifying `getVestingAmount(address)` tests here, I identified an issue with the existing test in that for test cases that verify vested amounts during the period, we were simply asserting against the read value itself, making the testcase pointless.

To remedy this, I added a helper function to simply calculate the expected amount, since we also need it to test the new `getVestingAmount(address,uint256)` function that takes in specific timestamps.

I also went ahead and took care of another chore regarding the unnecessary usage of `as any` typing for `ethers.provider`, this was written before we fixed our utility to accept different provider types.


XavierEkkel (approved):
LGTM after comments

PR #1507, opened 2023-07-14
feat: add snapshotting to accurately calculate claimable amounts at any time

ramenforbreakfast:
Since we need to be able to distinguish vested amounts (but already claimed) versus those that have not been claimed, and know these amounts at any given time in the past for snapshot voting, we need to add snapshots and rework our `getClaimableAmount` and `getClaimedAmount` functions.

This PR is quite large, but these features really can't be detached since they are so tightly connected, here is a summary of these changes:
- Unfortunately, checkpoints are not an OZ module that I can import, however, I reused a lot of code present in https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v4.8.1/contracts/token/ERC20/extensions/ERC20Votes.sol
- This contract uses the same `Checkpoint` system, with some slight modifications, mainly to accomodate `timestamp`, rather than `blockNumber`. The choice to simplify the structure and simply have two `uint256`s, rather than tightly packing to save bits, is mostly to avoid having to perform casting and `uint32` being questionable for a timestamp.
- `getClaimedAmount` and `getClaimableAmount` now have 2 versions, one that gets the current value, and one that allows a specifiable timestamp. These both call `internal` versions to reduce code duplication.
- `_getClaimedAmount` uses a binary search algorithm to find the `Checkpoint` that is closest to the specified timestamp, this was taken from the OZ code linked to above.
- Note: You will see in the OZ code that a internal function called `_unsafeAccess` is used, which is an optimization for accessing array entries. This function was causing issues when I tried to reuse it for our purposes and so I decided to simply access the entries normally.


apps/smart-contracts/token/test/Vesting.test.ts
```diff
@@ -503,19 +509,97 @@ describe('Vesting', () => {
...
+        Array(numberOfClaims).keys(),
+        (x) => testStartTime + (x + 1) * timeBetweenEachWithdrawal
+      )
+      const amountToClaimEachTime = amountsAllocated[0].div(numberOfClaims)
+
+      /**
+       * Ensure invariants are not violated across multiple claims,
+       * for both retroactive and future timestamps.
+       */
+      for (let i = 0; i < numberOfClaims; i++) {
```
XavierEkkel, line 558: could these complicated loops be extracted out into a helper function?

apps/smart-contracts/token/test/Vesting.test.ts
```diff
@@ -503,19 +509,97 @@ describe('Vesting', () => {
...
+        await setNextTimestamp(ethers.provider, claimingTimes[i])
+        await vesting.connect(user1).claim()
+        // eslint-disable-next-line no-restricted-syntax
+        for (let j = 0; j < numberOfClaims; j++) {
+          if (j <= i) {
+            const claimedAmountJustBefore = await vesting['getClaimedAmount(address,uint256)'](
```
XavierEkkel, line 564:
```suggestion
            const previousClaimedAmount = await vesting['getClaimedAmount(address,uint256)'](
```

apps/smart-contracts/token/test/Vesting.test.ts
```diff
@@ -503,19 +509,97 @@ describe('Vesting', () => {
...
+      for (let i = 0; i < numberOfClaims; i++) {
+        await setNextTimestamp(ethers.provider, claimingTimes[i])
+        await vesting.connect(user1).claim()
+        // eslint-disable-next-line no-restricted-syntax
+        // Skip first entry since we only care about time in between claims
+        for (let j = 1; j < numberOfClaims; j++) {
+          expect(
+            await vesting['getClaimedAmount(address,uint256)'](user1.address, claimingTimes[j - 1])
+          ).eq(
+            await vesting['getClaimedAmount(address,uint256)'](user1.address, claimingTimes[j] - 1)
```
XavierEkkel, line 595:
```suggestion
            await vesting['getClaimedAmount(address,uint256)'](user1.address, claimingTimes[j] - amountsAllocated[0])
```

apps/smart-contracts/token/test/Vesting.test.ts
```diff
@@ -503,19 +509,97 @@ describe('Vesting', () => {
...
       await expect(tx)
         .to.emit(vesting, 'Claim(address,uint256)')
         .withArgs(user1.address, claimedAmount)
     })
   })
 
-  describe('# getClaimableAmount', () => {
+  describe('# getClaimedAmount(address,uint256)', () => {
+    before(() => {
+      amountsAllocated = [parseEther('1')]
```
XavierEkkel, line 526: nit:
`amountsAllocated` as the array is only used once, whereas `amountsAllocated[0]` is used repeatedly

make `amountsAllocated[0]` a variable instead, and use `[amountsAllocated]` for the one time you need the array

apps/smart-contracts/token/test/Vesting.test.ts
```diff
@@ -503,19 +509,97 @@ describe('Vesting', () => {
...
+      const duration = testEndTime - testStartTime
+      const timeBetweenEachWithdrawal = duration / numberOfClaims
+      const claimingTimes = Array.from(
+        Array(numberOfClaims).keys(),
+        (x) => testStartTime + (x + 1) * timeBetweenEachWithdrawal
+      )
+      const amountToClaimEachTime = amountsAllocated[0].div(numberOfClaims)
+
+      /**
+       * Ensure invariants are not violated across multiple claims,
```
XavierEkkel, line 555: it's just a single invariant (the eq check) right? (not multiple)

apps/smart-contracts/token/test/Vesting.test.ts
```diff
@@ -503,19 +509,97 @@ describe('Vesting', () => {
...
+      for (let i = 0; i < numberOfClaims; i++) {
+        await setNextTimestamp(ethers.provider, claimingTimes[i])
+        await vesting.connect(user1).claim()
+        // eslint-disable-next-line no-restricted-syntax
+        for (let j = 0; j < numberOfClaims; j++) {
+          if (j <= i) {
+            const claimedAmountJustBefore = await vesting['getClaimedAmount(address,uint256)'](
+              user1.address,
+              claimingTimes[j] - 1
+            )
```
XavierEkkel, line 567:
```suggestion
            )
            
```

apps/smart-contracts/token/test/Vesting.test.ts
```diff
@@ -503,19 +509,97 @@ describe('Vesting', () => {
...
+      /**
+       * Ensure invariants are not violated across multiple claims,
+       * for both retroactive and future timestamps.
+       */
+      for (let i = 0; i < numberOfClaims; i++) {
+        await setNextTimestamp(ethers.provider, claimingTimes[i])
+        await vesting.connect(user1).claim()
+        // eslint-disable-next-line no-restricted-syntax
+        for (let j = 0; j < numberOfClaims; j++) {
+          if (j <= i) {
```
XavierEkkel, line 563: can move this into the for loop condition check
and in general, strongly avoid nestings this deep

apps/smart-contracts/token/test/Vesting.test.ts
```diff
@@ -503,19 +509,97 @@ describe('Vesting', () => {
...
+      const timeBetweenEachWithdrawal = duration / numberOfClaims
+      const claimingTimes = Array.from(
+        Array(numberOfClaims).keys(),
+        (x) => testStartTime + (x + 1) * timeBetweenEachWithdrawal
+      )
+      const amountToClaimEachTime = amountsAllocated[0].div(numberOfClaims)
+
+      /**
+       * Ensure invariants are not violated across multiple claims,
+       * for both retroactive and future timestamps.
```
XavierEkkel, line 556: generally not a good idea to test multiple types of scenarios in one test as if it fails it's hard to know why. ideally separate into separate tests w/ separate helpers. this may also make it easier to read+understand

apps/smart-contracts/token/test/Vesting.test.ts
```diff
@@ -453,20 +455,20 @@ describe('Vesting', () => {
 
     it('transfers multiple times if still vesting', async () => {
       await mockERC20Token.transfer(vesting.address, amountsAllocated[0])
-      const numberOfWithdrawals = 2
+      const numberOfWithdrawals = 5
```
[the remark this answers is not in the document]
ramenforbreakfast: I felt 2 claims was not sufficient enough to demonstrate a condition holding across multiple timeframes

apps/smart-contracts/token/test/Vesting.test.ts
ramenforbreakfast: I just added this comment to explain why the testcases following all use this double for loop structure. It is so that we can check all relevant timestamps at each claim, for multiple claims.

apps/smart-contracts/token/test/Vesting.test.ts
ramenforbreakfast: Im not sure if there is a misunderstanding, but this isn't "multiple scenarios". All this test does is assert that the claimed amount is added to the amount before claiming, when a claim happens. Everytime a claim happens, we are asserting this condition holds for all timestamps in the future, including the current timestamp.

apps/smart-contracts/token/test/Vesting.test.ts
ramenforbreakfast: Yes!

apps/smart-contracts/token/contracts/vesting/Vesting.sol
```diff
@@ -162,6 +191,48 @@ contract Vesting is IVesting, Pausable, SafeOwnable, WithdrawERC20 {
...
+        high = mid;
+      } else {
+        low = mid + 1;
+      }
+    }
+    return
+      high == 0 ? 0 : _recipientToClaimedAmount[recipient][high - 1].amount;
+  }
+
+  function _getClaimableAmount(address recipient, uint256 timestamp)
```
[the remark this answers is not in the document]
ramenforbreakfast: Not sure why we wouldn't just change it in this PR? It doesn't make the PR smaller because the function signatures need to be changed anyway due to the presence of 2 `getUnclaimedAmount` (previously `getClaimableAmount`) functions.

apps/smart-contracts/token/test/Vesting.test.ts
ramenforbreakfast: This is only being done because `amountsAllocated` is used by all other testcases (I'm not a fan of how this is done in this test suite, I didnt write it and didn't want to waste time refactoring this) and is already an array type, i cannot turn it into a standalone number type like you want.

apps/smart-contracts/token/test/Vesting.test.ts
ramenforbreakfast: I think you are misunderstanding what this testcase is doing.
The `[j-1]` here, means _The last time a claim happened_
The `[j]-1` here, means _Right before the current claim_

The time in between these points is the time in between claims, and we are ensuring the value does not change in-between these times, which is what the test case states it is testing.

apps/smart-contracts/token/test/Vesting.test.ts
```diff
@@ -503,19 +509,97 @@ describe('Vesting', () => {
...
+    before(() => {
+      amountsAllocated = [parseEther('1')]
+    })
+
+    beforeEach(async () => {
+      await deployVesting()
+      await vesting.setAllocations([user1.address], amountsAllocated)
+      await mockERC20Token.transfer(vesting.address, amountsAllocated[0])
+    })
+
```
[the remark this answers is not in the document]
ramenforbreakfast: The end tests already exist, and kind of are overkill since this is dependent on claiming, not necessarily a specific time case. These end tests were bundled with the "never claimed" tests, and I have now just put them as separate "never claimed" tests.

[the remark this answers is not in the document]
ramenforbreakfast: I am instead going to create a test case to verify that a user does not affect another user's claimed amount, since it is inaccurate to say the function returns 0 if another user claims.

apps/smart-contracts/token/test/Vesting.test.ts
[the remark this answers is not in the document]
ramenforbreakfast: Resolving since it has been decided to add additional simpler 2 round test cases in a follow up PR.

apps/smart-contracts/token/test/Vesting.test.ts
ramenforbreakfast: Resolving this since a follow up PR will be adding testcases that will not have for loops

apps/smart-contracts/token/test/Vesting.test.ts
ramenforbreakfast: Resolving this since a follow up PR will be adding testcases that will not have for loops

apps/smart-contracts/token/test/Vesting.test.ts
ramenforbreakfast: Resolving this since a follow up PR will be adding testcases that will not have for loops

ramenforbreakfast:
This PR has been modified with the following major changes:
- `getClaimableAmount` has been changed to `getUnclaimedAmount`
- for loops for looping through different timestamps have been simplified to remove conditional logic that can just be moved into the loop condition.
- Added further commenting to for loop test cases
- Added a bunch of =0, start time, end time test cases for both `getClaimedAmount` and `getUnclaimedAmount`

Simplified versions of the more complex loop testcases will be submitted later.

apps/smart-contracts/token/test/Vesting.test.ts
```diff
@@ -585,25 +726,210 @@ describe('Vesting', () => {
...
+          const unclaimedAmount = await vesting['getUnclaimedAmount(address,uint256)'](
+            user1.address,
+            claimingTimes[j]
+          )
+          expect(unclaimedAmount).eq(0)
+        }
+      }
+    })
+
+    it('returns claimable amount at times in between a claim', async () => {
```
XavierEkkel, line 865: nit: change claimable -> unclaimed in these test case names

XavierEkkel (approved):
LGTM after nit

PR #1508, opened 2023-07-18
feat: add simplified versions of invariant testcases without loops

ramenforbreakfast:
This contains versions of the testcases submitted in the prior PR that do not involve loops and only use <=2 claims to validate their claims.


apps/smart-contracts/token/test/Vesting.test.ts
```diff
@@ -862,6 +942,40 @@ describe('Vesting', () => {
       }
     })
 
+    it('returns claimable amount at times in between a claim', async () => {
```
XavierEkkel, line 945: could you restructure it as follows:
Act: do a claim, move time forward, then do another claim, move time forward
Assert: check unclaimed amounts at 0, before+at first claim, between first+second claim, at second claim, after second claim (before current timestamp), at current timestamp, at future timestamp
id also call it `returns unclaimed amounts after multiple claims`

this also makes the other tests around multiple claims redundant

XavierEkkel (approved):
LGTM after comment

PR #1509, opened 2023-07-19
feat: redirect to `eth.limo`

MeLlamoPablo:
This PR redirects the production app at `app.prepo.io` to `prepo.eth.limo`

The following should be true after this PR:

* `app.prepo.io` redirects to `prepo.eth.limo`
  * The redirect happens server side and no javascript is loaded
* PR previews don't redirect to `prepo.eth.limo`
* `prepo.eth.limo` (nor the FE viewed via any other gateway like `eth.link` or `cf-ipfs.com`) doesn't redirect to `prepo.eth.limo` (obviously)


XavierEkkel (approved):
LGTM

PR #1519, opened 2023-07-19
feat: use prepo shared contracts with foundry without tgz

chrisling-dev:
https://www.notion.so/prepo/6ff94d2f628847cd97c80bfbe667b5b9?v=1160865fbe4c4639af9d63918c7b7d17&p=19ed9dcfa8b9469cb58443d5da6e5a8b&pm=s

This solution uses remappings to map @prepo-shared-contracts to node_modules/@prepo-shared-contracts via foundry.toml rather than modifying how it's installed.


ramenforbreakfast (approved):
Looks good, thanks a bunch! I have also gone ahead and removed the tgz file.

PR #1524, opened 2023-07-21
chore: upgrade to node v20

MeLlamoPablo:
[v16 hits EOL by September 11th](https://nodejs.org/en/blog/announcements/nodejs16-eol), so we should be upgraded by then


MeLlamoPablo:
@ramenforbreakfast I had to upgrade `hardhat` because the current version blocks us from using anything above v16. Is this okay?

chrisling-dev (approved):
LGTM!

PR #1532, opened 2023-07-25
feat: upgrade React to v18 and Next to v13

MeLlamoPablo:
This PR upgrades React to v18 and Next to v13. This is required to integrate Web3Auth, which doesn't work with react <18.


chrisling-dev (approved):
LGTM

PR #1540, opened 2023-07-28
feat: add loading skeleton for balance value

MeLlamoPablo:
https://www.notion.so/prepo/add-loading-box-for-balance-value-rather-than-the-entire-balance-text-just-popping-in-fa63a3c6c9284874bdf461abccdbfa43


chrisling-dev (approved):
LGTM after fixing type

packages/prepo-ui/src/components/CurrencyInput/CurrencyInput.tsx
```diff
@@ -196,13 +197,19 @@ const CurrencyInput: React.FC<
...
+          {!balanceNotVisible && (
             <>
-              {`Balance: ${displayDecimals(balance)}`}
-              {value !== balanceAfterGas && !isBalanceZero && (
-                <MaxButton disabled={disabled} onClick={handleMax}>
-                  MAX
-                </MaxButton>
+              {balance === undefined ? (
+                <Skeleton width={140} height={16} />
+              ) : (
```
chrisling-dev, line 204: Could we have a loading prop instead, which is only true if connected && balance === undefined

currently it shows loading for non connected account
```suggestion
              {loading ? (
                <Skeleton width={140} height={16} />
              ) : (
```

chrisling-dev (changes requested):

chrisling-dev (approved):

PR #1546, opened 2023-08-01
refactor: switch to use onlyAdmin instead of onlyOwner

ramenforbreakfast:
This PR involves changing `SafeOwnable` to `SafeAdmin`, and the `onlyOwner` modifier to `onlyAdmin`.

For simplicity, all that has been done is change the name of the contract and add an additional modifier `onlyAdmin` that is a carbon copy of `onlyOwner`.

However, it was not specified whether there was a desire to wholly change the `Owner` nomenclature used throughout. If so, this will require keeping a local copy of `Ownable` and its dependencies, modified to use the `Admin` naming instead of `Owner` and `Ownership`. 

Additionally, there are some minor drawbacks, in that if we change things like `owner` to `admin`, tools like OZ Defender depend on the assumption of an `owner` getter.

I have chosen to keep it simple and not completely change it since the majority of the changes would apply here regardless (`SafeAdmin` + `onlyAdmin` renames) and can be added in a future PR.


ramenforbreakfast:
Closing due to decision not to go through with renaming

PR #1548, opened 2023-08-01
refactor: move checkpoint functions to library

ramenforbreakfast:
Moved basic `Checkpoint` functions like pushing an element and a lookup into a shared library. These are taken from OpenZeppelin.


ramenforbreakfast:
Updated with a very quick and simple optimization. OZ contains many other superfluous methods I felt no reason to replicate, but they do have a simple `latest` method to obtain the latest element of a Checkpoint array. 

I felt that this was unnecessary to have as a separate method and just have a simple line to detect if the timestamp being queried is the current one. Having to search when we already know we want the latest element can be very wasteful especially if the array is large enough.

packages/prepo-shared-contracts/contracts/libraries/Checkpoints.sol
```diff
@@ -0,0 +1,63 @@
...
+
+library Checkpoints {
+  struct Checkpoint {
+    uint256 timestamp;
+    uint256 amount;
+  }
+
+  error InvalidCheckpointKey();
+
+  function _push(
```
XavierEkkel, line 14:
```suggestion
  function push(
```

packages/prepo-shared-contracts/contracts/libraries/Checkpoints.sol
```diff
@@ -0,0 +1,63 @@
...
+    uint256 timestamp,
+    uint256 amount
+  ) internal {
+    uint256 pos = checkpoints.length;
+    if (pos > 0) {
+      // Copying to memory is important here.
+      Checkpoint memory last = checkpoints[pos - 1];
+      // Checkpoints keys must be increasing.
+      if (last.timestamp > timestamp) revert InvalidCheckpointKey();
+      // Update or push new checkpoint
```
XavierEkkel, line 25: nit: self-explanatory
```suggestion
```

XavierEkkel (approved):
LGTM after comments

apps/smart-contracts/token/contracts/vesting/Vesting.sol
```diff
@@ -203,27 +194,7 @@ contract Vesting is IVesting, Pausable, SafeOwnable, WithdrawERC20 {
     view
```
XavierEkkel, line 194: since this function is now a one-liner, it's not needed anymore. just call the relevant checkpoint library function directly

PR #1551, opened 2023-08-02
chore: commit Yarn cache to make builds faster

MeLlamoPablo:
This PR commits the Yarn cache to the repo. The `.yarn/cache` directory weights ~500MB and contains a compressed copy of every dependency. [The directory structure is optimized to be committed](https://yarnpkg.com/features/zero-installs#is-it-different-from-just-checking-in-the-node_modules-folder). This makes the tradeoff of slowing down initial clones to gain the following advantages:

* Build times will be reduced as all packages won't have to be redownloaded on every build (they will be downloaded via `git clone` but this is faster, especially on GitHub actions)
    * Measurements on the "Checks" action:
        * Clone time +20s
        * Install time -6m15s
* `yarn install` can now run without an internet connection (except when installing new packages, obviously) which is great because we stop relying on third party servers like NPM
* We become more resilient against supply-chain vulnerabilities, because the CI will not download packages from external sources, only use the pre-approved cache committed to the repo.
* We have increased visibility over the dependencies that we're installing (e.g. if we upgrade one dependency but the old package isn't deleted, it means that other parts of the project are not upgraded and still use the old dependency - this was harder to visualize before)
* Moves a step towards migrating to [Yarn Plug'n'Play](https://yarnpkg.com/features/pnp) - this is a mode in which `yarn install` is not required at all and instead Node accesses `.yarn/cache` on runtime, replacing our ~3.2 GB `node_modules` directory with a ~500MB `.yarn/cache`.

    Migrating to this will require some effort and it may not be worthwhile to prioritize it, but it's still good to have this option open. 

The tradeoff of having slower initial clones is fine because initial clones are invoked much less frequently than yarn installs.


chrisling-dev (changes requested):
https://github.com/prepo-io/prepo-monorepo/pull/1554
https://github.com/prepo-io/prepo-monorepo/pull/1555

PRs branched off this seemed to be failing checks, could you take a look

MeLlamoPablo:
> #1554 #1555
> 
> PRs branched off this seemed to be failing checks, could you take a look

#1554 seems to include additional cache files that are not necessary, and Yarn fails because of that. Where did they come from?

chrisling-dev:
> https://github.com/prepo-io/prepo-monorepo/pull/1554 seems to include additional cache files that are not necessary, and Yarn fails because of that. Where did they come from?

Not sure, I pulled your branch, ran yarn, and simply added everything to commit, I'll try again

MeLlamoPablo:
> > #1554 seems to include additional cache files that are not necessary, and Yarn fails because of that. Where did they come from?
> 
> Not sure, I pulled your branch, ran yarn, and simply added everything to commit, I'll try again

This branch contains enough cache files. Running `yarn` shouldn't download more. Maybe they are leftovers from another branch, you could try:

```
# ensure your working directory is clean
rm -r .yarn/cache
git reset HEAD --hard
```

chrisling-dev (approved):

PR #1552, opened 2023-08-02
feat: add checkpointing to allocations

ramenforbreakfast:
Added checkpointing to allocations and updated all access of allocation state to use shared `Checkpoint` library functions.

One thing you might notice is that, since obtaining these values is just directly accessing the state itself (as opposed to `getUnclaimedAmount` or `getVested`), I did not bother to separate these into `internal` functions. This deviates from `getClaimedAmount`, which we did add an `internal` function for even though its also just accessing state directly. However, unlike `getClaimedAmount`, we only access allocations once in our logic, so it doesn't make the code much prettier anyway.


apps/smart-contracts/token/test/Vesting.test.ts
```diff
@@ -312,7 +327,78 @@ describe('Vesting', () => {
 
       await vesting.setAllocations([user1.address], [higherAllocationAmount])
 
-      expect(await vesting.getAllocatedAmount(user1.address)).eq(higherAllocationAmount)
+      expect(await vesting['getAllocatedAmount(address)'](user1.address)).eq(higherAllocationAmount)
+    })
+
+    it('checkpoints individual allocations', async () => {
+      const timeInBetweenAllocations = (testEndTime - testStartTime) / 3
```
XavierEkkel, line 334: basing it off these variables will be fragile e.g. if start time and end time are set too close
better to just hardcode a big enough value here

same for the other tests

apps/smart-contracts/token/test/Vesting.test.ts
```diff
@@ -1132,48 +1286,92 @@ describe('Vesting', () => {
...
+      expect(
+        await vesting['getVestedAmount(address,uint256)'](user1.address, allocationChangeTime)
+      ).eq(
+        increasedAllocationAmount
+          .mul(allocationChangeTime - testStartTime)
+          .div(testEndTime - testStartTime)
+      )
+    })
+
+    it('changes unclaimed amount if allocation decreased', async () => {
```
XavierEkkel, line 1354:
```suggestion
    it('changes vested amount if allocation decreased', async () => {
```

XavierEkkel (approved):
LGTM after comments

PR #1553, opened 2023-08-03
feat: add claim authorization state and adjustor

ramenforbreakfast:
Since this was a fresh test I was writing and not one I was editing from the existing test suite, I decided to greatly simplify the format for readability by removing for loops and simple integer amounts since for verifying whether the setter is adding/subtracting, we do not need significant amounts to avoid integer math lossiness.


apps/smart-contracts/token/test/Vesting.test.ts
```diff
@@ -402,6 +403,117 @@ describe('Vesting', () => {
...
+      ).revertedWithCustomError(vesting, 'ArrayLengthMismatch')
+    })
+
+    it('authorizes to single recipient', async () => {
+      await vesting.adjustAuthorizedAmounts([user1.address], [1])
+
+      expect(await vesting['getAuthorizedAmount(address)'](user1.address)).eq(1)
+    })
+
+    it('authorizes to multiple recipients', async () => {
```
XavierEkkel, line 431: add test for 'multiple same recipients' and call this one 'multiple different recipients'

XavierEkkel (approved):
LGTM after comments

PR #1556, opened 2023-08-03
feat: move intercom to `SettingsCard` on mobile

MeLlamoPablo:
https://github.com/prepo-io/prepo-monorepo/assets/11708035/b085bfab-a905-4454-8623-3194f33d491d

This PR refactors the `Intercom` provider to expose an `useIntercom` hook which the rest of the app can use to control it.

Since the loading of `Intercom` is async, `useIntercom` is made suspending (which means that components that use it must be inside `Suspense`). Suspending hooks allow for a much more ergonomic API, because they can return awaited values rather than forcing its users to deal with `loading` and `error` states. In order to create a suspending hook, the tiny `suspend-react` utility is added.

The `IntercomButton` component is introduced. This component leverages `useIntercom` to hide the default widget while it is rendered. Thus, we don't have to add any complex synchronization logic: as long as `IntercomButton` is being rendered anywhere within the app, the default button will be hidden.


apps/frontend/core-dapp/src/components/SettingsMenu/SettingsCard.tsx
```diff
@@ -54,6 +58,15 @@ const SettingsCard: React.FC<{ onClose: () => void; portfolioValue?: string }> =
...
         <NetworkStatus />
+        {isMobile && (
+          <IntercomButton>
+            {({ onClick }) => (
+              <SettingsMenuItem iconName="help-circle" onClick={onClick}>
+                Help and Feedback
+              </SettingsMenuItem>
+            )}
+          </IntercomButton>
+        )}
```
chrisling-dev, line 69: Alternative solution to avoid using external JS logics, wrap with a div that hides with css

apps/frontend/core-dapp/src/components/Intercom/IntercomButton.tsx
```diff
@@ -0,0 +1,32 @@
...
+    }
+  }, [intercom])
+
+  if (!intercom) return null
+
+  return <>{children({ onClick: intercom.show })}</>
+}
+
+const Wrapper: FC<Props> = ({ children }) => (
+  <Suspense fallback={<>{/* nothing */}</>}>
```
chrisling-dev, line 27: I believe you could use null to achieve the same thing:
```suggestion
  <Suspense fallback={null}>
```

apps/frontend/core-dapp/src/components/Intercom/IntercomButton.tsx
```diff
@@ -0,0 +1,32 @@
+import { FC, ReactNode, Suspense, useEffect } from 'react'
+import { useIntercom } from './index'
+
+type Props = {
+  children: (params: { onClick: () => void }) => ReactNode | undefined
```
chrisling-dev, line 5: Probably should enforce return of ReactNode?

```suggestion
  children: (params: { onClick: () => void }) => ReactNode
```

apps/frontend/core-dapp/src/components/SettingsMenu/SettingsDropdown.tsx
```diff
@@ -31,6 +31,11 @@ const SettingsDropdown: React.FC = () => {
         trigger={['click']}
         placement="bottomRight"
         size="sm"
+        // Render even when hidden to:
+        // 1. Preload icons
+        // 2. Render IntercomButton which in turn may hide the default Intercom
+        //    widget
```
chrisling-dev, line 37:
```suggestion
        // Render even when hidden to:
        // 1. Preload icons
        // 2. Render IntercomButton which in turn may hide the default Intercom widget
```

apps/frontend/core-dapp/src/components/Intercom/IntercomButton.tsx
MeLlamoPablo: Apparently can also remove it entirely haha

apps/frontend/core-dapp/src/components/SettingsMenu/SettingsCard.tsx
MeLlamoPablo: But we need the JS logic to make the call to `window.Intercom` and tell them to hide the default widget

apps/frontend/core-dapp/src/components/Intercom/IntercomButton.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+  useEffect(() => {
+    if (!intercom) return () => {}
+
+    intercom.setWidgetVisibility(false)
+
+    return () => {
+      intercom.setWidgetVisibility(true)
+    }
+  }, [intercom])
```
chrisling-dev, line 19: Unclear what this is doing, could you elaborate? Does setWidgetVisibility(false) hide the Intercom widget on desktop?

chrisling-dev (changes requested):
Have some questions around the design of IntercomButton, but need more context around Intercom

apps/frontend/core-dapp/src/components/Intercom/IntercomButton.tsx
chrisling-dev, line 19: okay so it seems this logic is controlling whether the Intercom widget should show, and hence the overall logic seems to be
```
if isMobile -> show IntercomButton

if IntercomButton is rendered -> hide widget
else -> show widget
```

I think there's 2 problems with this approach:
1. The overall logic isn't intuitive. Whether Intercom widget is hidden should be directly connected to whether this is on mobile, and not controlled via the visibility of IntercomButton. I'd recommend just having the isMobile logic in the `Intercom` component that acts as wrapper in `_app`.
2. The IntercomButton seems like it's made to be reusable with everything being very generic but I suspect it will start causing unexpected behaviour if we reuse it. (e.g. if we somehow show the IntercomButton on desktop somewhere, it will cause the Intercom widget to be hidden.) 
-> And also since realistically, there's no plan to reuse this component anywhere else, I think we should remove the useEffect logic, move this component to `SettingsMenu` and build the logics there directly like other SettingsMenu components, which I believe would make the code more readable

chrisling-dev (changes requested):

apps/frontend/core-dapp/src/components/SettingsMenu/SettingsDropdown.tsx
chrisling-dev, line 37: and then this comment should be updated as it already seems like bad design choice that the visibility of Intercom widget is affected by SettingsDropdown component, but makes sense to keep it for preloading icons
```suggestion
        // force render to preload icons
```

chrisling-dev (changes requested):

apps/frontend/core-dapp/src/components/SettingsMenu/SettingsCard.tsx
chrisling-dev: It's not straightforward/readable that the visibility of intercom widget is decided by a logic here in the `SettingsCard`, consider the newer changes I've requested

MeLlamoPablo:
@chrisling-dev simplified the code a bit, thanks for your suggestions! I think it's more straightforward now.

* Removed `IntercomButton` - if we're not reusing it right now it doesn't make sense
* Added logic on `Intercom/index.tsx` to hide it on mobile
  * This requires `Intercom` to load after `AppBootstrap` given that the theme is being used
* Called `intercom.show` directly on `SettingsCard`

chrisling-dev (approved):
Awesome! LGTM and tested that it works on both desktop/mobile

PR #1557, opened 2023-08-03
feat: hide Intercom when Torus is open

MeLlamoPablo:
https://github.com/prepo-io/prepo-monorepo/assets/11708035/3d8ddd5a-5b49-4a23-b7d0-98dedfe532f1

This PR updates the Intercom behavior to be hidden when Torus is open, because Torus provides its own Intercom-like messenger button at the same position.

The PR works by defining a `TorusTrap` component with a `MutationObserver` that listens to new elements attached to `document.body` and scans for an `iframe#torusIframe`. If detected, a `ResizeObserver` listens to changes of the size of the iframe. When it is closed, the iframe measures 70x70. When opened, it occupies 100% of the screen. Using that information we can determine when should we hide our Intercom.


apps/frontend/core-dapp/src/components/TorusTrap.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+export function useDetectTorusIframe(): HTMLIFrameElement | undefined {
+  const [torusIframe, setTorusIframe] = useState<HTMLIFrameElement | undefined>(undefined)
+
+  useEffect(() => {
+    const node = document.querySelector('iframe#torusIframe')
+    if (node instanceof HTMLIFrameElement) {
+      setTorusIframe(node)
+    }
+  }, [setTorusIframe])
```
chrisling-dev, line 12:
```suggestion
  }, [])
```

apps/frontend/core-dapp/src/components/TorusTrap.tsx
```diff
@@ -0,0 +1,97 @@
...
+      observer.disconnect()
+    }
+  }, [torusIframe])
+
+  return isOpen
+}
+
+const TorusTrap: FC = () => {
+  const intercom = useIntercom()
+  const torusIframe = useDetectTorusIframe()
```
chrisling-dev, line 76: The only use of `torusIframe` is to be passed into `useDetectTorusIframeOpen`. I'd recommend combining both `useDetectTorusIframe` and `useDetectTorusIframeOpen` and simply have a single `useTorus` hook, which returns { isOpen }

chrisling-dev (changes requested):
Nice work, LGTM after combining both hooks

apps/frontend/core-dapp/src/components/TorusTrap.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+export function useDetectTorusIframe(): HTMLIFrameElement | undefined {
+  const [torusIframe, setTorusIframe] = useState<HTMLIFrameElement | undefined>(undefined)
+
+  useEffect(() => {
+    const node = document.querySelector('iframe#torusIframe')
+    if (node instanceof HTMLIFrameElement) {
+      setTorusIframe(node)
+    }
+  }, [setTorusIframe])
```
chrisling-dev, line 12: saw your latest commit to remove this but somehow this is still here?

```suggestion
  }, [])
```

apps/frontend/core-dapp/src/components/TorusTrap.tsx
```diff
@@ -0,0 +1,93 @@
+import { FC, Suspense, useEffect, useState } from 'react'
+import { useIntercom } from './Intercom'
+
+export function useDetectTorusIframe(): HTMLIFrameElement | undefined {
```
chrisling-dev, line 4: While I prefer having the logics in both hooks combined into 1 hook rather than making 2 hooks, if you prefer this approach, we should not export `useDetectTorusIframe` as it is not useful anywhere else
```suggestion
function useDetectTorusIframe(): HTMLIFrameElement | undefined {
```

chrisling-dev (changes requested):

apps/frontend/core-dapp/src/components/TorusTrap.tsx
MeLlamoPablo: There were two of them, fixed both now

apps/frontend/core-dapp/src/components/TorusTrap.tsx
MeLlamoPablo: Good catch, removed `export` from both. You're right I'd rather keep them separate as I think their purpose is different (one detects the iframe presence, other detects the iframe size)

chrisling-dev (approved):
LGTM!

PR #1558, opened 2023-08-04
feat: add logic for checking authorizations during claiming

ramenforbreakfast:
I decided that the cleanest way to implement this would be to simply claim the authorized amount if it was not sufficient for the whole unclaimed amount. If this ends up being zero, it will be caught by our existing zero checking logic.

Also, since this is technically a change in authorization, I decided to emit the `AuthorizationChange` event here as well.


XavierEkkel (approved):
LGTM after comments

PR #1560, opened 2023-08-10
feat: resolve .eth and .arb names via Space ID

MeLlamoPablo:
Screenshot 2023-08-10 at 14 41 06

Screenshot 2023-08-10 at 14 41 42

This PR integrates ENS and .arb resolution via [Space ID](https://space.id/). If the connected wallet resolves to a .eth or .arb name the UI will display it (giving priority to .arb). Additionally, if it doesn't have a .arb name but has a .eth name with an avatar, the avatar will also be displayed.


apps/frontend/core-dapp/src/components/SettingsMenu/WalletIdentity.tsx
```diff
@@ -0,0 +1,93 @@
...
+  getName: (address: string) => Promise<{ name: string | null }>
+}
+
+const useSpaceId = ({
+  address,
+}: {
+  address: string
+}): {
+  dotArbName?: string
+  dotEthName?: string
```
chrisling-dev, line 26: could we make a type for this

apps/frontend/core-dapp/src/components/SettingsMenu/WalletIdentity.tsx
```diff
@@ -0,0 +1,93 @@
...
+    if (dotArbName) return { dotArbName }
+    return { dotEthName: dotEthName ?? undefined }
+  }, [address, arbitrumProvider])
+}
+
+const PlainAddressIdentity: FC<
+  Props & {
+    dotArbName?: string
+    dotEthName?: string
+  }
```
chrisling-dev, line 69: and reuse here
```suggestion
  Props & SpaceID
```

apps/frontend/core-dapp/src/components/SettingsMenu/WalletIdentity.tsx
```diff
@@ -0,0 +1,93 @@
...
+      sidAddress: getSidAddress(NETWORKS.mainnet.chainId),
+    })
+
+    const [{ name: dotArbName }, { name: dotEthName }] = await Promise.all([
+      arbId.getName(address),
+      ethId.getName(address),
+    ])
+
+    if (dotArbName) return { dotArbName }
+    return { dotEthName: dotEthName ?? undefined }
```
chrisling-dev, line 61: Wouldn't the empty value of dotArbName and dotEthName be undefined?
```suggestion
    return { dotArbName, dotEthName }
```

apps/frontend/core-dapp/src/components/SettingsMenu/WalletIdentity.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+const SpaceId: FC<Props> = ({ address }) => {
+  const { dotArbName, dotEthName } = useSpaceId({ address })
+  return <PlainAddressIdentity address={address} dotArbName={dotArbName} dotEthName={dotEthName} />
+}
+
+const WalletIdentity: FC<Props> = ({ address }) => (
+  <Suspense fallback={<PlainAddressIdentity address={address} />}>
+    <SpaceId address={address} />
+  </Suspense>
```
chrisling-dev, line 90: Doesn't seem like we need suspense here

The WalletIdentity could just implement the useSpaceId hook and the jsx directly. Since the fallback is PlainAddressIdentity without ens name `{dotArbName ?? dotEthName ?? getShortAccount(address)}` would essentially achieve the same thing:

- when spaceid isn't loaded: show address
- if spaceid is loaded: show ens
- if unavailable: shw address

chrisling-dev (changes requested):

apps/frontend/core-dapp/src/components/SettingsMenu/WalletIdentity.tsx
```diff
@@ -0,0 +1,93 @@
...
+      diameterDesktop={23}
+      diameterMobile={23}
+    />
+    {dotArbName ?? dotEthName ?? getShortAccount(address)}
+  </Flex>
+)
+
+const SpaceId: FC<Props> = ({ address }) => {
+  const { dotArbName, dotEthName } = useSpaceId({ address })
+  return <PlainAddressIdentity address={address} dotArbName={dotArbName} dotEthName={dotEthName} />
```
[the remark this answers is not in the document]
chrisling-dev: irrelevant if you take the approach I recommend in the last comment

apps/frontend/core-dapp/src/components/SettingsMenu/WalletIdentity.tsx
MeLlamoPablo: No, because the Space ID SDK returns null if they can't find an ID, so I'd rather not mix null and undefined

apps/frontend/core-dapp/src/components/SettingsMenu/WalletIdentity.tsx
MeLlamoPablo: Yes, `usePromise` would essentially achieve the same thing. The reason I did suspense is so that we easily could migrate this code to another codebase (also why I'm not storing the results of space ID/avatar on a store and just pulling the data on hooks). But I'm not opposed to using `usePromise` here, if we ever migrate this we can consider switching back.

apps/frontend/core-dapp/src/hooks/useSpaceId.ts
```diff
@@ -0,0 +1,56 @@
...
+        ? coreProvider
+        : createFallthroughProvider(NETWORKS.arbitrumOne),
+    [coreProvider]
+  )
+
+  const result = usePromise(
+    useCallback(async () => {
+      // @ts-expect-error - Library does not provide types
+      const { default: SID, getSidAddress } = await import('@siddomains/sidjs')
+
```
chrisling-dev, line 33: since arbitrumProvider is not reused I reckon we could just create it within the callback?
```suggestion
  const result = usePromise(
    useCallback(async () => {
      // @ts-expect-error - Library does not provide types
      const { default: SID, getSidAddress } = await import('@siddomains/sidjs')
      const arbProvider = coreProvider.network.chainId === NETWORKS.arbitrumOne.chainId
          ? coreProvider
          : createFallthroughProvider(NETWORKS.arbitrumOne)
```

chrisling-dev: also the pattern is more consistent to how `ethProvider` is created

chrisling-dev (approved):
LGTM after comments

apps/frontend/core-dapp/src/hooks/useSpaceId.ts
```diff
@@ -0,0 +1,56 @@
...
+        sidAddress: getSidAddress(NETWORKS.mainnet.chainId),
+      })
+
+      const [{ name: dotArbName }, { name: dotEthName }] = await Promise.all([
+        arbId.getName(address),
+        ethId.getName(address),
+      ])
+
+      if (dotArbName) return { dotArbName }
+      return { dotEthName: dotEthName ?? undefined }
```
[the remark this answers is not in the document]
MeLlamoPablo: This is done intentionally, because `useSpaceId` should decide which identity is displayed in case the user has multiple. Since `.arb` has priority, if the user has both, `.eth` is not returned. This is important because in case the user has `.arb` and `.eth`, we don't want to display the `.eth` avatar (and `.arb` has no avatar suppport, so displaying `.eth` avatar with `.arb` name can be confusing)

So instead to avoid being misleading I've updated the `SpaceId` type to explicitly describe that both can't coexist at the same time.

PR #1567, opened 2023-08-12
fix: update shared-contracts and token folder for foundry

ramenforbreakfast:
To prepare for writing foundry tests for new contracts, as well as our Checkpoint library, I have incorporated fixes Chris has outlined in PR #1529, and put them in here.

@chrisling-dev you should rebase your existing foundry work off this PR, since it's better we separate mandatory configuration fixes/changes like the ones included here from the actual test files you wrote in 1529, since review of these files would block the fixes from being pushed.


ramenforbreakfast:
Going to remove reviewers off this PR. @chrisling-dev it seems that the chai matchers in all of our tests are no longer accessible, causing all tests to fail. I have a feeling it is related to the yarn v3 nohoist command you added, specifically 
```
"installConfig": {
     "hoistingLimits": "workspaces"
}
```

ramenforbreakfast:
Update: I think this is due to yarn v3 no longer supporting per-package granularity for nohoisting.

ramenforbreakfast:
Issues with yarn v3 no-hoisting have been solved. Issue was due to a conflict with nohoist forcing chai to be saved individually across each workspace. Moving chai to the root directory solved this conflict.

XavierEkkel (approved):

